PR #237, opened 2015-01-22
better logging

bhuism:


brettwooldridge:
Looking at this further, while I agree it is not a good practice to hard-code a different class into a logger, such as ``AbstractHikariConfig``:
```java
Logger LOGGER = LoggerFactory.getLogger(HikariConfig.class);
```
I also don't like logging from an "abstract class".  In this case, I prefer:
```java
Logger LOGGER = LoggerFactory.getLogger(getClass());
```
``AbstractHikariConfig`` has two subclasses, ``HikariConfig`` and ``HikariDataSource``.  Depending on which is logging, that would be class that appears in the log.  Unfortunately for you, this will still likely result in ``HikariConfig`` appearing in the log.  Users will have to do the mental math if they jump into the code in HikariConfig.java to understand that they need to drill into the abstract class.

bhuism:
I would not recommend sharing instances of Logger i.e. always private, I would recommend always to:
```private static final Logger log = LoggerFactory.getLogger(Foo.class);```
if not static (but still private) you can do:
```private final Logger log = LoggerFactory.getLogger(getClass());```
I'm sure there are best practices for logging around
I you don't, I would still use your connection pool :+1:

PR #254, opened 2015-02-04
Expose the PoolMbean calls to the datasource and verify they actually work

ryandietrich:
Hi.  I messaged you some time ago about some NuProcess threading issues, thanks for helping out on that front.  Today I was upgrading my database connection library in a project and saw that BoneCP was no longer being maintained.  I looked over Hikari and came away impressed!  The only functionality missing from BoneCP that I needed was the ability to access meta-data on the pool in real time for some generic status services my system exposes.

So, here's a simple patch.  If you'd rather do this some other that's fine.  I'm just angling for a solution, and thought it would be better to show up with one than to just ask you for one.

### DIFF
--- a/hikaricp-common/src/main/java/com/zaxxer/hikari/HikariDataSource.java
+++ b/hikaricp-common/src/main/java/com/zaxxer/hikari/HikariDataSource.java
@@ -79,6 +79,26 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
       multiPool.put(new MultiPoolKey(getUsername(), getPassword()), pool);
    }
 
+   /** Expose the PoolMbean's idle connection count */
+   public int getIdleConnections() {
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@ryandietrich Thanks for the pull request.  We have considered adding methods on the datasource, but as we increase the number of metrics it would result in an increasing surface area in terms of API.

HikariCP exposes statistics and metrics that are accessible programmatically in three ways, you can read about them here:

[JMX Monitoring](https://github.com/brettwooldridge/HikariCP/wiki/JMX-Monitoring)<br/>
[DropWizard Metrics](https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-Metrics)<br/>
[DropWizard HealthChecks](https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-HealthChecks)<br/>

PR #284, opened 2015-03-14
Fix #280: Use class name for Logger in BaseHikariPool

### LINKED ISSUE #280: log level is not respected?
Hi,
I was happily using the version 2.2.5 of the HikariCP-java6 artifact with the log level set to INFO. I'm using Slf4j with log4j. Then, I came upon a new version (2.3.3) and wanted to switch to it. But the pool clean up stats seem to be logged with the DEBUG level, even though I did not change any logging setting. Logger name I'm using is "com.zaxxer.hikari".

harsel:
This resolves issue #280.


brettwooldridge:
I went ahead and made the logger non-static and used getClass()

harsel:
Thanks for fixing the issue.

PR #294, opened 2015-04-01
293 - Export Dropwizard Metrics via JMX.

dharrigan:
This small patch enables the exporting of Dropwizard Metrics via JMX if
HikariCP JMX is enabled.

-=david=-

### DIFF
--- a/hikaricp-common/src/main/java/com/zaxxer/hikari/pool/HikariMBeanElf.java
+++ b/hikaricp-common/src/main/java/com/zaxxer/hikari/pool/HikariMBeanElf.java
@@ -16,6 +16,9 @@
 
 package com.zaxxer.hikari.pool;
 
+import com.codahale.metrics.JmxReporter;
+import com.codahale.metrics.MetricRegistry;
+
 import java.lang.management.ManagementFactory;
 
 import javax.management.MBeanServer;
@@ -74,6 +77,12 @@ public final class HikariMBeanElf
       catch (Exception e) {
          LOGGER.warn("Unable to register management beans.", e);
       }
+
+      final Object metricRegistry = configuration.getMetricRegistry();
+      if (metricRegistry != null) {
+          JmxReporter.forRegistry((MetricRegistry)metricRegistry).build().start();
+      }
+
    }
 
    /**


brettwooldridge:
My feeling is that this is something best left to the applicaton, but I have some questions anyway:
 * What happens if the MetricsRegistry passed into HikariCP was already used by the application to start a JmxReporter instance?
 * A MetricsRegistry can be shared across HikariCP pools (there can be N number of pool instances in a JVM).  HikariCP registers MBeans in JMX under a unique name for each pool, and creates uniquely named metrics (prefixed by poolName) within a the MetricsRegistry.  What is the effect of creating N instances of JmxReporter, all pointing at a single MetricsRegistry?

dharrigan:
Hi Brett,

Good questions. 

With regard the first question, it's possible to set a "name" per JmxReporter created, similar to what you achieve with the setting of the poolConfigName and poolName within the same class - this way, each JmxRegister would have a different name.

With regard the second question - no idea yet :-)

Say, do you use IntelliJ for your code? I'm having an "interesting" time, trying to get it to work with the project layout... :-(

-=david=-

brettwooldridge:
Well, since this would be a fairly major change (if we pick it up), it would likely go into the 2.4.0 release (*dev* branch) rather than 2.3.x.  The *dev* branch (where your pull should come from) has been completely flattened, and should be much more IDE friendly.

dharrigan:
Hi Brett,

Wonderful :-) Thanks for the clarification. We can talk more about this for 2.4.0 if you wish at a later date when you are ready to consider it :-) In the mean time, I'm going to try and get the 'dev' branch to work with IntelliJ :-)

-=david=-

dharrigan:
Yup,

dev branch imported successfully into IntelliJ Ta. :-)

-=david=-

dharrigan:
Hi,

Actually, looked at this a bit closer. It appears that where, for example, you add in the configuration.getPoolName() into the "type" (lines 61-62) of HikariMBeanElf, one is unable to do this with the JmxReporter. Although the domain can be set in the JmxReporter (builder) as in 'inDomain("blah")', the "type" is set at lines 655-657 in JmxReporter (v3.1.1) - where it uses "timers" or "meters" as the "type".

It appears that without subclassing the DefaultObjectNameFactory to append on the poolName (in createName), it would not be possible to uniquely identify the metrics by pool name....

Thoughts?

-=david=-

brettwooldridge:
Interesting.  I'll have to dig into it a little deeper.  Any assistance you can provide researching (or asking dropwizard guys) will surely help.  Unfortunately, it's rolling up on midnight here in Tokyo, so I definitely won't get a look at it tonight.

jnehlmeier:
Why would this be useful? If you use dropwizard metrics then you already have a reporter in place in your app which should already cover pool metrics since they are registered to the app wide MetricsRegistry singleton. Of course you could have multiple MetricsRegistries but usually you don't and usually you only have a low number of reporters.

So I would not like to see Hikari automatically create JmxReporters if Hikari MBeans are enabled because:
1.) Usually you already have a reporter for your app
2.) Just because you want the HIkari MBean to be active does not automatically mean that you want metrics to be reported to JMX. For example our app directly reports metrics to a time series database.
3.) An app can have hundreds of pools and having hundreds of JmxReporter each also registering one MetricsRegistryListener but then only handle a small subset of all registered metrics seems quite a bit of overhead in that situation. 

I believe that reporting is app specific and this feature must be configurable to be able to opt-in/out.

brettwooldridge:
While we do appreciate the effort, we tend.to agree. The only other option would be a pool setting that controlled the behavior, but if you've read any of our rants about minimalism and keeping the configuration "surface area" as small as possible we would have to decline such a request.

dharrigan:
Hi!

Not a problem, feel free to close the issue - I fully understand the points of view and after consideration tend to agree too :-) That's why I like software development, bouncing ideas of clever people with insights that I didn't consider before :-) 

-=david=-

PR #307, opened 2015-04-15
Improvements

ash2k:
Some improvements. Nothing particularly big, but still useful (I hope). WDYT?

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -16,12 +16,13 @@
 
 package com.zaxxer.hikari;
 
-import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
@@ -800,8 +801,8 @@ public class HikariConfig implements HikariConfigMBean
 
    protected void loadProperties(String propertyFileName)
    {
-      final File propFile = new File(propertyFileName);
-      try (final InputStream is = propFile.isFile() ? new FileInputStream(propFile) : this.getClass().getResourceAsStream(propertyFileName)) {
+      final Path propFile = Paths.get(propertyFileName);
+      try (final InputStream is = Files.isRegularFile(propFile) ? Files.newInputStream(propFile) : this.getClass().getResourceAsStream(propertyFileName)) {
          if (is != null) {
             Properties props = new Properties();
             props.load(is);
--- a/src/main/java/com/zaxxer/hikari/HikariDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/HikariDataSource.java
@@ -22,6 +22,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Wrapper;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.sql.DataSource;
 
@@ -43,7 +44,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
 {
    private static final Logger LOGGER = LoggerFactory.getLogger(HikariDataSource.class);
 
-   private volatile boolean isShutdown;
+   private final AtomicBoolean isShutdown = new AtomicBoolean();
 
    private final HikariPool fastPathPool;
    private volatile HikariPool pool;
@@ -78,7 +79,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
    @Override
    public Connection getConnection() throws SQLException
    {
-      if (isShutdown) {
+      if (isShutdown.get()) {
          throw new SQLException("Pool " + pool.getConfiguration().getPoolName() + " has been shutdown");
       }
 
@@ -232,7 +233,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
     */
    public void evictConnection(Connection connection)
    {
-      if (!isShutdown && pool != null && connection instanceof IHikariConnectionProxy) {
+      if (!isShutdown.get() && pool != null && connection instanceof IHikariConnectionProxy) {
          pool.evictConnection((IHikariConnectionProxy) connection);
       }
    }
@@ -243,7 +244,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
     */
    public void suspendPool()
    {
-      if (!isShutdown && pool != null) {
+      if (!isShutdown.get() && pool != null) {
          pool.suspendPool();
       }
    }
@@ -253,7 +254,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
     */
    public void resumePool()
    {
-      if (!isShutdown && pool != null) {
+      if (!isShutdown.get() && pool != null) {
          pool.resumePool();
       }
    }
@@ -264,11 +265,9 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
    @Override
    public void close()
    {
-      if (isShutdown) {
+      if (isShutdown.getAndSet(true)) {
          return;
       }
-
-      isShutdown = true;
 
       if (pool != null) {
          try {
@@ -291,7 +290,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
     */
    public boolean isClosed()
    {
-      return isShutdown;
+      return isShutdown.get();
    }
 
    /**
--- a/src/main/java/com/zaxxer/hikari/HikariDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/HikariDataSource.java
@@ -142,6 +142,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
    }
 
    /** {@inheritDoc} */
+   @Override
    public java.util.logging.Logger getParentLogger() throws SQLFeatureNotSupportedException
    {
       throw new SQLFeatureNotSupportedException();
--- a/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
+++ b/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
@@ -85,7 +85,7 @@ public class HikariJNDIFactory implements ObjectFactory
       String jndiName = properties.getProperty("dataSourceJNDI");
       DataSource jndiDS = (DataSource) context.lookup(jndiName);
       if (jndiDS == null) {
-         context = (Context) (new InitialContext());
+         context = new InitialContext();
          jndiDS = (DataSource) context.lookup(jndiName);
       }
 
--- a/src/main/java/com/zaxxer/hikari/metrics/CodaHaleMetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/CodaHaleMetricsTracker.java
@@ -100,7 +100,7 @@ public final class CodaHaleMetricsTracker extends MetricsTracker
    @Override
    public void recordConnectionUsage(final PoolBagEntry bagEntry)
    {
-      connectionUsage.update(elapsedTimeMs(bagEntry.lastOpenTime));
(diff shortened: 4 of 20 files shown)


brettwooldridge:
@ash2k I'm a bit concerned about the nanoTime() change, here is why.  Strictly following the spec. as provided by the JavaDoc for ``System.nanoTime()``:
```
The value returned represents nanoseconds since some fixed but arbitrary origin time
(perhaps in the future, so values may be negative).
```
The fact that values *may* be negative, even in the absence of a long wraparound, significantly complicates it's use.  Take this example:
```java
1  long timeoutMs = 30000;  // 30 seconds in ms
2  final long hardTimeout = TimeUnit.MILLISECONDS.toNanos(timeout);
3  final long start = System.nanoTime();
4  do {
5      // do something
6      ...
7      timeoutMs = hardTimeout - TimeUnit.NANOSECONDS.toMillis( System.nanoTime() - start );
8  while (timeoutMs > 0L);
```
If it is theoretically possible for ``System.nanoTime()`` to start from a negative base, which according to the JavaDoc it is, rather than ``timeoutMs`` getting smaller with each loop, as would be intuitive, it instead grows.

For example, if on line 3 the ``System.nanoTime()`` method sets ``start = -9999999``, and after some period the ``timeoutMs`` is calculated as on line 8, with ``System.nanoTime()`` now returning ``-8888888``, the "elapsed" calculation is ``-9999999 - (-8888888) = -1111111``, and therefore ``timeoutMs = hardTimeout - (-1111111)``, which is *increasing* the timeout rather than decreasing it.

This of course can be guarded against by using ``Math.abs()``, but doing so further obscures readability and impacts performance.

In many environments, ``nanoTime()`` actually drifts more than the wall-clock by a significant amount:

http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6519418
https://community.oracle.com/thread/1662207

ash2k:
=> `timeoutMs = hardTimeout - TimeUnit.NANOSECONDS.toMillis( System.nanoTime() - start )`
=> `timeoutMs = hardTimeout - TimeUnit.NANOSECONDS.toMillis(-8888888 - -9999999)`
=> `timeoutMs = hardTimeout - TimeUnit.NANOSECONDS.toMillis(-8888888 + 9999999)`
=> `timeoutMs = hardTimeout - TimeUnit.NANOSECONDS.toMillis(1111111)`
=> all good =)

That bug is about JavaDocs if I'm reading it correctly. It was fixed long time ago. The forum thread is also 9 years old.
[This one](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6458294) is interesting - note the last paragraph:

> Certainly nanoTime should be no worse than currentTimeMillis() and so should always be preferred for doing timing measurements.

Interesting blogpost [Nanotrusting the Nanotime](http://shipilev.net/blog/2014/nanotrusting-nanotime/#_timers):
> We are not taking care of currentTimeMillis, because its precision is beyond any joke. We skip new Date() and other "clever" ways to measure time for the same reason.

Also that blogpost mentions [Inside the Hotspot VM: Clocks, Timers and Scheduling Events](https://blogs.oracle.com/dholmes/entry/inside_the_hotspot_vm_clocks), which tells us:
> If you are interested in measuring absolute time then always use System.currentTimeMillis(). Be aware that its resolution may be quite coarse (though this is rarely an issue for absolute times.)
>
> If you are interested in measuring/calculating elapsed time, then always use System.nanoTime(). On most systems it will give a resolution on the order of microseconds. Be aware though, this call can also take microseconds to execute on some platforms.

Also [this one](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6900441) and referenced bugs are interesting. Most of them were fixed or will be fixed eventually.

So it looks like `System.nanoTime()` is not perfect but it is the best tool provided by the JVM at the moment.

petercgrant:
@brettwooldridge @ash2k is correct.  Using System.currentTimeMillis() to measure elapsed time can give [incorrect results](http://stackoverflow.com/questions/1964363/wrong-time-with-system-currenttimemillis-java).  I've seen this most pronounced on Xen guests, which have historically had [issues with clock drift](http://blog.neutrino.es/2010/workaround-for-the-time-drift-issue-on-xen-keep-your-guests-synced/) that are then corrected by NTP, causing currentTimeMillis to run forward too fast then backward.  Using System.nanoTime() to measure elapsed time is the correct solution; [the nanoTime javadoc](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()) even says overflow is only an issue if the two calls are separated by 292 years.  If it has any sway, [Guava's Stopwatch](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/base/Stopwatch.html) is implemented using nanoTime.  Please accept the rest of this pull request.

brettwooldridge:
@ash2k @petercgrant Ok, I'm convinced, pending no significant impact in the benchmarks.

ash2k:
So, what happened to the PR? I see commits from my branch in the `dev` branch, but changes are not there. I don't understand.

brettwooldridge:
@ash2k Patience, please.  I'm juggling several balls at once.  The merge was a cherry-pick, but the ``nanoTime()`` change is not forgotten, and I will come back around and pick it up.  But it may be a few days before that happens.

brettwooldridge:
@ash2k @petercgrant 

Revisiting this.  We're seeing a severe degradation in the benchmarks after the ``nanoTime()`` change.

**Without** ``nanoTime()`` (branch "experimental"):
```
Benchmark                       (maxPoolSize)  (pool)   Mode  Cnt       Score      Error   Units
ConnectionBench.cycleCnnection             32  hikari  thrpt   16   33071.000 ± 1240.955  ops/ms
StatementBench.cycleStatement              32  hikari  thrpt   16  144820.638 ± 5977.288  ops/ms
```

**With** ``nanoTime()`` (branch ''poor-nano"):
```
Benchmark                       (maxPoolSize)  (pool)   Mode  Cnt       Score       Error   Units
ConnectionBench.cycleCnnection             32  hikari  thrpt   16   16983.590 ±   550.191  ops/ms
StatementBench.cycleStatement              32  hikari  thrpt   16  127759.977 ± 16657.963  ops/ms
```

In the Connection Cycle benchmark we are seeing roughly 50% throughput.  This of course will vary by OS, but there is now the question whether this change is worth it.

Alexsey's comments re: ``currentTimeMillis()`` should certainly be taken in context -- i.e. measuring the elapsed time of extremely brief events.  Also in the same article, Alexsey's states:
```
The lessons from this part are:
...
  * nanoTime is a scalability bottleneck; up to the point that everything you measure would be
    the contended performance of nanoTime itself.
```
In terms of calculated intervals, HikariCP essentially has no sub-second timing requirements at all.  The primary purported benefit of ``nanoTime()`` to HikariCP is that it is *monotonic* and not subject to *drift/skew*.  But at least at present, monotonicity is only guaranteed on Solaris and MacOS X (but with the above performance penalty on OS X).

([bug_id=6458294](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6458294))
```
CLOCK_MONOTONIC can be affected by skew due to system time changes. The more recent
CLOCK_MONOTONIC_RAW (linux only) is not supposed to be affected at all. We can't use the
new clock until our supported build platform is upgraded to aa version of Linux that supports it.
```

([DOs and DON’Ts of Conducting Performance Measurements in Java](http://d3s.mff.cuni.cz/~steinhauser/icpe2015tt.pdf))
```
The most easily accessible time source in Java is calling System.nanoTime. The exact behavior
of this method is platform-dependent, for example OpenJDK 8 on Linux queries the CLOCK 
MONOTONIC system clock, OpenJDK 8 on Windows uses the QueryPerformanceCounter call.
...
Internally, the high precision time sources may use hardware counters with unknown or varying
frequency. Such sources are callibrated against counters with known frequency but possibly
lower accuracy, this callibration can yield slightly different frequency estimate on each
initialization. Also, the CLOCK MONOTONIC system clock is subject to adjustments on systems
with NTP or PTP support, even during measurement.
```
Even though [this article](http://stas-blogspot.blogspot.jp/2012/02/what-is-behind-systemnanotime.html) is old, it seems much of what it documents is still true.  Even on systems with HPET, the underlying OS APIs that Java tends to use rely on the TSC, which is subject to skew and retrograde motion across threads in multi-core CPUs.  It also seems that nanoTime() drift due to the TSC gets worse over longer measurement intervals (10s of seconds or minutes), which is also the kinds of durations being measured in HikariCP.

We may be left with implementing an "abstract" clock source that is selected at runtime based on the OS, but the question stands, "What is gained?"  Particularly in the face of increased complexity.

petercgrant:
I was under the impression that nanoTime ran slower than currentTimeMillis on Windows but just as fast on every other OS; I guess that's wrong.

I've never seen nanoTime run backwards (its contract says it's not supposed to), but I have seen currentTimeMillis do just that, repeatedly and by design.  What is gained by using nanoTime is correct behavior when the system clock changes (barring bugs such as you found that cause it not to fulfill its contract).  What is lost (on at least some platforms) is apparently a great deal of performance.  I can understand why you wouldn't want to change it or would want to make it configurable.

ash2k:
@brettwooldridge 
Actually in the Alexey's blogpost there is a [link](https://github.com/shipilev/timers-bench) to a repository with timers benchmarks. Worth a look. There are also results of measurements on Linux.

Just run it on my laptop (Macbook pro A1398, 2.5GHz)
```
Java(TM) SE Runtime Environment, 1.8.0_45-b14
Java HotSpot(TM) 64-Bit Server VM, 25.45-b02
Mac OS X, 10.10.3, x86_64

Burning up to figure out the exact CPU count....... done!

Running with 1 threads and [-client]:
       granularity_currentTime:  996187.465 ±    338.172 ns
          granularity_nanotime:      34.440 ±      0.433 ns
           latency_currentTime:      39.339 ±      1.361 ns
              latency_nanotime:      34.086 ±      0.383 ns

Running with 4 threads and [-client]:
       granularity_currentTime:  997599.575 ±    648.268 ns
          granularity_nanotime:     176.333 ±      8.322 ns
           latency_currentTime:      40.663 ±      1.233 ns
              latency_nanotime:     174.466 ±      8.795 ns

Running with 8 threads and [-client]:
       granularity_currentTime: 1012589.914 ±   6263.736 ns
          granularity_nanotime:     240.029 ±      1.970 ns
           latency_currentTime:      48.511 ±      0.711 ns
              latency_nanotime:     239.878 ±      4.096 ns

Running with 1 threads and [-server]:
       granularity_currentTime:  995976.401 ±    302.372 ns
          granularity_nanotime:      33.940 ±      0.217 ns
           latency_currentTime:      38.608 ±      0.729 ns
              latency_nanotime:      34.016 ±      0.398 ns

Running with 4 threads and [-server]:
       granularity_currentTime:  997411.202 ±    727.904 ns
          granularity_nanotime:     184.165 ±      7.418 ns
           latency_currentTime:      40.264 ±      0.307 ns
              latency_nanotime:     173.603 ±      9.694 ns

Running with 8 threads and [-server]:
       granularity_currentTime: 1025633.732 ±  18043.076 ns
          granularity_nanotime:     240.391 ±      1.882 ns
           latency_currentTime:      54.397 ±      2.876 ns
              latency_nanotime:     240.437 ±      5.113 ns
```
I confirm that your results are reproducible on my machine. I don't know what the Hikari benchmark is showing us but it does look very suspicious to me. I have several ideas what may be causing this and I will try to find out (as time permits).

I still think it is worth using `nanoTime()` to measure intervals because it is a question of correctness. That bug [JDK-6458294](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6458294) is planned to be fixed in Java 9 and Java platform will eventually behave correctly. How likely is it that in a DB-interacting application many threads will be contending on `nanoTime()` call inside the connection pool? I doubt anyone will be having a performance bottleneck in a database interaction because of `nanoTime()` =)

brettwooldridge:
@ash2k Ii'm not sure what the source of the latency is, except that on some platforms nanoTime() uses a hardware source directly. Even so, intuitively I wouldn't have expected so much contention in the HikariCP benchmark because both the borrow and return code paths are two or three dozen lines each. However, if you envision an execution pipeline like a lane of traffic, once one car slows down, everything starts to pile up behind -- so eventually, possibly quite quickly, contention arises.

petercgrant:
@brettwooldridge I'm a little late on this, but thank you for giving this issue very thorough consideration and implementing a solution that balances correctness and performance.  The leap second we're having here in a few minutes reminded me of this issue and our beloved Java clock options.  Thanks!

brettwooldridge:
@petercgrant Hey, thanks Peter.  Let me know if you see any hiccups in HikariCP.  :smile:

PR #316, opened 2015-05-16
Fixes

ash2k:


src/main/java/com/zaxxer/hikari/HikariDataSource.java
```diff
@@ -159,13 +158,12 @@ public int getLoginTimeout() throws SQLException
...
-            return (T) pool.getDataSource();
+         final DataSource dataSource = pool.getDataSource();
+         if (iface.isInstance(dataSource)) {
+            return (T) dataSource;
          }
 
-         if (pool.getDataSource() instanceof Wrapper) {
-            return (T) pool.getDataSource().unwrap(iface);
-         }
+         return dataSource.unwrap(iface);
```
brettwooldridge: This seems like a good idea, but the reality is there are drivers out there compiled before Java 6, and calling like this without the *instanceof* check will result in an java.lang.AbstractMethodError -- rather than the more useful exception we throw below.

src/main/java/com/zaxxer/hikari/HikariDataSource.java
```diff
@@ -180,13 +178,8 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException
...
       if (pool != null) {
-         if (iface.isInstance(pool.getDataSource())) {
-            return true;
-         }
-
-         if (pool.getDataSource() instanceof Wrapper) {
-            return pool.getDataSource().isWrapperFor(iface);
-         }
+         final DataSource dataSource = pool.getDataSource();
+         return iface.isInstance(dataSource) || dataSource.isWrapperFor(iface);
```
brettwooldridge: Same as above.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -192,7 +193,7 @@ public final Connection getConnection(final long hardTimeout) throws SQLExceptio
             final long now = clockSource.currentTime();
             if (bagEntry.evicted || (clockSource.toMillis(now - bagEntry.lastAccess) > ALIVE_BYPASS_WINDOW_MS && !isConnectionAlive(bagEntry.connection))) {
                closeConnection(bagEntry, "connection evicted or dead"); // Throw away the dead connection and try again
-               timeout = hardTimeout - clockSource.elapsedTimeMs(startTime);
```
brettwooldridge: I find the *elapsedTimeMs()* form more readable, and performance-wise this method gets inlined anyway so the two are equivalent.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -210,7 +211,13 @@ public final Connection getConnection(final long hardTimeout) throws SQLExceptio
       }
 
       logPoolState("Timeout failure ");
-      throw new SQLTimeoutException(String.format("Timeout after %dms of waiting for a connection.", clockSource.elapsedTimeMs(startTime), lastConnectionFailure.getAndSet(null)));
+      final Throwable cause = lastConnectionFailure.getAndSet(null);
+      final SQLTimeoutException exception = new SQLTimeoutException(String.format("Timeout after %dms of waiting for a connection.", clockSource.elapsedTimeMs(startTime)));
+
+      if (cause != null) {
```
brettwooldridge: I don't understand the benefit of this.  Either way, if the cause is null the result is same, but the current code is simpler.

src/main/java/com/zaxxer/hikari/pool/PoolUtilities.java
```diff
@@ -56,7 +56,7 @@ public void quietlyCloseConnection(final Connection connection, final String clo
       final String addendum = closureReason != null ? " (" + closureReason + ")" : "";
       try {
          LOGGER.debug("Closing connection {} in pool {}{}", connection, poolName, addendum);
-         if (connection != null && !connection.isClosed()) {
+         if (!connection.isClosed()) {
```
brettwooldridge: It is possible for *connection* to be null here, when called from the catch of *HikariPool.addConnection()*.

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -380,7 +379,7 @@ public void setCatalog(String catalog) throws SQLException
    @Override
    public final boolean isWrapperFor(Class<?> iface) throws SQLException
    {
-      return iface.isInstance(delegate) || (delegate instanceof Wrapper && delegate.isWrapperFor(iface));
+      return iface.isInstance(delegate) || delegate.isWrapperFor(iface);
```
brettwooldridge: See previous comments re: the Wrapper interface.

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -388,13 +387,6 @@ public final boolean isWrapperFor(Class<?> iface) throws SQLException
...
    {
-      if (iface.isInstance(delegate)) {
-         return (T) delegate;
-      }
-      else if (delegate instanceof Wrapper) {
-          return (T) delegate.unwrap(iface);
-      }
-
-      throw new SQLException("Wrapped connection is not an instance of " + iface);
+      return iface.isInstance(delegate) ? (T) delegate : delegate.unwrap(iface);
```
brettwooldridge: Ditto.

src/main/java/com/zaxxer/hikari/proxy/ResultSetProxy.java
```diff
@@ -84,13 +83,6 @@ public void deleteRow() throws SQLException
...
-         return (T) delegate;
-      }
-      else if (delegate instanceof Wrapper) {
-          return (T) delegate.unwrap(iface);
-      }
-
-      throw new SQLException("Wrapped ResultSet is not an instance of " + iface);
-   }   
+      return iface.isInstance(delegate) ? (T) delegate : delegate.unwrap(iface);
+   }
```
brettwooldridge: Ditto.

src/main/java/com/zaxxer/hikari/proxy/StatementProxy.java
```diff
@@ -222,13 +221,6 @@ public ResultSet getResultSet() throws SQLException {
...
    {
-      if (iface.isInstance(delegate)) {
-         return (T) delegate;
-      }
-      else if (delegate instanceof Wrapper) {
-          return (T) delegate.unwrap(iface);
-      }
-
-      throw new SQLException("Wrapped statement is not an instance of " + iface);
+      return iface.isInstance(delegate) ? (T) delegate : delegate.unwrap(iface);
```
brettwooldridge: Ditto.

brettwooldridge:
What's up with the test failure?  Seems related to the HikariConfig changes possibly?
```java
com.zaxxer.hikari.TestMetrics
testHealthChecks(com.zaxxer.hikari.TestMetrics)  Time elapsed: 2.021 sec  <<< ERROR!
java.lang.NullPointerException: null
	at com.zaxxer.hikari.TestMetrics.testHealthChecks(TestMetrics.java:145)
```

brettwooldridge:
Selectively merged.

src/main/java/com/zaxxer/hikari/HikariDataSource.java
ash2k: I suspected there is some issue like this. Thanks, good to know.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
ash2k: If you set cause to `null` then it is impossible to use `initCause()` to set it to something useful (if needed) because it is like already set. Not much improvement, I agree. Other thing that is fixed here is that current code passes cause as a second argument to `String.format()` instead of the second argument to exception constructor (i.e. misplaced bracket).

```java
- throw new SQLTimeoutException(String.format("Timeout after %dms of waiting for a connection.", clockSource.elapsedTimeMs(startTime), lastConnectionFailure.getAndSet(null)));
+ throw new SQLTimeoutException(String.format("Timeout after %dms of waiting for a connection.", clockSource.elapsedTimeMs(startTime)), lastConnectionFailure.getAndSet(null));
```

PR #331, opened 2015-06-18
Allow the use of a user provided MetricsTracker

arthurdandrea:
This pull request intends to allow the use of other metrics libraries than Dropwizard Metrics. It uses the current MetricsTracker class without modification, it simply adds a MetricsTrackerFactory interface that will create an instance of a subclass of MetricsTracker.

The Dropwizard/CodeHale current implementation has been adapted to use a MetricsTrackerFactory, some methods have been rewritten but no method has been deleted or deprecated.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -42,6 +42,8 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.zaxxer.hikari.pool.PoolElf;
 import com.zaxxer.hikari.util.PropertyElf;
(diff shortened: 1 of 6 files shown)


brettwooldridge:
@arthurdandrea Thank you for the contribution.  Sorry for the delay integrating it.  Can you rebase it against the latest dev changes?  After the pull there are a few modifications that I'll make to hide some of the internal state and classes that are leaking (eg. ``PoolBagEntry``), but don't worry about changing anything in this pull, just rebase it.

brettwooldridge:
@arthurdandrea This is be merged post-2.4.0 release.  Please let me know if you can rebase your changes against the latest dev, and squash them if possible.

arthurdandrea:
Sorry for the delay, I rebased and squashed as asked.

I imagine that issue #335 can be solved using the same abstraction created in this pull request, am I right?

PR #345, opened 2015-07-16
Fix resolving property files

Shredder121:


Shredder121:
I have another PR incoming, but the tests ran (I by default skip tests during builds not related to testing) and this failed.

PR #347, opened 2015-07-20
HikariPool: Log consitency

nitincchauhan:
HikariPool: Log consitency
PoolElf: I think two more round trips to db/driver is overkill (for pool
to support bad driver) especially jdbc spec reads 'closing statement
must close any current result set'. It is already slow (compare to
others) having network timeout as well as query timeout to validate
connection!

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -208,7 +208,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
       if (originalException instanceof SQLException) {
          sqlState = ((SQLException) originalException).getSQLState();
       }
-      throw new SQLTransientConnectionException(poolName + " - Connection not available, timeout after " + clockSource.elapsedMillis(startTime) + "ms.", sqlState, originalException);
+      throw new SQLTransientConnectionException(poolName + " - Connection is not available, request timed out after " + clockSource.elapsedMillis(startTime) + "ms.", sqlState, originalException);
    }
 
    /**
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -196,7 +196,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          while (timeout > 0L);
       }
       catch (InterruptedException e) {
-         throw new SQLException("Interrupted during connection acquisition", e);
+         throw new SQLException(poolName + " - Interrupted during connection acquisition", e);
       }
       finally {
          suspendResumeLock.release();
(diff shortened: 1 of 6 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -1,652 +1,652 @@
...
-            }
-         }
-         
-         logPoolState("After cleanup ");
-
-         fillPool(); // Try to maintain minimum connections
-      }
-   }
-}
+/*
```
brettwooldridge: Nice to have a pull request.  However, it seems like some setting, either eclipse editor setting or eclipse egit setting, messed with every line-ending.  So rather than a diff. of only changed lines, it is show that every line was deleted and added.  Fix the line-ending issue (preferably Unix-style) and commit again.  I think the pull request will automatically be updated.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -1,652 +1,652 @@
...
+            else {
+               metricsContext.setConnectionLastOpen(bagEntry, now);
+               metricsContext.stop();
+               return ProxyFactory.getProxyConnection(bagEntry, leakTask.start(bagEntry), now);
+            }
+         }
+         while (timeout > 0L);
+      }
+      catch (InterruptedException e) {
+         throw new SQLException(poolName + " - Interrupted during connection acquisition", e);
```
brettwooldridge: Contrary to logs, I do not include pool names in exceptions.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: hmm ! I will try again.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: It will help to have poolName in exceptions too, not much overhead !

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -208,7 +208,7 @@ public final Connection getConnection(final long hardTimeout) throws SQLExceptio
       if (originalException instanceof SQLException) {
          sqlState = ((SQLException) originalException).getSQLState();
       }
-      throw new SQLTimeoutException(String.format("Timeout after %dms of waiting for a connection.", clockSource.elapsedMillis(startTime)), sqlState, originalException);
+      throw new SQLTimeoutException(poolName + " - Cannot acquire connection, Timeout after " + clockSource.elapsedMillis(startTime) + "ms.", sqlState, originalException);
```
brettwooldridge: Formatting much better, btw.  I'd like different wording than "Cannot acquire connection" ... it is ambiguous and the user might think it implies an acquire from the DB directly.  A little longer message is acceptable, "No connection available, ...".  Also, please de-capitalize "Timeout".

brettwooldridge: While we are touching this code, it turns out that [SQLTimeoutException](http://docs.oracle.com/javase/7/docs/api/java/sql/SQLTimeoutException.html) is not the correct exception to throw here.  HikariCP should be throwing [SQLTransientConnectionException](http://docs.oracle.com/javase/7/docs/api/java/sql/SQLTransientConnectionException.html) instead.  Both are subclasses of ``SQLTransientException``, but ``SQLTimeoutException`` is only for *Statement* execution timeouts.

nitincchauhan:
done and some more but it is failing tests:

Failed tests: 
  ShutdownTest.testAfterShutdown:228 null
Tests in error: 
  TestValidation.validateBothDriverAndDS:183 » IllegalArgument cannot use driver...
  TestValidation.validateMissingUrl:59 » IllegalArgument jdbcUrl is required wit...

nitincchauhan:
anything more required for merging ?

PR #361, opened 2015-07-31
[2.4.x] Ensure closed connections returned to pool

schlosna:
If an exception occurs while closing a connection's open statements
(e.g. NullPointerException due to concurrent close of statement with
connection), the connection was not returned to the pool.

We now check for this possibility and ensure the connection is returned
to the pool.


brettwooldridge:
How is this bug reached, and is it fixed by this commit https://github.com/brettwooldridge/HikariCP/commit/82536c0fe6c58279247f0ef4429c4a540ab927e3?

schlosna:
The repro is same as #362.

I will pull that commit and try to run a test on a cluster Monday.

Issue #270, opened 2015-02-26
NullPointerException when closing connection


hvesalai:
Hi,

we get this exception when running our code.

```
java.lang.NullPointerException: null
        at com.zaxxer.hikari.proxy.ConnectionProxy.close(ConnectionProxy.java:178)
```

My guess it that this is what happens:

- the are two threads
- thread 1 creates a statement from a connection object shared by the threads
- thread 2 closes the connection, entering line [ConnectionProxy#174](https://github.com/brettwooldridge/HikariCP/blob/2.3.x/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java#L174), where it reads the size of the openStatements list
- thread 1 closes the statement, resulting in the statement being removed from the openStatements list on line [ConnectionProxy#123](https://github.com/brettwooldridge/HikariCP/blob/2.3.x/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java#L123). This results in the openStatements list to be shrunk and a null to be inserted into the previously last position in line [FastList.java#L117](https://github.com/brettwooldridge/HikariCP/blob/2.3.x/hikaricp-common/src/main/java/com/zaxxer/hikari/util/FastList.java#L117)
- thread 2 advances to line [ConnectionProxy.java#L178](https://github.com/brettwooldridge/HikariCP/blob/2.3.x/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java#L178), where it get's `null` from the last `get(i)`

We have fixed our code so that it doesn't share connections between threads, so this is no longer an issue for us. JDBC specification doesn't say whether Connection should be thread safe, but at least psql jdbc driver is https://jdbc.postgresql.org/documentation/81/thread.html

Heikki

brettwooldridge:
JDBC 1.0 (1997), if I remember correctly, required Connections to be threadsafe but support was not widely adopted.  Starting with JDBC 2.0, Sun added language that largely backed away from the requirement without conceiting failure:
```
"In practice we expect that most of the JDBC objects will only be accessed in a single threaded
way. However some multi-thread support is necessary, and our attempts in previous drafts to
specify some classes as MT safe and some as MT unsafe appeared to be adding more confusion
than light."
```

PR #362, opened 2015-07-31
[2.3.x] Ensure closed connections returned to pool

schlosna:
If an exception occurs while closing a connection's open statements
(e.g. NullPointerException due to concurrent close of statement with
connection), the connection was not returned to the pool.

We now check for this possibility and ensure the connection is returned
to the pool.

### DIFF
--- a/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
+++ b/hikaricp-common/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
@@ -178,19 +178,9 @@ public abstract class ConnectionProxy implements IHikariConnectionProxy
       if (delegate != ClosedConnection.CLOSED_CONNECTION) {
          leakTask.cancel();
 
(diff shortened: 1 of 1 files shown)


schlosna:
Hi @brettwooldridge, thanks for the great work with Hikari! We recently switch some applications from c3p0 over to Hikari and have been running load tests and encountered the following NPE which seems like the same as issue #270. I realize that the JDBC spec does not require Connection, Statement, etc. to be thread safe, but I'd like to propose the changes on this PR to avoid these exceptions on close.

```
java.lang.NullPointerException
	at com.zaxxer.hikari.proxy.ConnectionProxy.close(ConnectionProxy.java:182)
```

brettwooldridge:
I don't have an issue with this change, in general, however I will note that as the FastList collection that holds the statements is not threadsafe (for performance), this change still does not provide a 100% guarantee that multi-thread access is safe.

brettwooldridge:
Thanks for the contribution!

PR #371, opened 2015-08-06
Dev

nitincchauhan:
fix arg to log


brettwooldridge:
Close for second try against correct branch.

PR #373, opened 2015-08-06
keeping test and assignment close in con.close()

nitincchauhan:


nitincchauhan:
changed hikariconfig as suggested.
reverted changes made in connection proxy ... reason was 'it may take longer between check in if and assignment in finally', meanwhile another call to close() or isClosed() can cause problems. coz connection.close() on already closed con should be no-op
maybe similar to isShutdown in HikariDataSource here would be better ?

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -177,13 +177,11 @@ public final void close() throws SQLException
    {
       if (delegate != ClosedConnection.CLOSED_CONNECTION) {
          leakTask.cancel();
+         closeOpenStatements();
```
brettwooldridge: closeOpenStatements() accidentally moved out of try block ... can you move it back?

nitincchauhan: done

PR #378, opened 2015-08-11
tweak isConnectionAlive and few other

nitincchauhan:
tweak isConnectionAlive(): do not set query timeout if network time out is already set
shutdown(): mainly, close connection bag after closing executor first.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -632,18 +632,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          }
 
          logPoolState("Before cleanup ");
-         boolean evicted = false;
-         List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
-         for (PoolBagEntry bagEntry : bag) {
-            if (bagEntry.evicted) {
-               closeConnection(bagEntry, "(connection evicted)");
-               evicted = true;
-            }
-         }
-         if (evicted) {
-        	//get fresh bag
-            bag = connectionBag.values(STATE_NOT_IN_USE);
-         }
+         final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
          int removable = bag.size() - config.getMinimumIdle();
          for (PoolBagEntry bagEntry : bag) {
             if (connectionBag.reserve(bagEntry)) {
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -631,15 +631,22 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          }
 
          logPoolState("Before cleanup ");
-         final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
+         boolean evicted = false;
+         List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
+         for (PoolBagEntry bagEntry : bag) {
+            if (bagEntry.evicted) {
+               closeConnection(bagEntry, "(connection evicted)");
+               evicted = true;
+            }
+         }
+         if (evicted) {
+        	//get fresh bag
+            bag = connectionBag.values(STATE_NOT_IN_USE);
+         }
          int removable = bag.size() - config.getMinimumIdle();
          for (PoolBagEntry bagEntry : bag) {
             if (connectionBag.reserve(bagEntry)) {
-               if (bagEntry.evicted) {
-                  closeConnection(bagEntry, "(connection evicted)");
-                  removable--;
-               }
-               else if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
+               if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
                   closeConnection(bagEntry, "(connection passed idleTimeout)");
                   removable--;
                }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -242,9 +242,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          LOGGER.info("{} - is closing down.", poolName);
          logPoolState("Before closing ");
 
-         connectionBag.close();
-         softEvictConnections();
-         addConnectionExecutor.shutdownNow();
+         addConnectionExecutor.shutdown();
          addConnectionExecutor.awaitTermination(5L, TimeUnit.SECONDS);
          if (config.getScheduledExecutorService() == null) {
             houseKeepingExecutorService.shutdown();
@@ -253,6 +251,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
 
          final ExecutorService assassinExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), "Hikari connection assassin",
                                                                            config.getThreadFactory(), new ThreadPoolExecutor.CallerRunsPolicy());
+         connectionBag.close();
          try {
             final long start = clockSource.currentTime();
             do {
@@ -365,7 +364,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
             while (poolState == POOL_NORMAL && totalConnections.get() < maxPoolSize && getIdleConnections() <= minimumIdle && !addConnection()) {
                // If we got into the loop, addConnection() failed, so we sleep and retry
                quietlySleep(sleepBackoff);
-               sleepBackoff = Math.min(connectionTimeout / 2, (long) ((double) sleepBackoff * 1.5));
+               sleepBackoff = Math.min((long) (connectionTimeout / 2), (long) (sleepBackoff * 1.5));
             }
          }
       }, true);
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -243,6 +243,8 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          LOGGER.info("{} - is closing down.", poolName);
          logPoolState("Before closing ");
 
+         connectionBag.close();
+         softEvictConnections();
          addConnectionExecutor.shutdown();
          addConnectionExecutor.awaitTermination(5L, TimeUnit.SECONDS);
          if (config.getScheduledExecutorService() == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -242,9 +242,7 @@ public final synchronized void shutdown() throws InterruptedException
          LOGGER.info("{} - is closing down.", poolName);
          logPoolState("Before closing ");
 
-         connectionBag.close();
-         softEvictConnections();
-         addConnectionExecutor.shutdownNow();
+         addConnectionExecutor.shutdown();
```
brettwooldridge: What is this?

nitincchauhan: shutdown(): mainly, close connection bag after. closing add connection executor first.

src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java
```diff
@@ -182,8 +182,10 @@ public boolean compareAndSet(int expect, int update)
    @Override
    public String toString()
    {
-      return String.format("%s (created %s, last release %dms ago, %s)",
-                           connection, formatDateTime(creationTime), ClockSource.INSTANCE.elapsedMillis(lastAccess), stateToString());
+      return connection
+         + ", created " + formatDateTime(creationTime)
+         + ", last release " + ClockSource.INSTANCE.elapsedMillis(lastAccess) + "ms ago, "
+         + stateToString();
```
brettwooldridge: At least replace with something like:
```
return new StringBuilder()
   .append(connecton)
   .append(", created ").append(formatDateTime(creationTime)
   .append(", last release ").append(ClockSource.INSTANCE.elapsedMillis(lastAccess)).append("ms ago")
   .append(stateToString()).toString();
```

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -219,7 +219,9 @@ boolean isConnectionAlive(final Connection connection, AtomicReference<Throwable
          final int originalTimeout = getAndSetNetworkTimeout(connection, validationTimeout);
    
          try (Statement statement = connection.createStatement()) {
-            setQueryTimeout(statement, timeoutSec);
+        	if (isNetworkTimeoutSupported != TRUE) {
+               setQueryTimeout(statement, timeoutSec);
```
brettwooldridge: indentation is not correct.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -365,7 +364,7 @@ public void run()
             while (poolState == POOL_NORMAL && totalConnections.get() < maxPoolSize && getIdleConnections() <= minimumIdle && !addConnection()) {
                // If we got into the loop, addConnection() failed, so we sleep and retry
                quietlySleep(sleepBackoff);
-               sleepBackoff = Math.min(connectionTimeout / 2, (long) ((double) sleepBackoff * 1.5));
+               sleepBackoff = Math.min((long) (connectionTimeout / 2), (long) (sleepBackoff * 1.5));
```
brettwooldridge: What is this change?

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: The bag is closed first to prevent connections from being added.  When the ``addConnection()`` calls ``concurrentBag.add()`` an ``IllegalStateException`` will be thrown, and that exception will be caught and the connection forced closed.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: should invoke math.min( long,long ) method

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
nitincchauhan: somehow In my IDE indentation looks ok !

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: It's already calling ``Math.min(long, long)``.

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
brettwooldridge: Maybe github is showing it wrong.  The 'if' and the 'setQueryTimeout()' appear at the same indentation in the pull request on their site.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: IMO it is bad to throw exception to divert program flow.
bag add() is innocent and should not throw exception.
when shutdown() closes executor, there wont be any 'new' connection added.
those already in queue (to be added) or in bag will be closed in eviction later in shutdown

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: connectionTimeout / 2 ...may not be long always

brettwooldridge: it is always a long because ``connectionTimeout`` is long, as defined by Java language specification.

nitincchauhan: you are right.

src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java
nitincchauhan: effect would be same but less clutter without 'appoend' and who knows '+' be optimized more in future?

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,13 +631,18 @@ public void run()
...
             if (connectionBag.reserve(bagEntry)) {
                if (bagEntry.evicted) {
                   closeConnection(bagEntry, "(connection evicted)");
+                  totIdle --;
                }
-               else if (idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
+               else if (totIdle > minIdle && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
                   closeConnection(bagEntry, "(connection passed idleTimeout)");
+                  totIdle --;
                }
```
brettwooldridge: Try:
```java
int removable = bag.size() - config.getMinimumIdle();
for (PoolBagEntry bagEntry : bag) {
   if (connectionBag.reserve(bagEntry)) {
      if (bagEntry.evicted) {
         closeConnection(bagEntry, "(connection evicted)");
         removable--;
      }
      else if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout){
         closeConnection(bagEntry, "(connection passed idleTimeout)");
         removable--;
      }
      else {
         connectionBag.unreserve(bagEntry);
      }
   }
}
```

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,13 +631,24 @@ public void run()
...
+         int removable = bag.size() - config.getMinimumIdle();
+         for (PoolBagEntry bagEntry : bag) {
             if (connectionBag.reserve(bagEntry)) {
-               if (bagEntry.evicted) {
-                  closeConnection(bagEntry, "(connection evicted)");
-               }
-               else if (idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
+               if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
                   closeConnection(bagEntry, "(connection passed idleTimeout)");
+                  removable--;
```
brettwooldridge: That's not going to work, and is overly complex.  It won't work because ``closeConnection()`` is *asynchronous*, so even if evicted connections are "closed" the second call to get STATE_NOT_IN_USE bag entries is going to return the exact same collection 99.9% of the time.  Take my suggested code, and simply remove the decrement of ``removable`` from the evicted path, its better and simpler:
```java
int removable = bag.size() - config.getMinimumIdle();
for (PoolBagEntry bagEntry : bag) {
   if (connectionBag.reserve(bagEntry)) {
      if (bagEntry.evicted) {
         closeConnection(bagEntry, "(connection evicted)");
      }
      else if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout){
         closeConnection(bagEntry, "(connection passed idleTimeout)");
         removable--;
      }
      else {
         connectionBag.unreserve(bagEntry);
      }
   }
}
```

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: So backout the cast in front of ``(connectionTimeout / 2)`` ...

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: That is certainly not a safe bet.  What happens if Connections are taking several seconds to create each?

brettwooldridge: Additionally, the call to ``softEvictConnections()`` *before* ``addConnectionExecutor.awaitTermination()`` is **intentional**.  It gives connections a better chance to close cleaning before we start aborting them forcefully.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: call to poolElf.quietlyCloseConnection() is async in closeConnection()
I considered not to decrement removables in case of evicted ..but that is problematic as removable is not calculated correctly with some evicted connection in bag

src/main/java/com/zaxxer/hikari/pool/HikariPool.java

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: reverted

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: Actually, I think we should remove the eviction check entirely here.  Basically, an evicted connection cannot escape being closed in normal code paths and I don't think the housekeeper needs to worry about it:

https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L186

https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L223

nitincchauhan: housekeeper does not close connections marked as evicted anymore. updated pull request

PR #383, opened 2015-08-14
break loop early

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -635,8 +635,11 @@ public void run()
          final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
          int removable = bag.size() - config.getMinimumIdle();
          for (PoolBagEntry bagEntry : bag) {
+            if (removable <= 0) {
+               break;
+            }
             if (connectionBag.reserve(bagEntry)) {
-               if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
+               if (idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
                   closeConnection(bagEntry, "(connection passed idleTimeout)");
```
brettwooldridge: Ew, that just looks icky.

Admitedly less efficient ... bypassing the reservation but still not breaking the loop.  Likely less than a microsecond of "spin" even if the pool had 500 idle connections.  Reads much nicer:
```java
for (PoolBagEntry bagEntry : bag) {
   if (removable > 0L && connectionBag.reserve(bagEntry)) {
      ...
   }
}
```

src/main/java/com/zaxxer/hikari/pool/PoolInitializationException.java
```diff
@@ -31,6 +31,6 @@
     */
    public PoolInitializationException(Throwable t)
    {
-      super("Exception during pool initialization", t);
+      super(t);
```
brettwooldridge: That's going to weird in a log, especially if the user suppresses stack traces.  If you're going to change it, this would be better:
```java
super("Exception during pool initialization: " + t.getMessage(), t);
```

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: probably bypassing reservation would be helpful under load :) and 3 lines more, code and intention is much clear.

src/main/java/com/zaxxer/hikari/pool/PoolInitializationException.java
nitincchauhan: will do

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: aha ...now i see :) will do as well

src/main/java/com/zaxxer/hikari/pool/PoolInitializationException.java
nitincchauhan: done

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -526,7 +526,7 @@ private void fillPool()
          addConnectionExecutor.execute(new Runnable() {
             @Override
             public void run() {
-               logPoolState("After fill ");
+               logPoolState("After fill\t");
```
brettwooldridge: Is that all the callers?  If so, might as well just put the \t in the ``logPoolState()`` method.

nitincchauhan: yes other two callers are from test package ..MiscTest and PostgresTest and called without parameter

nitincchauhan: not adding \t to method. \t with conditional param there looks ugly

Issue #379, opened 2015-08-12
HouseKeeper should not close idle connections once reached to min number of connections


yaojuncn:
https://github.com/brettwooldridge/HikariCP/blob/bc010fba486b27ae3d034cc9701e0c4217457ddb/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
line 633 to 650, in housekeeper check,  it will close all the idle connections; and then try to fill the connection pool;
in prod I've seen logs showing that the total connections are less than the configured min number;

this is incorrect in terms of 2 points: 
1. connection pool should keep the contract of keeping min number of connections in the pool; 
2. more important, the fillPool() in this way will cause a connection storm issue, which means the databases will receive a large number of new connections request, will might be a big overhead to database; 
while this issue is not that serious on single web servers, it becomes a big problem in distributed systems with good number of application servers connecting to the same database; 
in practice there're cases that connection storm caused a big problem for db;

brettwooldridge:
The minimumIdle is not a contract, the documentation states clearly:

This property controls the minimum number of idle connections that HikariCP
***tries*** to maintain in the pool. *If the idle connections dip below this value, 
HikariCP will make a* **best effort** *to add additional connections quickly 
and efficiently.*

Having said that, connections are added by a single threaded executor and is highly unlikely to overload a server.  In a environment with "a good number of application servers" within each pool connections will reach idle at different times, distributing the renewal of connections.

Actually, even within the same pool connection, connections with have different distribution as to when idle is reached.

yaojuncn:
1. "in practice there're cases that connection storm caused a big problem for db;" 
This is from practice as things like this did happen.  that's under a large volume, high concurrency , distributed system. 

although the different servers may create the connections at different times, when you say unlikely, they are still possible, which means in some cases they will happen; when there several thousand application servers connecting to one database, and usually the database server is under heavy load, any connection storm spike has a big impact on the whole system. 

2. with HikariCP, I've also seen the total number of connection reach to a very small number such as 4; 
(the config is 10-40);    if at this time there're some sudden traffic coming in, it will have to create many more connections (6 more in this case) because the pool has too little connections;  the current housekeeper logic may also lead to 0 connections in some cases, which at the moment is like no pool at all; 
 
3. stop closing the connections when reaching at min number of connections is surely better than closing them all and then refill them. the stop check is actually the common implementations in some other connection pools;  I do not see any reason why HikariCP wouldn't want to do it?

[...]

PR #384, opened 2015-08-16
revise connection pool cleanup logic by reducing unncessary loop,

yaojuncn:
some details from
https://github.com/brettwooldridge/HikariCP/issues/379
based on original pull request
https://github.com/brettwooldridge/HikariCP/pull/383

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -635,7 +635,7 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
             final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
             int removable = bag.size() - config.getMinimumIdle();
             for (PoolBagEntry bagEntry : bag) {
-               if (removable == 0) {
+               if (removable <= 0) {
                   break;
                }
                if (clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -631,17 +631,20 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          previous = now;
 
          logPoolState("Before cleanup\t");
-         final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
-         int removable = bag.size() - config.getMinimumIdle();
-         for (PoolBagEntry bagEntry : bag) {
-            if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
-               if (connectionBag.reserve(bagEntry)) {
-                  closeConnection(bagEntry, "(connection passed idleTimeout)");
-                  removable--;
-               }
-            }
-         }
-         
+          if(idleTimeout > 0L) {
+              final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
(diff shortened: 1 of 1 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,17 +631,20 @@ public void run()
...
-         for (PoolBagEntry bagEntry : bag) {
-            if (removable > 0 && idleTimeout > 0L && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
-               if (connectionBag.reserve(bagEntry)) {
-                  closeConnection(bagEntry, "(connection passed idleTimeout)");
-                  removable--;
-               }
-            }
-         }
-         
+          if(idleTimeout > 0L) {
```
brettwooldridge: Space between 'if' and parenthesis.  Indentation is 3 spaces, not 4.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,17 +631,20 @@ public void run()
...
+              final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
+              int removable = bag.size() - config.getMinimumIdle();
+              if (removable > 0) {
```
nitincchauhan: if (removable > 0) check needs to be inside for loop too

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,17 +631,20 @@ public void run()
...
+                  for (PoolBagEntry bagEntry : bag) {
+                      if (clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout) {
```
nitincchauhan: please remove outer if (removable > 0) and add it here:
if (removable > 0 && clockSource.elapsedMillis(bagEntry.lastAccess, now) > idleTimeout)

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -631,17 +631,21 @@ public void run()
          previous = now;
 
          logPoolState("Before cleanup\t");
```
brettwooldridge: Go ahead and move the before cleanup and after cleanup log inside of the ``if (idleTimeout < 0)`` conditional.

yaojuncn:
Btw, although it looks to me the last line of fillPool() can also be put into the if (idleTimeout < 0) part because ideally that would not be needed to fillPool() from the housekeeper any more; 
personally I'd prefer to keep it outside just in case; and there's actually no harm to be there;

brettwooldridge:
The ``fillPool()`` call still needs to be there.  Connections can be retired in a number of ways and need to be replaced.

For example, a separate task in the scheduler executes when a connection reaches its maxLifetime.  User's can also explicitly evict connections.  And finally, certain SQLExceptions from the database can result in a connection being evicted from the pool.

yaojuncn:
I see; Will read your code more for internals:)

brettwooldridge:
For quick reference, the maxLifetime retirement occurs [here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java#L87), user evictions occur [here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java#L87) and [here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L413), and SQLException related evictions occur [here](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java#L105).

yaojuncn:
Thx

PR #389, opened 2015-08-19
in close, if connection is marked for eviction, bypass to finally

nitincchauhan:


src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -147,8 +147,9 @@ public final void markCommitStateDirty()
       return statement;
    }
 
-   private final void closeOpenStatements()
+   private final boolean closeOpenStatements()
    {
+      boolean success = true;
```
brettwooldridge: Move the ``boolean success = true;`` inside the ``if (size > 0)`` conditional...

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -160,11 +161,13 @@ private final void closeOpenStatements()
             }
             catch (SQLException e) {
                checkException(e);
+               success = success ?  !poolEntry.evict : false;
```
brettwooldridge: Change to ``success &= !poolEntry.evict;``

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -160,11 +161,13 @@ private final void closeOpenStatements()
...
             }
          }
 
          openStatements.clear();
```
brettwooldridge: Add ``return sucess;`` here.

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -160,11 +161,13 @@ private final void closeOpenStatements()
...
       }
+      return success;
```
brettwooldridge: Change to ``return true;``

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
nitincchauhan: can't :( it has to close other statements

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
nitincchauhan: can't. it has to return true in all cases except in case of exception that sets evict in 'any' statement closing

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
nitincchauhan: can't :( it has to be based on var set in catch

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
nitincchauhan: if NOT closing statement here does not leak any resource, returning is better

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
brettwooldridge: Maybe it wasn't clean, I mean [this](https://gist.github.com/brettwooldridge/22171b1b83f7882d6aa7#file-snippet-java).

nitincchauhan: updated with suggestions

src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -147,24 +147,27 @@ public final void markCommitStateDirty()
...
-
-         openStatements.clear();
+         catch (SQLException e) {
+            checkException(e);
+            success &= !poolEntry.evict;
+         }
       }
+      openStatements.clear();
+      return success;
    }
```
brettwooldridge: I know it's nitpicking, but I tried rearranging the flow like this, but for some reason it reads worse to me.  I like the format in the gist (variation on your original).

nitincchauhan: you mean:
if (size <= 0) {
   return true;
}
I think returning early this way is more readable
this saves rest of method from indentation as well

nitincchauhan: ofcource change to style you like after merge :+1:

brettwooldridge:
Just to explain my reasoning on the formatting...
```java
private final boolean closeOpenStatements()
{
   final int size = openStatements.size();
   if (size <= 0) {
      return true;
   }
```
When you are reading the method for the first time, and you get to this point, it is not clear what returning ``true`` actually *means*.
```java
   boolean success = true;
   for (int i = 0; i < size; i++) {
```
And the distance in lines between where ``size`` is initialized and where it is referenced again here is longer, with a mental context switch of the return between.

The committed version...
```java
private final boolean closeOpenStatements()
{
   final int size = openStatements.size();
   if (size > 0) {
      boolean success = true;
      for (int i = 0; i < size; i++) {
```
has neither of these issues.  The ``boolean success`` declaration gives a clue about the meaning of the return value right away.  And the entire mental scope of the ``size`` variable occurs within 4 lines.

That's just the way my brain works, at least.

nitincchauhan:
thanks for details. makes sense

PR #390, opened 2015-08-20
fix: should consider connection marked for eviction else where too

nitincchauhan:
moved check so now it takes care of evict set during CloseOpenStatements() and 'before' that too.
at present it is not effective for most cases as it is checking evict only when user forgets to close statement.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
+++ b/src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
@@ -104,18 +104,20 @@ public abstract class ConnectionProxy implements IHikariConnectionProxy
    @Override
    public final SQLException checkException(final SQLException sqle)
    {
-      String sqlState = sqle.getSQLState();
-      if (sqlState != null) {
-         boolean isForceClose = sqlState.startsWith("08") || SQL_ERRORS.contains(sqlState);
-         if (isForceClose) {
-            poolEntry.evict = true;
-            LOGGER.warn("{} - Connection {} marked as broken because of SQLSTATE({}), ErrorCode({})",
-                        poolEntry.parentPool, poolEntry, sqlState, sqle.getErrorCode(), sqle);
-         }
-         else {
-            SQLException nse = sqle.getNextException();
-            if (nse != null && nse != sqle) {
-               checkException(nse);
+      if (!poolEntry.evict) {
+         String sqlState = sqle.getSQLState();
+         if (sqlState != null) {
+            boolean isForceClose = sqlState.startsWith("08") || SQL_ERRORS.contains(sqlState);
+            if (isForceClose) {
+               poolEntry.evict = true;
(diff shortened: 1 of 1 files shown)


src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -104,18 +104,20 @@ public final PoolBagEntry getPoolBagEntry()
...
-            LOGGER.warn("{} - Connection {} marked as broken because of SQLSTATE({}), ErrorCode({})",
-                        poolEntry.parentPool, poolEntry, sqlState, sqle.getErrorCode(), sqle);
-         }
-         else {
-            SQLException nse = sqle.getNextException();
-            if (nse != null && nse != sqle) {
-               checkException(nse);
+      if (!poolEntry.evict) {
+         String sqlState = sqle.getSQLState();
+         if (sqlState != null) {
```
brettwooldridge: Kill the 'if' on 107 above and change 109 to:
```java
if (!poolEntry.evict && sqlState != null) {
```

nitincchauhan: umm lost it ! why call getSQLState() ?

brettwooldridge: Combining the two conditions avoids the extra indentation.  This is a low frequency call and the overhead of getSQLState() is likely nanoseconds.

nitincchauhan: I think it is not worth to create var and call method, however negligible or infrequent it is for indention of 10 line method.

brettwooldridge:
@nitincchauhan It's actually a pointless argument.  I started benchmarking that method when a obvious truth hit me in the face...

As I said back on 2014/12/05, *"That is optimizing for the 0.5% case ...  and paying a penalty 99.5% of the time".*

``SQLExceptions`` are extremely common things.  ``SQLExceptions`` *with disconnection codes* are **extremely** rare ... like I can count on two hands how many times I've encountered them in the past 15 years.

Checking the ``evict`` flag -- an object dereference, a volatile access, a comparison, and a jump -- for *every* ``SQLException`` performs worse 99.5% of the time (realistically probably 99.995%), and performs better 0.5% of the time.

nitincchauhan:
closed change to closeOpenStatements() too !?

brettwooldridge:
Absolutely.  How is any different?

The new code pays the penalty in ``close()`` -- an object dereference (*poolEntry*), a volatile access (*evict*), and boolean negation -- and why?  In order to optimize ``close()`` for the eviction case, when 99.995% of the time ``close()`` is called *without* evicted connections.

The current code, ``closeOpenStatements()``, never dereferences *poolEntry* and never accesses the *evict* volatile, except in the rare exception case.  That is correct.  The common path is fast, and only the exceptional path (literally) pays the penalty.

PR #395, opened 2015-08-25
cleanup

nitincchauhan:
set lastAccess only if rolled back
method and var renamed,  closeStatements() moved out of try, already catching SQLExceptions
remove resetPoolEntry(), calling individual method if changed
unintentional logging

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
@@ -270,6 +270,7 @@ public final class PoolElf
 
       if (poolEntry.networkTimeout != networkTimeout) {
          setNetworkTimeout(poolEntry.connection, networkTimeout);
+         poolEntry.setNetworkTimeout(networkTimeout);
          resetBits |= 0b10000;
       }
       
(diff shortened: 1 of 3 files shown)


src/main/java/com/zaxxer/hikari/proxy/ConnectionProxy.java
```diff
@@ -177,14 +177,13 @@ public final void close() throws SQLException
    {
       if (delegate != ClosedConnection.CLOSED_CONNECTION) {
          leakTask.cancel();
+         closeStatements();
```
brettwooldridge: Why this move?  If closeStatements() throws a RuntimeException, or some subclass of it, the finally will not be executed, resulting in the connection not being released back to the pool.

nitincchauhan: done

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -245,27 +245,32 @@ void resetConnectionState(final PoolBagEntry poolEntry) throws SQLException
...
          poolEntry.connection.setCatalog(catalog);
+         poolEntry.setCatalog(catalog);
          resetBits |= 0b01000;
       }
 
       if (poolEntry.networkTimeout != networkTimeout) {
          setNetworkTimeout(poolEntry.connection, networkTimeout);
+         poolEntry.setNetworkTimeout(networkTimeout);
          resetBits |= 0b10000;
       }
```
brettwooldridge: I like this set of changes.

PR #396, opened 2015-08-27
do not set connection with property that user have not specified.

nitincchauhan:
auto commit and read only should be treated similarly too. do not set explicitly, let it be driver default.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
@@ -186,8 +186,14 @@ public final class PoolElf
       if (transactionIsolation == -1) {
          transactionIsolation = connection.getTransactionIsolation();
       }
-      else if (transactionIsolation != Connection.TRANSACTION_NONE) {
-         connection.setTransactionIsolation(transactionIsolation);
+      else {
+         int defaultTILevel = connection.getTransactionIsolation();
+         if (transactionIsolation != defaultTILevel) {
+            transactionIsolation = defaultTILevel;
+         }
+         if (transactionIsolation != Connection.TRANSACTION_NONE) {
+            connection.setTransactionIsolation(transactionIsolation);
+         }
       }
 
       if (catalog != null) {
--- a/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
@@ -183,9 +183,10 @@ public final class PoolElf
       connection.setAutoCommit(isAutoCommit);
       connection.setReadOnly(isReadOnly);
 
-      final int defaultLevel = connection.getTransactionIsolation();
-      transactionIsolation = (transactionIsolation < 0 ? defaultLevel : transactionIsolation);
-      if (transactionIsolation != defaultLevel) {
+      if (transactionIsolation == -1) {
+         transactionIsolation = connection.getTransactionIsolation();
+      }
+      else if (transactionIsolation != Connection.TRANSACTION_NONE) {
          connection.setTransactionIsolation(transactionIsolation);
       }
 
--- a/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolElf.java
@@ -183,17 +183,12 @@ public final class PoolElf
       connection.setAutoCommit(isAutoCommit);
       connection.setReadOnly(isReadOnly);
 
-      if (transactionIsolation == -1) {
-         transactionIsolation = connection.getTransactionIsolation();
-      }
-      else {
-         int defaultTILevel = connection.getTransactionIsolation();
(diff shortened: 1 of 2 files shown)


src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -183,9 +183,10 @@ void setupConnection(final Connection connection, final long connectionTimeout)
       connection.setAutoCommit(isAutoCommit);
       connection.setReadOnly(isReadOnly);
 
-      final int defaultLevel = connection.getTransactionIsolation();
```
brettwooldridge: This gets the connection default isolation...

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -183,9 +183,10 @@ void setupConnection(final Connection connection, final long connectionTimeout)
...
-      transactionIsolation = (transactionIsolation < 0 ? defaultLevel : transactionIsolation);
```
brettwooldridge: This sets the ``transactionIsolation`` member variable to the user specified isolation level, *or if none was specified, sets it to the connection default isolation*...

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -183,9 +183,10 @@ void setupConnection(final Connection connection, final long connectionTimeout)
...
-      if (transactionIsolation != defaultLevel) {
```
brettwooldridge: *Only if* the user specified level was different than the default, we set the connection isolation level.  If the user did not specifiy, an isolation, this is always false, and we don't set anything.

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -183,9 +183,10 @@ void setupConnection(final Connection connection, final long connectionTimeout)
...
+      if (transactionIsolation == -1) {
+         transactionIsolation = connection.getTransactionIsolation();
+      }
```
brettwooldridge: This block is fine in theory...

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -183,9 +183,10 @@ void setupConnection(final Connection connection, final long connectionTimeout)
...
+      else if (transactionIsolation != Connection.TRANSACTION_NONE) {
```
brettwooldridge: ...but even if the user specifies ``Connection.TRANSACTION_NONE`` we would need to call the setter.

Additionally, this code is *less efficient* because the code on the left only calls ``setTransactionIsolation()`` if the user specified level is different than the default.  *This* code will always call ``setTransactionIsolation()``, even if the user specified level is the same as the connection default.

nitincchauhan: http://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#setTransactionIsolation(int)
specially notes not to call method with TRANSACTION_NONE

nitincchauhan: updated code. not calling method when config is same as default

brettwooldridge: >
> http://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#setTransactionIsolation(int)
> specially notes not to call method with TRANSACTION_NONE


That convinces me further that the original code was more correct.  The
fact that a driver *can* return TRANSACTION_NONE from
*getTransactionIsolation()* is an indication by that driver that
transactions are not supported.

We therefore *must* *always* call *getTransactionIsolation()*, and even if
the user has specified READ_COMMITTED or some other isolation level, that
must be ignored and *transactionIsolation* should be set to
TRANSACTION_NONE.

This existing code is correct, except that this line should read:

transactionIsolation = (transactionIsolation < 0 || defaultLevel ==
Connection.TRANSACTION_NONE) ? defaultLevel : transactionIsolation);

nitincchauhan: done :)

nitincchauhan: just like this and catalog, hope to avoid calling setReadOnly and setAutoCommit if not set in config.

nitincchauhan: but this test fails !?
com.zaxxer.hikari.ConnectionStateTest.testTransactionIsolation(ConnectionStateTest.java:58)

brettwooldridge:
You need to update [StubConnection](https://github.com/brettwooldridge/HikariCP/blob/dev/src/test/java/com/zaxxer/hikari/mocks/StubConnection.java#L53) to have a default isolation that is not 0 (TRANSACTION_NONE).

nitincchauhan:
done

brettwooldridge:
As always, thank you.

PR #403, opened 2015-08-31
simplified PoolBagEntry creation, removed PoolElf#resetPoolEntry()

nitincchauhan:
could not fix TestConcurrentBag, I am not sure what to pass as parameters there

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -494,9 +494,9 @@ public class HikariPool implements HikariPoolMXBean, IBagStateListener
          String password = config.getPassword();
 
          connection = (username == null) ? dataSource.getConnection() : dataSource.getConnection(username, password);
-         poolElf.setupConnection(connection, connectionTimeout);
+         PoolBagEntry pbe = poolElf.createBagEntry(connection, connectionTimeout, this);
 
-         connectionBag.add(new PoolBagEntry(connection, this));
+         connectionBag.add(pbe);
          lastConnectionFailure.set(null);
          LOGGER.debug("{} - Added connection {}", poolName, connection);
 
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java
@@ -68,17 +68,21 @@ public final class PoolBagEntry implements IConcurrentBagEntry
       DATE_FORMAT = new SimpleDateFormat("MMM dd, HH:mm:ss.SSS");
    }
 
-   public PoolBagEntry(final Connection connection, final HikariPool pool)
+   public PoolBagEntry(final Connection connection, final HikariPool pool, final boolean isReadOnly, final String catalog, final boolean isAutoCommit, final int networkTimeout, final int transactionIsolation)
    {
       this.connection = connection;
       this.parentPool = pool;
+      this.isReadOnly = isReadOnly;
+      this.catalog = catalog;
+      this.isAutoCommit = isAutoCommit;
+      this.networkTimeout = networkTimeout;
+      this.transactionIsolation = transactionIsolation;
+
       this.creationTime = System.currentTimeMillis();
       this.poolElf = pool.poolElf;
       this.state = new AtomicInteger(STATE_NOT_IN_USE);
       this.lastAccess = ClockSource.INSTANCE.currentTime();
       this.openStatements = new FastList<>(Statement.class, 16);
-
-      poolElf.resetPoolEntry(this);
 
       final long maxLifetime = pool.config.getMaxLifetime();
       final long variance = maxLifetime > 60_000 ? ThreadLocalRandom.current().nextLong(10_000) : 0;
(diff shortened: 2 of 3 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -494,9 +494,9 @@ private boolean addConnection()
          String password = config.getPassword();
 
          connection = (username == null) ? dataSource.getConnection() : dataSource.getConnection(username, password);
-         poolElf.setupConnection(connection, connectionTimeout);
+         PoolBagEntry pbe = poolElf.createBagEntry(connection, connectionTimeout, this);
 
-         connectionBag.add(new PoolBagEntry(connection, this));
+         connectionBag.add(pbe);
```
brettwooldridge: This is *less* readable to me.  On the right I can clearly see that the connection is being setup.  With this change, that action becomes a side-effect of creating a bag entry, and is non-obvious to the reader.

src/main/java/com/zaxxer/hikari/pool/PoolBagEntry.java
```diff
@@ -68,18 +68,22 @@
       DATE_FORMAT = new SimpleDateFormat("MMM dd, HH:mm:ss.SSS");
    }
 
-   public PoolBagEntry(final Connection connection, final HikariPool pool)
+   public PoolBagEntry(final Connection connection, final HikariPool pool, final boolean isReadOnly, final String catalog, final boolean isAutoCommit, final int networkTimeout, final int transactionIsolation)
```
brettwooldridge: This is getting close to violating standards on the [number of parameters](http://checkstyle.sourceforge.net/config_sizes.html#ParameterNumber).

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -171,8 +171,9 @@ else if (jdbcUrl != null && dataSource == null) {
     * @param connection a Connection
     * @param connectionTimeout the connection timeout
     * @throws SQLException thrown from driver
+    * @return PoolBagEntry
     */
-   void setupConnection(final Connection connection, final long connectionTimeout) throws SQLException
+   PoolBagEntry createBagEntry(final Connection connection, final long connectionTimeout, HikariPool pool) throws SQLException
```
brettwooldridge: Finally, now PoolElf has turned into a factory for PoolBagEntry objects?

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: true. but I am sure it is not that difficult to understand either having pbe defined right above the line.
this change removes juggling of method calls while creating bag entry with one one less method

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
```diff
@@ -281,15 +284,6 @@ void resetConnectionState(final PoolBagEntry poolEntry) throws SQLException
...
 
-   void resetPoolEntry(final PoolBagEntry poolEntry)
-   {
-      poolEntry.setReadOnly(isReadOnly);
-      poolEntry.setCatalog(catalog);
-      poolEntry.setAutoCommit(isAutoCommit);
-      poolEntry.setNetworkTimeout(networkTimeout);
-      poolEntry.setTransactionIsolation(transactionIsolation);
-   }
-
```
brettwooldridge: All that for the purpose of removing ``resetPoolEntry()``?

This is a lot of change, that makes the code harder to follow, move responsibilities to unobvious places, and is ultimately in such a low traffic path  it likely gets executing less than once per minute.

nitincchauhan: To me it is much simpler now, earlier code was HARD to follow having call to resetPoolEntry() from constructor and that method was calling methods on self to set each member.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: What I mean is, on the left I clearly see a method ``setupConnection()``, which clearly indicates what it does.

With the ``createBagEntry()`` call, I have no indication that it might execute SQL or otherwise change the connection state.  It "appears" to simply create a ``PoolBagEntry``.

Lastly, passing ``connectionTimeout`` into ``createBagEntry()`` make it appear that it is a part of the bag entry *state*.  However, it is only used during connection setup, which is crystal clear in the code on the left.

src/main/java/com/zaxxer/hikari/pool/PoolElf.java
brettwooldridge: I will grant that ``resetPoolEntry()`` is not clear.  When naming the method, it was originally was called from more than one place.

I would suggest reverting all of the changes above, and renaming this method to something like ``initializeDefaultState()`` or ``configureStateFromDefaults()``, ...

The ``PoolElf`` holds all of the pool defaults derived from the ``HikariConfig`` and after determining supported APIs, etc.  That is why it is responsible for ``resetConnectionState(PoolBagEntry)`` as well as ``resetPoolEntry(PoolBagEntry)``.  While the former is clearly named, the later is clearly not.

nitincchauhan: close this please. sorry for your time.

nitincchauhan: BTW renaming method is not going to solve mess of back and forth method calls.
Thanks

PR #408, opened 2015-09-03
create and collect less garbage with little speed as bonus

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -634,11 +634,13 @@ public void run()
          if (idleTimeout > 0L) {
             logPoolState("Before cleanup\t");
             final List<PoolBagEntry> bag = connectionBag.values(STATE_NOT_IN_USE);
-            int removable = bag.size() - config.getMinimumIdle();
-            for (PoolBagEntry bagEntry : bag) {
+            final int bagSize = bag.size();
```
brettwooldridge: Because of JIT optimizations, the iterator form of ``for`` is [faster](https://aliteralmind.wordpress.com/2014/03/22/for_foreach/).

src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
```diff
@@ -245,13 +245,14 @@ public void close()
...
    public List<T> values(final int state)
    {
-      final ArrayList<T> list = new ArrayList<>(sharedList.size());
-      for (final T reference : sharedList) {
-         if (reference.getState() == state) {
-            list.add(reference);
+      final int size = sharedList.size();
+      final ArrayList<T> list = new ArrayList<>(size);
+      for (int i = 0; i < size; i++) {
+         T entry = sharedList.get(i);
```
brettwooldridge: This is inherently dangerous.  ``sharedList`` is a ``CopyOnWriteArrayList`` which is *safe for iteration* even while it is modified.  However, once you abandon the iterator, it is possible for the ``sharedList`` size to decrease while you are in the loop, and ``sharedList.get(i)`` will generate an ``ArrayIndexOutOfBounds`` exception.

src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
```diff
@@ -321,9 +322,10 @@ public int getPendingQueue()
     */
    public int getCount(final int state)
    {
-      int count = 0;
-      for (final T reference : sharedList) {
-         if (reference.getState() == state) {
+      int count = 0, size = sharedList.size();
+      for (int i = 0; i < size; i++) {
+         T entry = sharedList.get(i);
```
brettwooldridge: Same as above.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -634,11 +634,13 @@ public void run()
...
+            int removable = bagSize - config.getMinimumIdle();
+            for (int i = 0; i < bagSize; i++) {
                if (removable <= 0) {
                   break;
                }
+               final PoolBagEntry bagEntry = bag.get(i);
```
brettwooldridge: Additionally, with the iterator the bounds are only checked once, but with the ``get(index)`` method the bounds are checked with every invocation.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: 'for-each' is nice to read. and its faster 'only' for list with no random access.

src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
nitincchauhan: for-each throws ConcurrentModificationException see  item 7 here https://today.java.net/pub/a/today/2006/11/07/nuances-of-java-5-for-each-loop.html

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: negligible difference if at all.
but with for-each, GC will make difference in tight situation for creating and collecting iterator

nitincchauhan: other side of coin here http://howtodoinjava.com/2013/03/26/performance-comparison-of-different-for-loops-in-java/#comment-866

src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
brettwooldridge: The iterator of *CopyOnWriteArrayList* explicitly does not throw, that is why it was selected as a collection.

PR #419, opened 2015-09-15
Mediator: simplified check of validation modes

nitincchauhan:
ConnectionProxy: made memebers private, added missing methods of
connectionState. not sure why ide does not complain !

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -59,7 +59,6 @@ import com.zaxxer.hikari.util.ClockSource;
 import com.zaxxer.hikari.util.ConcurrentBag;
 import com.zaxxer.hikari.util.ConcurrentBag.IBagStateListener;
 import com.zaxxer.hikari.util.DefaultThreadFactory;
-import com.zaxxer.hikari.util.PropertyElf;
 
 /**
  * This is the primary connection pool class that provides the basic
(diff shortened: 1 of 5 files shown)


nitincchauhan:
IMHO, The pattern mediator itself is considered bad and tempts people easy.
I feel it is code smell with 3 methods ('Mediators accessors' as you call it) returning 'this' 
HikariPool.java has member of type PoolEntryMediator and initialised as this.entryMediator = mediators.getConnectionStateMediator(); reads odd !
I hope this would be ironed out with another refactor.

brettwooldridge:
At a minimum the ``getConnectionStateMediator()`` should be renamed ``getPoolEntryMediator()``.

Many of the code smell issues would be cleaned up if Java provided cross-package exports (one of the goals of JDK 9 Project Jigsaw).  The reality is that there is a fairly intimate relationship between the *pool* package and the *proxy* package.  However, putting them in the same package leads to its own separation of concerns issues.

Putting them in different packages (as they are now) has led to forcing classes to be *public* that should be *package* scope.  And has led to circular dependencies between packages -- *some of which* are cleaned up in this refactor.

I, of course, will continue to experiment with the code structure, possibly including merging the two packages.  Lastly, some of the structural concessions end up being concessions in favor of performance.  Whether those can ever be completely removed remains to be seen.

brettwooldridge:
On a final note, the refactor was prompted by your original pull request that moved creation of the PoolEntry objects into the ``PoolElf`` (now ``Mediator``).  In reality, the mediator pattern was always there, but the responsibilities of the mediator was never called out.  Defining interfaces that the ``Mediator`` implements makes the responsibilities explicit.

The ``PoolEntry`` used to use ``PoolElf.someMethod()`` in several places.  This included easy invocability of ``PoolElf.registerMBeans()`` or ``PoolElf.isConnectionAlive()`` ... something it has no business doing.  Now, it is passed a ``PoolEntryMediator`` instance, and the methods it can invoke are limited to what it *should* invoke.

Of course, in the case of static methods, any developer *can* go around the mediator interface and invoke against the Mediator implementation class directly, but that would hopefully raise a flag in their brain that they were violating a boundary.

nitincchauhan:
Thanks for insight and continue good work :+1:

Issue #412, opened 2015-09-07
PoolBagEntry sensitive fields are exposed through MetricsContext


brenuart:
Method `MetricsContext.setConnectionLastOpen()` is passed a `PoolBagEntry` instance. Most its fields are public meaning any custom `MetricsContext` implementation has read/write access to them. Unfortunately, most of those fields are very sensitive and are not intended to be exposed to elements outside of Hikari internals.

As far as I can tell, `MetricsContext` only needs to update the `lastOpenTime` field - access to the other properties could remain read-only. More over, it is questionable whether `MetricsContext` is responsible to update that field or if Hikari could take care of it.

brettwooldridge:
@brenuart Already working on a seperate refactor that covers this... thanks.

brettwooldridge:
This is fixed in 2.4.2-SNAPSHOT, but as the CHANGELOG states:
```
Ongoing com.zaxxer.hikari.metrics refactors. This is not considered public API until such time as we announce it. Caveat lector.
```

PR #423, opened 2015-09-17
removed redundant creationTime. lastAceess is initialized at creation as

nitincchauhan:
well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,7 +360,6 @@ Here is a list of JDBC *DataSource* classes for popular databases:
 | Apache Derby     | Derby        | org.apache.derby.jdbc.ClientDataSource |
 | Firebird         | Jaybird      | org.firebirdsql.pool.FBSimpleDataSource |
 | H2               | H2           | org.h2.jdbcx.JdbcDataSource |
-| IBM AS400        | IBM          | com.ibm.as400.access.AS400JDBCDriver |
 | HSQLDB           | HSQLDB       | org.hsqldb.jdbc.JDBCDataSource |
 | IBM AS400        | IBM          | com.ibm.as400.access.AS400JDBCDriver |
 | IBM DB2          | DB2          | com.ibm.db2.jcc.DB2SimpleDataSource |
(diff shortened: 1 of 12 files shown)


brettwooldridge:
I am particularly interested in seeing the ``creationTIme`` (via toString()), independent of the ``lastAccess`` time.  That's why it is there.

nitincchauhan:
in Log it would show creation time formatted, besides that there is no 'real' use.

brettwooldridge:
That *is* the real use.  It is there for my troubleshooting, because there were several bugs in the past about connections not being retired.  It should not be removed.

nitincchauhan:
log also shows when connection was last accessed. how creation time would be more useful than this?

nitincchauhan:
now changed lastOpenTime to update on borrow so you may use it as well instead.

brettwooldridge:
There was some trickery going on, for performance, that probably wasn't apparent...

[This code](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L176):
```java
metricsContext.setConnectionLastOpen(poolEntry, now);
```
calls [this method](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/metrics/MetricsTracker.java#L69) when there are no metrics being recorded:
```java
public void setConnectionLastOpen(final PoolEntry bagEntry, final long now) {
     // do nothing
}
```
which you can see actually does *nothing*.  Only when metrics are *enabled* does ``poolEntry.lastOpenTime`` get set in the ``CodaHaleMetricsTracker`` [version of that method](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/metrics/CodaHaleMetricsTracker.java#L130).

That avoids a *volatile* access in the "hot path" when metrics are off.  Here is the difference in the benchmark between the current code and your pull:
```
Current:
Benchmark                       Mode  Cnt      Score     Error   Units
ConnectionBench.cycleCnnection  thrpt   48  29497.680 ± 396.660  ops/ms

Pull:
Benchmark                       Mode  Cnt      Score     Error   Units
ConnectionBench.cycleCnnection  thrpt   48  26988.262 ± 737.596  ops/ms
```

nitincchauhan:
I was aware of this trick  ;)
This change is first step to protect pool entry and its fields... With your help, Hope to achieve both performance as well as protection from this start.

nitincchauhan:
lastOpenTime is initialized only in getConnection() so it can be made non-volatile. it is ok to have old value in metrics if at all access happens with initialization exactly same time.

nitincchauhan:
Now with all these changes, I hope difference in benchmark would be acceptable.

nitincchauhan:
Also, as requested in https://github.com/brettwooldridge/HikariCP/issues/412, MetricsContext can not interfere with pool entry anymore.

PR #424, opened 2015-09-21
minor name changes.

nitincchauhan:


brettwooldridge:
Already in the middle of bunch of renames for these, I'll fold your changes in where it makes sense.

PR #433, opened 2015-09-26
separated release to pool & close connection EARLY by MOVING evicted

nitincchauhan:


brettwooldridge:
I feel like a broken record.  This code is optimizing for the 0.01% case (eviction) and penalizing the 99.9% case (non-eviction).

Without change:
```
Benchmark                       Mode  Cnt      Score     Error   Units
ConnectionBench.cycleCnnection  thrpt   48  30901.759 ± 714.116  ops/ms
```
With change:
```
Benchmark                       Mode  Cnt      Score     Error   Units
ConnectionBench.cycleCnnection  thrpt   48  28845.531 ± 814.553  ops/ms
```
Even without looking at the other changes, at a minimum the ``ProxyConnection.close()`` now involves two class member dereferences to ``poolEntry.evicted`` in the non-eviction path.

The code also looks messier.

And if an eviction occurs in ``closeStatements()``, ``delegate.rollback()``,  ``resetConnectionState()``, or ``delegate.clearWarnings()`` the connection will not be evicted until the next caller to ``getConnection()`` encounters it -- instead of immediately when the finally block calls ``recycle()``.

nitincchauhan:
>> As you assume, This change does not test evicted twice. It is "shifted" to beginning. benchmark does not look real.
>> probability of connection being evicted is much less during closing, than it is during its use before.

brettwooldridge:
Are you on Windows?  I can setup the benchmark so you can run it whenever you make changes.  I run the "long form" benchmark before I commit any changes in the hot path.

brettwooldridge:
Btw, it *does* test evicted twice.  On line 217 and 240 of your code.

nitincchauhan:
line 240 is as before, no change there.
in normal case, the code would only test evict once and if true it closes connection and 'returns' at line 220 never twice.

brettwooldridge:
In the normal case a connection is not evicted, right?  So no early return.  If the benchmark assumed every connection was evicted it would be a very strange benchmark indeed.

nitincchauhan:
change is NOT assuming every connection as broken. at present it is testing it too but too late in close proceedings.
I think MOVING test "early" is:
1 not disturbing "normal" operations or imposing any overhead
2 if evicted saves from "reset" operations and exceptions, particularly shutdown/suspend would benefit

nitincchauhan:
I agree that attempts before for this change were bad and disturbing.
your many refactor made this possible to implement, cleanly.

brettwooldridge:
@nitincchauhan I setup the benchmark for Windows for you.

https://github.com/brettwooldridge/HikariCP-benchmark

Run like:
```sh
benchmark.cmd long -p "pool=hikari" ".*Connection.*"
```
If your CPU has 4 cores and no hyper-threading, add: ``-t 4`` after the ``long`` option.

You can also replace ``long`` with ``quick`` or ``medium``, but when truly comparing a change before pull run with ``long``.  Note you should close all other programs (including Eclipse, browsers, etc.) and running benchmarks.  Kinda boring. :grin:

nitincchauhan:
@brettwooldridge Thanks for setup. I will run this soon I get some more time.

PR #438, opened 2015-10-01
removed isClosed() check

nitincchauhan:
no need for trip to driver/db for closing con.
http://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#close() states
Calling the method close on a Connection object that is already closed is a no-op.

moreover isClosed() was not guarded by network timeout.


brettwooldridge:
Reasonable.

PR #440, opened 2015-10-01
removed old call to close statements

nitincchauhan:


nitincchauhan:
something is not right 'still' :(

Failed tests: 
  TestConnectionTimeoutRetry.testConnectionIdleFill:239 Totals connections not as expected expected same:<10> was not:<7>

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -220,7 +223,6 @@ public final void close() throws SQLException
          leakTask.cancel();
 
          try {
-            closeStatements();
```
brettwooldridge: Good catch.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -198,6 +198,9 @@ private final void closeStatements()
             }
             catch (SQLException e) {
                checkException(e);
+               if (delegate == ClosedConnection.CLOSED_CONNECTION) {
+                  break; //connection closed in checkException
```
brettwooldridge: I know this is more efficient, but it is optimizing a case so rare that I really don't care about the performance.

The additional code just adds cognitive load to an already messy method.

brettwooldridge: Actually, it is infrequent enough for there to be unclosed Statements, especially with the prevalence of persistence frameworks (Hibernate, JPA, Play/Slick, etc.) that I'd be just as happy adding ``delegate != ClosedConnection.CLOSED_CONNECTION`` to the *for loop* conditional.

Issue #109, opened 2014-07-11
HikariCP and active connections in Postgresql


mantesat:
Hi all, Brett,

I'm using HikariCP for the past month, transitioning from BoneCP to remedy some leak connection problems. I'm using Hikari with Postgresql, Spring and Hibernate, and I'm printing some MBean statistics  about the active/idle connections. As reported by Hikari pool MBean, active connections slowly but steadily increase. However, when running 

select query, state, state_change from pg_stat_activity

to the postgresql, all connections are reported as idle.

I'm sure that when the active connections reach the max pool size, the application will freeze. Which one reports the correct status of the connections, Hikari or postgresql? Is there a way to see which query is running on the active connections from Hikari?

Thanks in advance
mantesat, Greece

mantesat:
Brett, apologies for re-posting to this after a long time, seems not to be resolved (although latest versions have considerably improved things). From statistics, it seems that slowly but steadily, active connections tend to increase.  I don't know if anyone has already reported such behavior, I'll try to use it from source code and find out why connections stay marked as active. Any suggestions are welcomed!

zsoca8711:
We have the same issue, since we have moved to Hikari on our postgres databases, we have frequent (once every 1-2 weeks) issues, the database runs out of connections. Restarting the tomcats using Hikari fixes the issue. 

Code sample:

```java
HikariConfig config = new HikariConfig(dcAccessDetails.getDetailsAsPropertiesForHikari());
    	config.setIdleTimeout(180000);
    	config.setMaxLifetime(1800000);
    	config.setInitializationFailFast(true);
    	config.setConnectionTestQuery("SELECT current_timestamp");
```
```java
public Properties getDetailsAsPropertiesForHikari(){
		Properties props = new Properties();
		props.setProperty("dataSourceClassName", DatabaseUtil.POSTGRES_DATASOURCE_CLASS);
		props.setProperty("dataSource.user", this.dcDbUser);
		props.setProperty("dataSource.password", this.dcDbPassword);
		props.setProperty("dataSource.databaseName", this.dcDbDatabaseName);
		props.setProperty("dataSource.serverName", this.dcDbServer);
		
		return props;
	}
```

prostobog:
I have a similar issue with Hibernate+Postgresql:
http://stackoverflow.com/questions/32968530/hikaricp-connection-is-not-available

amanakshe:
[...]

PR #444, opened 2015-10-06
align log with 'Before cleanup'

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -442,6 +442,7 @@ private boolean addConnection()
       // Speculative increment of totalConnections with expectation of success
       if (totalConnections.incrementAndGet() > config.getMaximumPoolSize()) {
          totalConnections.decrementAndGet(); // Pool is maxed out, so undo speculative increment of totalConnections
+         LOGGER.debug("{} - Cannot exceed maximum connections capacity: {}", poolName, config.getMaximumPoolSize());
```
brettwooldridge: This is likely to be a fire hose of logs, have you tested it?

nitincchauhan: not tested.
I thought it is worth logging in scenario while answering question at stackoverflow s/he linked it to https://github.com/brettwooldridge/HikariCP/issues/109 !)

nitincchauhan:
Also, There is something strange about this test failing regularly 'sometimes' :)

PR #448, opened 2015-10-08
using Throwable, better for backport too

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -99,11 +99,10 @@ public String toString()
 
    void quietlyCloseConnection(final Connection connection, final String closureReason)
    {
+      if (connection == null) {
+         return;
+      }
```
brettwooldridge: Insert newline after { ... before *try*

nitincchauhan:
again test failed :(
I think there is something more than timing problem there.

brettwooldridge:
It is a timing issue, and I'm pretty sure I understand the cause now...

nitincchauhan:
ok :)
wrt autocommit change:
IMO smart change in poolentry will reset state of proxy connection to real connection as:
Connection createProxyConnection(final ProxyLeakTask leakTask, final long now)  {
   ProxyConnection pc = ProxyFactory.getProxyConnection(this, connection, openStatements, leakTask, now);

    resetConnectionState(final ProxyConnection proxyConnection, final int dirtyBits)

    return pc;
}
where you know what to set dirtyBits ? Just an Idea

brettwooldridge:
I looked at that.  There is no way to use resetConnectionState() without also calling through to the actual connection and I didn't want to complicate the logic further.

PR #450, opened 2015-10-08
init 'other' vars too for proxy connection

nitincchauhan:


brettwooldridge:
The ``isAutoCommit`` has an *operational effect* in ``ProxyConnection.close()`` and therefore must be initialized properly.

However, initializing the other vars is not necessary as they are not used anywhere else, except to reset state, in which case they will have whatever value the user sets and their dirty bits will be set appropriately.

nitincchauhan:
in close() it should check read only before rollback :

    if (isCommitStateDirty && !isAutoCommit && !isReadOnly) {

brettwooldridge:
Agreed on *isReadOnly*.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -103,9 +102,18 @@ public final String toString()
...
 
    // ***********************************************************************
-   //                     Connection State Accessors
+   //                     Connection State init & getters
    // ***********************************************************************
 
+   final void init(boolean isReadOnly, boolean isAutoCommit, int networkTimeout, int transactionIsolation, String dbcatalog)
+   {
+      this.isReadOnly = isReadOnly;
+      this.isAutoCommit = isAutoCommit;
```
brettwooldridge: The next three do not need to be set.

nitincchauhan: ok

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
```diff
@@ -90,7 +90,9 @@ void setFutureEol(final ScheduledFuture<?> endOfLife)
 
    Connection createProxyConnection(final ProxyLeakTask leakTask, final long now)
    {
-      return ProxyFactory.getProxyConnection(this, connection, openStatements, leakTask, now, hikariPool.isAutoCommit());
+      ProxyConnection pc = ProxyFactory.getProxyConnection(this, connection, openStatements, leakTask, now);
+      hikariPool.initProxyConnection(pc);
```
brettwooldridge: I suspect that this is going to cause a performance hit, because this is in the hot path and the ``init()`` method in ``ProxyConnection`` introduces and additional dereference and method invocation.

I initially did it like this, but eventually introduced it as an additional stack parameter to the constructor to reduce the overhead.

brettwooldridge: EDIT: but with two members being initialized, maybe it comes out even.  I will have to benchmark before this merge.

nitincchauhan: it would need to duplicate readonly and autocommit in pool entry and touching parameter count check :) I hope jvm would inline these 'simple' methods.

nitincchauhan: You want me change to two parameters with two methods? I have it ready too

brettwooldridge: Let me benchmark it, and I'll let you know.  Almost midnight here, so probably tomorrow morning.

nitincchauhan:
calling methods back and forth is past for good :)
before: two method calls ( pool.Init(pc) and pc.init(1,2) )
now: two method calls (pool.getReadOnly and pool.getAutoCommit)
I liked later :)

PR #451, opened 2015-10-09
Race condition while adding threads

mtambara:
I'm sending this pull to notify you about and provide a potential solution for a bug in your code. There is a race condition when 2 threads call `HikariPool.getConnection()` while there are no idle connections available and `minimumIdle` is set to 0. Both threads will go into `ConcurrentBag.borrow(long, TimeUnit)`, see that there are no available threads to borrow, and create new worker threads to add one connection each in `HikariPool.addBagItem()`. The borrowing threads will then wait until signalled, at which point they will wake up and check for available connections again. The workers will execute sequentially and each try to create a new connection. This is the execution that leads to the race condition:

Worker 1 adds a connection as usual, signalling the borrow threads to wake up. 

Worker 1 finishes (`isDone()`)

Before the borrow threads can check for new connections, worker 2 tries to add a new connection.

Worker 2 checks if `getIdleConnections() <= minimumIdle`, sees the connection worker 1 added, and skips the `addConnection` call.

Borrow thread 1 sucessfully borrows a connection.

Borrow thread 2 cannot borrow a connection and goes to sleep waiting for Worker 2 to signal.

Worker 2 finishes (`isDone()`)

I have included a potential solution to this problem (signal at the end of the worker thread no matter what) and a test to reproduce this race condition (`AddConnectionRaceConditionTest`).

Let me know if you have any questions.

(cc @shuyangzhou)

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -304,6 +304,8 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
                quietlySleep(sleepBackoff);
                sleepBackoff = Math.min(connectionTimeout / 2, (long) (sleepBackoff * 1.5));
             }
+
+            connectionBag.getSynchronizer().signal();
          }
       }, true);
 
--- a/src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
+++ b/src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
@@ -309,6 +309,10 @@ public class ConcurrentBag<T extends IConcurrentBagEntry> implements AutoCloseab
       return synchronizer.getQueueLength();
    }
 
+   public QueuedSequenceSynchronizer getSynchronizer() {
+      return synchronizer;
+   }
+
    /**
     * Get a count of the number of items in the specified state at the time of this call.
     *
--- a/src/test/java/com/zaxxer/hikari/pool/AddConnectionRaceConditionTest.java
+++ b/src/test/java/com/zaxxer/hikari/pool/AddConnectionRaceConditionTest.java
@@ -1,0 +1,115 @@
+package com.zaxxer.hikari.pool;
+
+import java.lang.reflect.Field;
+
+import java.sql.Connection;
+
+import java.util.List;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+
+import org.junit.Test;
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+import com.zaxxer.hikari.mocks.StubConnection;
+import com.zaxxer.hikari.util.ConcurrentBag;
+
+/**
+ * @author Matthew Tambara (matthew.tambara@liferay.com)
+ */
+public class AddConnectionRaceConditionTest
+{
+   @Test
+   public void testRaceCondition() throws Exception {
+      HikariConfig config = new HikariConfig();
+      config.setMinimumIdle(0);
+      config.setMaximumPoolSize(10);
+      config.setInitializationFailFast(false);
+      config.setPoolName("Test Pool");
+      config.setDataSourceClassName("com.zaxxer.hikari.mocks.StubDataSource");
+
+      // Initialize HikariPool with no initial connections and room to grow
+
+      HikariDataSource ds = new HikariDataSource(config);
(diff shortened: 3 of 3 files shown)


shuyangzhou:
Hello @brettwooldridge,

A little background input, Matt and I work for Liferay, see our project at https://github.com/liferay/liferay-portal

We are currently pushing for our 7.0 release, which is a major release for us. We have added HikariCP as our jdbc pool support for a long time, and in 7.0 release we plan to make HikariCP as our default out of box jdbc pool.

However we ran into this issue lately. We really wish it can be fixed quickly so that we won't have to switch back to the old pool. Matt has created an unit to stably reproduce this race condition, and provided a potential solution. Please review this change and let us know if you need more input. If you are ok with solution and can push out bug fix release for this, that will be very helpful for us.

Thank you very much!

Shuyang

brettwooldridge:
@mtambara @shuyangzhou First, welcome Liferay. Thanks for reporting the issue.

It seems [we used to](https://github.com/brettwooldridge/HikariCP/blob/f6b0ee5ab947f86a4d0d465e976b993d8b3f619c/hikaricp/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L76) handle this differently, and [refactored it](https://github.com/brettwooldridge/HikariCP/blob/04a2234fec05dead8a561163825c1bf2750e1b41/hikaricp/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L75).

That refactor introduced a [different race condition](https://groups.google.com/d/msg/hikari-cp/x6d9OdKwgzw/Nt_2E0tcDXoJ) related to *minimumIdle=0*.  [Which was fixed](https://github.com/brettwooldridge/HikariCP/blob/43910e61dfee9fd69e8b82208c9f76bc05ecfe84/hikaricp-common/src/main/java/com/zaxxer/hikari/pool/BaseHikariPool.java#L413).

Before merging your change, I'd like to explore a few more options.  The only reason I'm not crazy about it is the leaking of control out of the ``ConcurrentBag`` ... just a separation of concerns ... concern.  At the same time, it may be the simplest solution, which itself carries a lot of merit.

What is the time horizon for 7.0?  Even if we committed this today, a release would need to go through a release candidate first.  As it happens, just two days ago we changed our version from 2.4.2-SNAPSHOT to 2.4.2-RC1 in preparation for a release, but we have not yet pulled the trigger.

shuyangzhou:
@brettwooldridge Thanks for your quick response!

We are aiming the end of this year or early next year, so there is still time for it. However this issue is affecting our CI system, we have a large farm for CI testing, and this issue is causing tons of machines to randomly fail tests. If the fix can not be merged soon, we can patch the HikariCP jar on our side for now, as long as we know this will be fixed before our 7.0 release.

We basically just have to run tests for long enough time on our testing system to be certain that there is no more random failures, so if this fix can at least go in 1~2 months before our release, that will give us enough time to feel confident with the stability.

Thanks @brettwooldridge, please take your time, and keep us updated with the progress of this issue.

brettwooldridge:
@shuyangzhou Certainly we will have this issue fixed within days (if not hours).  We do plan to publish an RC, with a likely release at the end of October.

With respect to your CI builds, would a temporary workaround be to set minimumIdle to a non-zero value?

shuyangzhou:
Yes, we can do that, just that only lower the chance of seeing this race condition, not eliminating it. Our CI system is large, and running all the time, it is just a matter of time to see it happens again.

We concurrently have our failing tests disabled. If we are talking about a bug fix release by the end of this month, that will be totally acceptable for us. We also have a half-automated patching system, so patching is very simple for us. We can just do our patch for now, then drop it when you get your release out.

Thanks @brettwooldridge , please kindly notify us, if you choose to use a different solution for the fix, we can adjust our patch to do it your way, that way we won't run into some random surprise at the time we switch to your official release.

brettwooldridge:
@mtambara @shuyangzhou Just so you know, the proposed patch does not fix the issue, it merely makes it less likely.  I have rewritten the test using a loop that starts two threads that each acquire then evict a connection.  Then wait for each to complete or encounter an exception, and if no exception, loop again.

After a few seconds (couple of thousand iterations) the race is encountered.

shuyangzhou:
Ugh, that's a bummer. @brettwooldridge will you start to work for a solution on your side?
We just need to plan out and be prepared for our release. Thanks!

brettwooldridge:
@shuyangzhou We're looking at it on our side, and expect a fix within a day or two.

shuyangzhou:
That sounds great, thanks @brettwooldridge !

PR #453, opened 2015-10-09
remove method calls from hot path and sleep in smaller steps

nitincchauhan:


nitincchauhan:
@brettwooldridge 
beside this: as close() should be no-op for already closed connection, class ClosedConnection should have ?

    if ("close".equals(methodName)) { 
       return Void.TYPE;
   }

PR #454, opened 2015-10-09
calling closedconnection.close() should not throw exception.

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -455,11 +455,14 @@ private static Connection getClosedConnection()
             public Object invoke(Object proxy, Method method, Object[] args) throws Throwable
             {
                final String methodName = method.getName();
-               if ("abort".equals(methodName)) { 
+               if ("isValid".equals(methodName)) {
+                  return Boolean.FALSE;
+               }
+               else if ("close".equals(methodName)) {
```
brettwooldridge: How can ``close()`` ever be called on the ``ClosedConnection`` class?  The ``close()`` method of the ``ProxyConnection`` [already checks](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L227).

nitincchauhan: in case someone does:
proxycon = ds.getConnection();
proxycon.close()
proxycon.unwrap().close()

brettwooldridge: Ah, there was a code path through [checkException()](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L155) ... the solution was to switch the evict() and the delegate reassignment.

Committed.

You can roll back this particular change.

brettwooldridge: proxycon.unwrap().close() is not valid in any pool.

nitincchauhan: its common to unwrap and call close. let pool handle this case gracefully.
wrt checkException(), I have not made any change there in this pool request.

brettwooldridge: Why on earth would someone unwrap and call close?  BTW, they *can* call ``proxycon.unwrap().close()``.  But they cannot call:
```java
proxycon.close()
proxycon.unwrap().close()
```
This is a "use after return" to pool.  In almost every other pool the ``proxycon.unwrap().close()`` call would close the Connection *after a new thread has acquired it*!

I would actually prefer for HikariCP to fail in spectacular fashion with lots of fireworks in this case.  If you want to implement this method as:
```java
if ("close".equals(methodName)) { 
   throw new IllegalStateException("WTF are you doing?!");
}
```
then I have no objection.

nitincchauhan: not 'literally' consecutive calls like close after close.
people use 'api' in ways author not intended for :)

It may happen... for some reason... unwrapped connection be held somewhere in code then later some time close would be called on it.

brettwooldridge: Neither ``isClosed()`` nor ``isValid()`` implies an alteration of state, they are read-only state accessors.  The ``isValid()`` method is there to cover failures at pool shutdown when connections are being closed even as the pool is still being called by application threads.

Again, if an application holds onto an *unwrapped* Connection and tries to close it *anytime* after ``close()`` has been called to return the Connection to the pool, it is an application error.  If I were a user, I would certainly *want* a failure to occur.  I would probably fire the engineer who wrote it.

brettwooldridge: Actually, I stand corrected.  The only way to get this delegate is to call ``unwrap()`` *after* already closing the connection.

We intercept ``isValid()`` because it is supposed to return ``false`` on closed connections.

nitincchauhan: undone, saved pity engineer :)

nitincchauhan:
this still have fix for #456, removed redundent call and minor cleanup

PR #459, opened 2015-10-14
removed check (claimed != null) from hotpath

nitincchauhan:
replaced old code now that you are done with fix


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -567,7 +566,7 @@ public Boolean call() throws Exception
          long sleepBackoff = 200L;
          do {
             final PoolEntry poolEntry = createPoolEntry();
-            if (poolEntry == MAXED_POOL_MARKER) {
+            if (poolEntry == PoolEntry.MAXED_POOL_MARKER) {
```
brettwooldridge: I prefer to keep the static import.

nitincchauhan: both static of pool entry is used like this in other places

PR #460, opened 2015-10-15
commit initSql if isIsolateInternalQueries is also true

nitincchauhan:


brettwooldridge:
Seems a little risky, but I'll accept it unless/until someone reports an issue.

PR #463, opened 2015-10-18
cleanup

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -69,10 +69,10 @@
 
    private static final ClockSource clockSource = ClockSource.INSTANCE;
 
-   private final long ALIVE_BYPASS_WINDOW_MS = Long.getLong("com.zaxxer.hikari.aliveBypassWindowMs", TimeUnit.MILLISECONDS.toMillis(500));
-   private final long HOUSEKEEPING_PERIOD_MS = Long.getLong("com.zaxxer.hikari.housekeeping.periodMs", TimeUnit.SECONDS.toMillis(30));
+   private static final long ALIVE_BYPASS_WINDOW_MS = Long.getLong("com.zaxxer.hikari.aliveBypassWindowMs", TimeUnit.MILLISECONDS.toMillis(500));
+   private static final long HOUSEKEEPING_PERIOD_MS = Long.getLong("com.zaxxer.hikari.housekeeping.periodMs", TimeUnit.SECONDS.toMillis(30));
```
brettwooldridge: These two members were *explicitly* made non-``static``.  These two values being *static* was responsible for 99% of the junit test failures.

These values are set during various junit tests, and junit runs test classes and methods in random order.  So, if they are *static* then whichever test happens to run first "wins" -- the values cannot be reassigned by another test for the lifetime of the JVM.

nitincchauhan:
done.
side question:
when ConcurrentBag#add() throws exception, connection is NOT added in the list. in that case where / when does the original connection gets closed ?

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: done.
side  question:
when ConcurrentBag#add() throws exception, connection is NOT added in the list. in that case where / when does the original connection gets closed ?

nitincchauhan:
done.
side  questions:
1 when ConcurrentBag#add() throws exception, connection is NOT added in the list. in that case where / when does the original connection gets closed ?
2 if called from housekeeper or eviction by user, soft evict connection must cancel eol and leak detection tasks too.

PR #484, opened 2015-11-02
if minimumIdle > maxPoolSize, no need to set maxPoolSize too

nitincchauhan:


src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -787,10 +787,14 @@ private void validateNumerics()
          validationTimeout = connectionTimeout;
       }
 
-      if (minIdle < 0 || minIdle > maxPoolSize) {
+      if (minIdle < 0) {
          minIdle = maxPoolSize;
       }
 
+      if (minIdle > maxPoolSize) {
```
brettwooldridge: this can be an ``else if...``

nitincchauhan:
done

PR #489, opened 2015-11-09
exit loop after reaching max iteration.

nitincchauhan:
avoids creating queue of sleeping callables in case of network outage

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -568,8 +568,8 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
             }
             final PoolEntry poolEntry = createPoolEntry();
             if (poolEntry != null) {
-               connectionBag.add(poolEntry);
                totalConnections.incrementAndGet();
+               connectionBag.add(poolEntry);
                return Boolean.TRUE;
             }
 
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -567,21 +567,23 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       @Override
       public Boolean call() throws Exception
       {
-         long sleepBackoff = 200L;
+         long sleepBackoff = 200L, min = 0L;
          do {
+            min = sleepBackoff;
             final PoolEntry poolEntry = createPoolEntry();
-            if (poolEntry == MAXED_POOL_MARKER) {
-               return Boolean.FALSE;
-            }
-            else if (poolEntry != null) {
+            if (poolEntry != null) {
                connectionBag.add(poolEntry);
                return Boolean.TRUE;
+            }
+            else if (poolEntry == MAXED_POOL_MARKER) {
+               return Boolean.FALSE;
             }
 
             // addConnection() failed, so we sleep and retry
             quietlySleep(sleepBackoff);
             sleepBackoff = Math.min(connectionTimeout / 2, (long) (sleepBackoff * 1.3));
-         } while (true);
+         } while (min < sleepBackoff);
+         return Boolean.FALSE;
       }
    }
 
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -567,7 +567,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       @Override
       public Boolean call() throws Exception
       {
-         long sleepBackoff = 200L, min = 0L;
+         long sleepBackoff = 200L, maxSleep = 0L;
          do {
             final PoolEntry poolEntry = createPoolEntry();
             if (poolEntry != null) {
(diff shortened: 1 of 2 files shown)


nitincchauhan:
Also it is time to find better solution for if (weakThreadLocals) checks in ConcurrentBag !

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -567,17 +567,23 @@ protected void update() {
...
       {
-         long sleepBackoff = 200L;
+         long sleepBackoff = 200L, maxSleep = 0L;
          do {
             final PoolEntry poolEntry = createPoolEntry();
-            if (poolEntry == MAXED_POOL_MARKER) {
-               return Boolean.FALSE;
-            }
-            else if (poolEntry != null) {
+            if (poolEntry != null) {
```
brettwooldridge: How can this work?  This is going to insert ``MAXED_POOL_MARKER`` into the bag.

brettwooldridge:
I am suspicous of this change.  Basically, the Callable should *never* stop trying ... even or *especially* in the case of a network outage.  Additionally, for example, if the pool is started while the database is down, and *initializationFailFast=false* the Callable keeps trying until such time as the database becomes available and then fills the pool.

nitincchauhan:
good thing is:
createPoolEntry() restricts requests (queue) to maximum pool size and then begins returning MAXED_POOL_MARKER :)

nitincchauhan:
I was wrong on requests being restricted to maximum pool size (createPoolEntry() before returning null decrements counter)
When db/network is ready, The next getConnection() or housekeeper would add new requests.
This would avoid long queue created by housekeeper/getConnection till db or network is down.

nitincchauhan:
fresh commit has better fix that restricts add requests to maximum pool size

nitincchauhan:
with this it is possible that total connections reported is not 'actual', it includes sleeping add requests

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -567,13 +559,15 @@ protected void update() {
       @Override
       public Boolean call() throws Exception
       {
+         // Speculative increment of totalConnections with expectation of success
+         if (totalConnections.incrementAndGet() > config.getMaximumPoolSize()) {
+            totalConnections.decrementAndGet(); // Pool is at max size
+            return Boolean.FALSE;
+         }
```
brettwooldridge: This check has to be inside of the ``do`` loop.

brettwooldridge: Nevermind, this is still a bad change.  Now the pool is going to sit, for example, with ``totalConnections=10`` even though no connections can be made to the database (because it is down).  That is incorrect.

nitincchauhan: Exactly what I said before 'total connections reported is not 'actual', it includes sleeping add requests'
needs some more workout to iron out even this flaw, your input is requested :)

nitincchauhan: fix commited

nitincchauhan: moved init and check inside loop

nitincchauhan:
removed extra var too

nitincchauhan:
wondering why tests fail with 'Internal accounting inconsistency, totalConnections=-1'
probably hidden bug exposed or is it new?
It increments totalConnections only after successful creation of connection and adding it to bag.
decrements only in close() why would it go negative?
earlier (commit with ''decrement not required'), tests passed with additional var. If explained, I can revert to that version

brettwooldridge:
Here is my biggest issue with this change, I don't see the benefit.  It is basically just moving proven reliable code around.

Maybe it is possible to fix the totalConnections issue, but doing so will increase the complexity of the code.  Again, for what purpose?

nitincchauhan:
to restrict add connection requests to maximum pool size, this will eliminate number of requests sleeping in time of db / network down. The change is for better than proven code.

brettwooldridge:
The issue you think you are solving is not an issue.  The executor is single threaded.  So whether there are 10 items in the queue or only 2, there is only one that is sleeping.  The queue only allows maxPoolSize items anyway, and the new code doesn't change that.  The second Callable in the queue will not execute until after the first completes.

nitincchauhan:
'The second Callable in the queue will not execute until after the first completes.'
any requests that comes after queue size more than max pool size will be discarded by executor?
I think this is change makes it obvious rather than figuring out and relying on executor setting. probably eases out pressure on executor as well.

nitincchauhan:
also reducing complexity by keeping totalConnection counter in smaller method and only increments it on success ('together' with addting connection to bag). not speculative increment or decrements on failure

nitincchauhan:
changed order of increment, as it was in createPoolEntry,  passes tests :)

brettwooldridge:
The queue for the [addConnectionExecutor](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L109) has a maximum size of ``maximimPoolSize`` and a policy of ``ThreadPoolExecutor.DiscardPolicy()``.

This change *only works* because ``addConnectionExecutor`` is single-threaded.  ``totalConnections.get() + 1`` is inherently thread unsafe and wouldn't work if the executor were multi-threaded, but I am willing to accept this change because I have no plans to allow a multi-threaded executor for adding connections.

brettwooldridge:
Do you know how to squash commits?  If so, please squash everything from 329c438..540a8e4 into a single commit, then force push to your repo.

nitincchauhan:
wish I knew :) I do not know to 'un-commit'.
If you click on 'Files changed 2' above, it shows only latest changes... you can merge this pull as is then.

brettwooldridge:
Squashed and merged.

nitincchauhan:
@brettwooldridge Thanks
gentle reminder: any plan to refactor if (weakThreadLocals) from ConcurrentBag ?

PR #491, opened 2015-11-11
Committing the SQL Log - Initial Version

mmuruganandam:
@brettwooldridge I have implemented the initial version of capturing the SQL that are running beyond the threshold set.  I like you to review this initial draft.  It is not anywhere close to the final version.  Just wanted to make this feature available.  

If you can run the performance benchmark (sorry, I didn't setup the benchmark process yet in my machine) to see how much of an overhead that we are looking at and what would be the value addition out of this feature.

I feel that this feature would be a valuable addition to HikariCP.


brettwooldridge:
@mmuruganandam I appreciate the effort.  There are some technical issues, and then a basic philosophical issue...

For PreparedStatements your code relies on ``getParameterMetaData()``.  This method is not implemented by many drivers, including MySQL.  Additionally, it depends on the driver overriding  ``Object.toString()`` for ``ParameterMetaData``.  Even if all drivers supported getting the metadata, this wouldn't work -- it would need to walk the data structure and compose output.

The last issue is philosophical.  If HikariCP even starts to support SQL logging it will open the gates to a flood of enhancement requests.  Everyone wants something different from logging.  Of course, each request by itself seems reasonable, and only a small incremental change, but as we say it is death by a thousand cuts.

Take a look at [log4jdbc-log4j2](https://code.google.com/p/log4jdbc-log4j2/) as an example of the logging flexibility users would expect:

```
jdbc.sqlonly
  Logs only SQL. SQL executed within a prepared statement is automatically shown
  with it's bind arguments replaced with the data bound at that position, for greatly
  increased readability.

jdbc.sqltiming
  Logs the SQL, post-execution, including timing statistics on how long the SQL
  took to execute.

jdbc.audit
  Logs ALL JDBC calls except for ResultSets. This is a very voluminous output, and
  is not normally needed unless tracking down a specific JDBC problem.

jdbc.resultset
  Even more voluminous, because all calls to ResultSet objects are logged.

jdbc.resultsettable
  Log the jdbc results as a table. Level debug will fill in unread values in the result set.

jdbc.connection
  Logs connection open and close events as well as dumping all open connection
  numbers. This is very useful for hunting down connection leak problems.
```
And that is just the beginning.  [log4jdbc-log4j2](https://code.google.com/p/log4jdbc-log4j2/) allows users to enable/disable logging based on statement type (select, insert, update, create, delete).  It allows different timing thresholds for WARN and ERROR level logs.  It allows logging only SQLExceptions.  It allows logging generated keys.  And a lot more.

The question is, why would HikariCP implement logging when there are sophisticated JDBC driver wrappers already available?  What can HikariCP do better?  If the answer is simply to eliminate a dependency by providing an all-in-one solution, unfortunately that does not reach the threshold to justify implementation.

Have you ever bought an all-in-one tool?  They're never as good as a set of tools, each specifically designed to do one thing well.

tl15

PR #496, opened 2015-11-13
do not add cons if pool is not normal, avoid exceptions in shutdown

nitincchauhan:


nitincchauhan:
travis log now shows
just 2 (this build: https://travis-ci.org/brettwooldridge/HikariCP/builds/90901894)
vs 62 (previous build: https://travis-ci.org/brettwooldridge/HikariCP/builds/90464435)
occurrence of log message ...ConcurrentBag has been closed, ignoring add()

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -222,13 +222,15 @@ public final synchronized void shutdown() throws InterruptedException
...
+         connectionBag.close();
+
+         final ExecutorService assassinExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), "Hikari connection assassin (pool " + poolName + ")",
                                                                            config.getThreadFactory(), new ThreadPoolExecutor.CallerRunsPolicy());
          try {
             final long start = clockSource.currentTime();
             do {
-               softEvictConnections();
                abortActiveConnections(assassinExecutor);
+               softEvictConnections();
```
brettwooldridge: Why were these two switched?  We should definitely be trying soft eviction before abort.

nitincchauhan:
to achieve sequence as:
soft eviction happens much before loop once that would queue idle cons for closing.
then abort borrowed cons
then close any cons left open

brettwooldridge:
Previous the sequence was this:

1. connectionBag.close();
2. softEvictConnections();
3. addConnectionExecutor.shutdown();  // waits 5 seconds
4. loop: <br>
4a.   softEvictConnections(); <br>
4b.   abortActiveConnections(assassinExecutor);

New sequence is this:

1. softEvictConnections();
2. addConnectionExecutor.shutdown();  // waits 5 seconds<br>
   *note: new connections can still be added for up to 5 seconds*
3. connectionBag.close(); <br>
   *note: not much point closing now, this is basically a no-op, because no new connections can be added*
4. loop: <br>
4a.    abortActiveConnections(assassinExecutor); <br>
4b.    softEvictConnections(); 

*Both* solutions are sub-optimal, because the ``softEvictConnections()`` call inside the loop doesn't really give connections a chance to close cleanly before aborting them.  Better would be a double loop:
```java
long start = clockSource.currentTime();
do {
   softEvictConnections();
} while (getTotalConnections() > 0 && clockSource.elapsedMillis(start) < TimeUnit.SECONDS.toMillis(5));

start = clockSource.currentTime();
do {
   abortActiveConnections(assassinExecutor);
} while (getTotalConnections() > 0 && clockSource.elapsedMillis(start) < TimeUnit.SECONDS.toMillis(5));
```

nitincchauhan:
you are probably missing the point.
1. change is to avoid exceptions, they are delaying shutdown (os/container may not give much time to shutdown pool). now log shows 2 exceptions vs 62 before
2. as soft eviction is already called (before loop), in loop it is better to 'abort' borrowed connections as soft eviction ''then' happens for (rare-newly added/ left over) 'idle' connections only.
separate loop for soft eviction is bad as it would take 5 seconds if any one connection is 'in use'

TomiTakussaari:
Hi @brettwooldridge 

Do you have any plans regarding releasing new version of HikariCP with changes done in this PR ?

We have somewhat large ESB-style application that connects to many different databases using HikariCP. We ran it in Tomcat7, and use Tomcats hot-deploy feature to deploy new versions of it.

As we connect to lots of different databases that may or may not be up at any given time, we have initializationFailFast set as false and minimumIdle as 0, so that our application could start and do something useful even if some databases are down.

Now, when we deploy new version of application when database is down, our application is not cleanly undeployed from servlet container. 
"Hikari connection filler" thread stays alive and seemingly tries to add more connections to pool, even when pool itself and application using it are shutdown.

This then seems to result in eventual OOM, after multiple such deployments, atleast when running our application in Tomcat7.
This is not that big problem in production, as databases generally stay up there, but with some test & dev environments this is getting somewhat of a nuisance.

Looking at changes in this PR, this should fixed now that PoolEntryCreator only tries to do its thing when pool is running.

So, would be great if new version could be released sometime soon.

brettwooldridge:
@TomiTakussaari I have published 2.4.3 to Sonatype.  You can get it directly from there, or wait a few hours before it is replicated to other repositories.

TomiTakussaari:
@brettwooldridge thanks, works as expected!

PR #501, opened 2015-11-18
simplified validation, in sync with initializeDataSource() and DriverDataSource()

nitincchauhan:


src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -766,26 +766,18 @@ public void validate()
...
-         throw new IllegalArgumentException("jdbcUrl is required with driverClassName");
+      if (dataSource != null) {
+         LOGGER.debug("{} - using dataSource={}", poolName, dataSource);
       }
-      else if (driverClassName != null && dataSourceClassName != null) {
-         LOGGER.error("cannot use driverClassName and dataSourceClassName together");
-         throw new IllegalArgumentException("cannot use driverClassName and dataSourceClassName together");
-      }
-      else if (jdbcUrl != null && dataSourceClassName != null) {
-         LOGGER.warn("using dataSourceClassName and ignoring jdbcUrl");
```
brettwooldridge: Why would you remove these errors and warnings?  *Nobody* runs with debug logging on.  These are some of the most common misconfigurations and misunderstandings.  Removing these is going to generate *more* support requests.

I'd be happier if you simply threw exceptions instead.

nitincchauhan: those 'ignoring' message or exception is 'our' creation. specifying both would not break anything as pool proceeds with initiating data source in order of dataSource then dataSourceClassName then jdbcUrl
However,  ...to avoid tickets ..adding these checks and exceptions ..in few mins

nitincchauhan:
@brettwooldridge Great ! just in time to celebrate '16 ;) Very Happy and Prosperous New Year :+1:

PR #502, opened 2015-11-18
Dev1

nitincchauhan:


nitincchauhan:
@brettwooldridge 
added validation exceptions as before

nitincchauhan:
@brettwooldridge 
Also
1. reduced sleep time, now for default 30 mins connection time out it would sleep for 7sec
2. fixed my mistake in calculating lifetime, now avoids negative.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -573,7 +573,7 @@ public Boolean call() throws Exception
 
             // failed to get connection from db, sleep and retry
             quietlySleep(sleepBackoff);
-            sleepBackoff = Math.min(connectionTimeout / 2, (long) (sleepBackoff * 1.3));
+            sleepBackoff = Math.min(connectionTimeout / 5, (long) (sleepBackoff * 1.3));
```
brettwooldridge: Why 5?  With the minimum ``connectionTimeout`` of 1000ms this will generate a constant load of 5 connection attempts per second (200ms apart).  ``/ 2`` gives a much more reasonable retry value.

brettwooldridge: The idea for the backoff is, for the default 30 second timeout:
```
Attempt  |  Delay
---------+----------
1st      | 200ms
2nd      | 260ms
3rd      | 338ms
4th      | 439ms
5th      | 571ms
...      | ...
10th   | 2.2s
...      | 
> 17   | 15s
```

nitincchauhan: connection attempts per second is another angle.
Knowing connectiontimeout is 'usually' default or more, not 1second, which is more than 6 seconds of sleep for every failed attempt, HikariCP will be reasonably quicker to respond to network failure and fixes.

brettwooldridge: If I were writing a high-traffic website, like Wix, I would absolutely set ``connectionTimeout`` to 1000ms.  Front-end users experiencing >1s delays tend to walk away.  According to [KISSMetrics](https://blog.kissmetrics.com/loading-time/), *"40% of people abandon a website that takes more than 3 seconds to load."*  A 1000ms connection timeout allows the application layer to failover or switch pools before abondonment occurs.

That said, if database connections are encrypted (SSL) then 200ms is the typical minimum setup time for a connection, and the one core will be at nearly 100% for the duration of the setup.  A backoff that never goes above 200ms will choke one core completely.

nitincchauhan: agreed. lets see another angle.
for 1000s. till 5th failure 200*1.3 plays good role. by then, it already slept for (sum) 1009ms
by then the client already notified of 'timeout failure'
considering both setup (high / low value of connection timeout), probably reasonable value to sleep minimum 250 mills - max 5 seconds?

nitincchauhan:
@brettwooldridge thought over and as you said before, using connectionTimeout  as is (not dividing). makes sense. now it is minimum of 5 sec or connectionTimeout or sleepBackoff * 1.3

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -573,7 +572,7 @@ public Boolean call() throws Exception
 
             // failed to get connection from db, sleep and retry
             quietlySleep(sleepBackoff);
-            sleepBackoff = Math.min(connectionTimeout / 2, (long) (sleepBackoff * 1.3));
+            sleepBackoff = Math.min( TimeUnit.SECONDS.toMillis(5), Math.min(connectionTimeout, (long) (sleepBackoff * 1.3)));
```
brettwooldridge: Am I missing something?  This is going to limit the backoff time to a maximum of 5 seconds?

nitincchauhan: yes. as I said before: it is minimum of 5 sec or connectionTimeout or sleepBackoff * 1.3

it sleeps for sleepBackoff time and when sleepBackoff becomes more than connectionTimeout, it sleeps for 5 second

that makes it good for smaller connectionTimeout as you wanted for high traffic / 1sec timeout.
also better for larger connectionTimeout. makes HikariCP responsive within 5-10 seconds

nitincchauhan:
timing issues again ? after regression fix for initializationFailFast it
Failed tests: 
  StatementTest.testStatementClose:40 Totals connections not as expected
  TestConnections.testCreate:78 Totals connections not as expected expected same:<1> was not:<0>
  TestConnections.testMaximumPoolLimit:357 expected:<4> but was:<5>
  UnwrapTest.testUnwrapConnection:48 Idle connections not as expected expected same:<1> was not:<0>

nitincchauhan:
@brettwooldridge 
before, tests were passing because it was **waiting** in getConnection()
tests started to fail because the call is removed for 'right' reason.

nitincchauhan:
@brettwooldridge 
something is not right in failing this test:
TestConnections.testMaximumPoolLimit:357 expected:<4> but was:<5>

nitincchauhan:
in logs count of 'Added connection' and 'Closing connection' is 4.
but StubConnection.count is 5 !? no logs of creation and closing of 5th connection in logs :(

nitincchauhan:
probably other eyes may find fix for
Failed tests in this PR: 
TestConnections.testMaximumPoolLimit:370 expected:<4> but was:<5>

brettwooldridge:
I'll try to look at it tonight.

nitincchauhan:
Passed finally. It was not really a bug in test or pool but misunderstanding on my part.
coincidence, realized it just in time with 'lucky' comment :)
that extra connection was created and closed 'without any trace' in initializeConnections() now.
moving StubConnection.count later in test should work too.

nitincchauhan:
I think there is something more to this. now it fails with LESS in connections counter !!
Failed tests: 
  TestConnections.testMaximumPoolLimit:367 StubConnection count not as expected expected same:<4> was not:<3>

nitincchauhan:
everything is alright :) sorry for noise.
realized StubConnection.count can not be reset in-between ..for correct tally, reset is required just 'before' call to fillPool()

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -313,7 +313,6 @@ private Connection newConnection() throws Exception
 
          connection = (username == null) ? dataSource.getConnection() : dataSource.getConnection(username, password);
          setupConnection(connection);
-         lastConnectionFailure.set(null);
```
brettwooldridge: Why was this removed?

nitincchauhan: as name lastConnectionFailure suggests, to keep it till getAndSet() call,
nullifying it after connection creation is going to remove 'last' failure happened else where.
however, no harm in reporting 'very old' failure

brettwooldridge: Reporting an old exception, after the database has already recovered, can be very confusing for the user.

If the getConnection() call times out because all connections are in use, but reports a communications failure, we've mislead the user -- especially because we know that the database has recovered.

I am prepared to merge this request, but not with this change.

nitincchauhan: reverted.

nitincchauhan: @brettwooldridge : ...'I am prepared to merge this request, but not with this change.'
any correction required for merging this?

nitincchauhan:
@brettwooldridge 
merge at your earliest :) this may shed some light or resolve issue 497.

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -848,7 +867,7 @@ private void logConfiguration()
                value = dsProps;
             }
             value = (prop.contains("password") ? "<masked>" : value);
-            LOGGER.debug((prop + "................................................").substring(0, 32) + (value != null ? value : ""));
+            LOGGER.debug((prop + "................................................").substring(0, 32) + value);
```
brettwooldridge: Why was this changed?  Isn't this going to cause messages like?
```
connectionTestQuery...................................null
``
I don't want to see ``null`` as a value in the output.

File: src/main/java/com/zaxxer/hikari/pool/HikariPool.java [hunk 5]
```diff
@@ -507,13 +506,7 @@ private void initializeConnections()
...
       if (config.isInitializationFailFast()) {
          try {
-            Connection connection = getConnection();
-            if (config.getMinimumIdle() == 0) {
-               evictConnection(connection);
-            }
-            else {
-               connection.close();
-            }
+            newConnection().close();
```
brettwooldridge: This is less efficient.  This will create a connection, and then close it.  The existing code will only close the connection if ``minimumIdle=0``, otherwise it will be returned to the pool.

nitincchauhan: 2.4.1 and before it was returning immediately with correct exception if any.
now It is not returning until 'connectionTimeout'

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -55,9 +55,9 @@
 
    protected Connection delegate;
 
-   private final PoolEntry poolEntry;
-   private final ProxyLeakTask leakTask;
-   private final FastList<Statement> openStatements;
```
brettwooldridge: Why was ``final`` removed?

nitincchauhan: to nullify them and help GC proxyconnection asap

src/main/java/com/zaxxer/hikari/HikariConfig.java
nitincchauhan: to distinguish it from (empty) space/s ..as was the case in one issue

brettwooldridge: If that is the goal, this doesn't accomplish it in terms of debugging.  Properties with trailing spaces would still be hard to debug.
```java
value = (value instanceof String) ? (value != null ? '"' + value + '"' : value) : value;
```

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
brettwooldridge: Remove this change, it does not accelerate GC at all because these are backward references toward the GC root.

src/main/java/com/zaxxer/hikari/HikariConfig.java
nitincchauhan: yes you can enhance it like that. however string properties are 'trimmed' using new utility method.

brettwooldridge: I didn't notice that.  If that is the case, we would prevent passwords with leading or trailing spaces.  Those values should not be trimmed.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: done

src/main/java/com/zaxxer/hikari/HikariConfig.java
nitincchauhan: not allproperties, getNullIfEmpty() is applied only on catalog, connectionInitSql, connectionTestQuery, transactionIsolationName, dataSourceClassName, dataSourceJndiName, driverClassName and jdbcUrl

nitincchauhan: left username and passowrd for same reason

brettwooldridge: That's fine.  Please integrate the change above to quote strings in the debug output.  Users of the pool do not know what is occurring internally (trimming, no trimming, some trimming) but the debug output can make it clear.

nitincchauhan: done.

PR #530, opened 2015-12-27
Updated Popular DataSource list

amallia:
Added OrientDataSource to the list of popular datasources. I have tested OrientDB with HikariCP and it is perfectly working!


amallia:
Thanks!

PR #534, opened 2015-12-31
missing bits from last pr merge and some

nitincchauhan:


src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -315,7 +315,8 @@ Connection newConnection() throws Exception
       }
       catch (Exception e) {
          lastConnectionFailure.set(e);
-         quietlyCloseConnection(connection, "(Failed to create/set connection)");
+         String reason = e.getMessage();
+         quietlyCloseConnection(connection, reason != null ? reason : "(Failed to create/set connection)");
```
brettwooldridge: Intentionally not merged.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -248,8 +248,10 @@ public final void close() throws SQLException
             }
          }
          finally {
-            delegate = ClosedConnection.CLOSED_CONNECTION;
-            poolEntry.recycle(lastAccess);
+            if (delegate != ClosedConnection.CLOSED_CONNECTION) {
+               delegate = ClosedConnection.CLOSED_CONNECTION;
+               poolEntry.recycle(lastAccess);
+            }
```
brettwooldridge: Intentionally not merged.

src/main/java/com/zaxxer/hikari/util/PropertyElf.java
```diff
@@ -41,7 +41,7 @@
 {
    private static final Logger LOGGER = LoggerFactory.getLogger(PropertyElf.class);
 
-   private static final Pattern GETTER_PATTERN = Pattern.compile("(get|is)[A-Z].+");
+   private static final Matcher METHOD_MATCHER = Pattern.compile("(get|is)[A-Z].+").matcher("");
```
brettwooldridge: Matcher is not thread-safe and cannot be held as a static member.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: it is better to know 'actual reason', this text is more important as there can be many reason and is used for logging purpose to help debug/trace problem

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: this avoids 'possibility' of adding to the bag and worse 'giving user' a ClosedConnection.CLOSED_CONNECTION.

src/main/java/com/zaxxer/hikari/util/PropertyElf.java
nitincchauhan: good catch

PR #538, opened 2016-01-04
static log

nitincchauhan:


brettwooldridge:
The problem with static loggers is that users with multiple pools in the same VM (scoped class loader) cannot independently control logging.  We may have other instances of static loggers, but they should be corrected.

nitincchauhan:
@brettwooldridge there are all of 9 instances. now that you are at it, pls fix where required.

PR #556, opened 2016-01-25
Skip the `testMetricUsage` test on Windows

Shredder121:
There I am again, I haven't had the spurious wake-up issue since, which is good.

I know that the reference code is not exactly public, but it makes the skip reason readable enough.

Let me know what you think.


brettwooldridge:
@Shredder121 Why is it necessary to skip *testMetricUsage* on Windows?

Shredder121:
It's not _necessary_, but it's just _slightly_ too unstable/flaky for my taste.
The metric occasionally (every other build) reports 249ms as I described in the commit 8adf36bc3d109aebb84a8ba37b7edb9cb87ca222.

Although I think that a normal `assumeFalse`also works.
Anyway, let me know your opinion first.

PR #566, opened 2016-01-30
Add metric for recording count and rate of connection timeouts

wvuong:
Added `MetricsTracker.recordConnectionTimeout()` and tied that into `HikariPool.getConnection()` when we hit the connection timeout threshold.  Also added an implementation to `CodaHaleMetricsTracker` with a counter and a meter.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/MetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/MetricsTracker.java
@@ -35,6 +35,10 @@ public class MetricsTracker implements AutoCloseable
    {
    }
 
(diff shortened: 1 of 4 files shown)


brettwooldridge:
@wvuong Nice work.

brettwooldridge:
@wvuong Having looked at this again, isn't the *ConnectionTimeouts* ``Count`` metric redundant, because ``getCount()`` is available from the *ConnectionTimeoutRate* ``Meter`` instance?  By nature the ``Meter`` must keep track of *count* in order to calculate the *rate*.

wvuong:
you're totally right :)

PR #567, opened 2016-02-01
avoid 'more' exceptions after 'un-recoverable' exception thrown before for the connection

nitincchauhan:


nitincchauhan:
@brettwooldridge reminder: if you like to change some 'static' loggers, it is now :)

nitincchauhan:
@brettwooldridge last innocent looking' commit failed one interesting test:
Failed tests: 
  ConnectionRaceConditionTest.testRaceCondition:67 Task failed

nitincchauhan:
@brettwooldridge test passed but mystery remains !

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -400,22 +399,27 @@ final void releaseConnection(final PoolEntry poolEntry)
    /**
     * Permanently close the real (underlying) connection (eat any exception).
     *
-    * @param poolEntry the connection to actually close
+    * @param poolEntry poolEntry having the connection to close
+    * @param closureReason reason to close
+    * @param callClose false when the connection is dead because of un-recoverable exception and calling any method would result in exception, true otherwise
     */
-   final void closeConnection(final PoolEntry poolEntry, final String closureReason)
+   final void closeConnection(final PoolEntry poolEntry, final String closureReason, final boolean callClose)
```
brettwooldridge: Regardless of whether the exception is un-recoverable or not, it is essential that we *always* call the driver's ``Connection.close()`` statement.  If we do not do this, we risk leaking system handles (Socket handles at a minimum), and we prevent the driver from cleaning up internal state such as tracked connections.

Additionally, even if the server disconnects or is unreachable, there is no guarantee that the ``close()`` itself would fail, that is entirely driver dependent.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -167,7 +167,7 @@ public final Connection getConnection(final long hardTimeout) throws SQLExceptio
 
             final long now = clockSource.currentTime();
             if (poolEntry.isMarkedEvicted() || (clockSource.elapsedMillis(poolEntry.lastAccessed, now) > ALIVE_BYPASS_WINDOW_MS && !isConnectionAlive(poolEntry.connection))) {
-               closeConnection(poolEntry, "(connection is evicted or dead)"); // Throw away the dead connection and try again
+               closeConnection(poolEntry, "(connection is dead)"); // Throw away the dead connection (passed max age or failed alive test)
```
brettwooldridge: I prefer the more explicit message.  Put it back or change to:
```
closeConnection(poolEntry, "(connection is " + (poolEntry.isMarkedEvicted() ? "evicted)" : "dead)");
```

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -423,6 +427,11 @@ public void run() {
    //                           Private methods
    // ***********************************************************************
 
+   private void closeConnection(final PoolEntry poolEntry, final String closureReason)
+   {
+	   closeConnection(poolEntry, closureReason, true);
```
brettwooldridge: Kill the third parameter to ``closeConnection()``, we cannot skip calling ``close()``.  The "penalty" in the rare case where an exception *is* thrown by ``close()`` is merely an additional log (at debug level!).

nitincchauhan:
@brettwooldridge suggestions committed

src/main/java/com/zaxxer/hikari/pool/HikariPool.java

src/main/java/com/zaxxer/hikari/pool/HikariPool.java

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: reverted

PR #570, opened 2016-02-04
separate validationTimeout and connectionTimeout and validate.

nitincchauhan:
init networkTimeout only once
use validationTimeout for network, query and validation


nitincchauhan:
something is not right about this commit ...shows other changes I never made

PR #571, opened 2016-02-04
separate validationTimeout and connectionTimeout and validation, init

nitincchauhan:
networkTimeout only once.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -234,14 +234,18 @@ public class HikariConfig implements HikariConfigMXBean
    @Override
    public void setConnectionTimeout(long connectionTimeoutMs)
    {
-      if (connectionTimeoutMs == 0L) {
+      if (connectionTimeoutMs == 0) {
          this.connectionTimeout = Integer.MAX_VALUE;
       }
-      else if (connectionTimeoutMs < 250L) {
+      else if (connectionTimeoutMs < 250) {
          throw new IllegalArgumentException("connectionTimeout cannot be less than 250ms");
       }
       else {
          this.connectionTimeout = connectionTimeoutMs;
+      }
+
+      if (validationTimeout > connectionTimeoutMs) {
+         this.validationTimeout = connectionTimeoutMs;
       }
    }
 
@@ -256,11 +260,15 @@ public class HikariConfig implements HikariConfigMXBean
    @Override
    public void setValidationTimeout(long validationTimeoutMs)
    {
-      if (validationTimeoutMs < 250L) {
+      if (validationTimeoutMs < 250) {
          throw new IllegalArgumentException("validationTimeout cannot be less than 250ms");
       }
       else {
          this.validationTimeout = validationTimeoutMs;
+      }
+
+      if (validationTimeout > connectionTimeout) {
+         this.validationTimeout = connectionTimeout;
       }
    }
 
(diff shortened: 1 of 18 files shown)


nitincchauhan:
strange: build failed with
Results :
Tests run: 97, Failures: 0, Errors: 0, Skipped: 0

nitincchauhan:
failing bcoz of this: https://travis-ci.org/brettwooldridge/HikariCP/builds/106946612#L1305

nitincchauhan:
more surprise: last commit fails this too ! earlier commits did NOT:
https://travis-ci.org/brettwooldridge/HikariCP/builds/106954043#L694

nitincchauhan:
for that same testConnectionIdleFill() test, log looks perfect in the first build:
https://travis-ci.org/brettwooldridge/HikariCP/builds/106946612#L672

brettwooldridge:
The code that allows a 250ms validation time should not be removed.  If setNetworkTimeout() is supported, it is perfectly possible to support sub-second and fractional-second timeouts, even while calling isValid() or setQueryTimeout() with full second values.

nitincchauhan:
I am lost as following comment is not shown inline ...where you want to use validationTimeout?
https://github.com/brettwooldridge/HikariCP/pull/571#commitcomment-15887554

brettwooldridge:
Look at the comments in [this context](https://github.com/brettwooldridge/HikariCP/pull/571/files).

brettwooldridge:
Basically, scrap everything and keep the changes to ``setupConnection()`` (with recommended changes).  The "unable" to "failed" logging change is fine as well.

nitincchauhan:
done.

nitincchauhan:
@brettwooldridge have not included additional log messages (validationTimeout is less than second or not multiple of) because it may add support questions and JDK doc for the method already says: ' If the JDBC driver does not support milliseconds, the JDBC driver will round the value up to the nearest second. '

Severe: Most numeric properties that are allowed to change at run time needs validation. may make pool behavior un-predictable to diagnose and worse crash / hang.

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -42,27 +41,30 @@
...
-   private volatile long validationTimeout;
-   private volatile long idleTimeout;
-   private volatile long leakDetectionThreshold;
-   private volatile long maxLifetime;
+   private volatile int connectionTimeout;
+   private volatile int validationTimeout;
+   private volatile int idleTimeout;
+   private volatile int leakDetectionThreshold;
+   private volatile int maxLifetime;
    private volatile int maxPoolSize;
```
brettwooldridge: While it might seem convenient, we cannot change these, it will break the ABI (application binary compatibility).  Users who have code like:
```java
config.setConnectionTimeout(TimeUnit.SECONDS.toMillis(20));
```
Would now get a compilation failure because ``long`` cannot be downcast to an ``int``.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -125,14 +126,14 @@ boolean isConnectionAlive(final Connection connection)
    {
       try {
          if (isUseJdbc4Validation) {
-            return connection.isValid((int) MILLISECONDS.toSeconds(Math.max(1000L, validationTimeout)));
+            return connection.isValid((int) MILLISECONDS.toSeconds(Math.max(1000, validationTimeout)));
```
brettwooldridge: If ``validationTimeout`` is an ``int``, we incur an additional unseen cast:
```java
MILLISECONDS.toSeconds(Math.max(1000, validationTimeout))
                       ^
                      here
```
Math.max() will return an ``int``.

src/main/java/com/zaxxer/hikari/HikariConfig.java
nitincchauhan: have not thought of compatibility. if you can please merge except 'long to int' change and close.

brettwooldridge:
I cherry picked what I liked and cut the parts I didn't.

Issue #576, opened 2016-02-08
relax lastAccess setting


nitincchauhan:
With narrower escape validation window and many refactor / updates later, I think it is time to revisit decision to have overhead (in case of read-only or auto-commit connections) of initializing lastAccess in markCommitStateDirty(). earlier it was always set in close() only.

brettwooldridge:
If we *always* set ``lastAccess`` to ``CloseSource.currentTime()`` in ``close()``, as we used to, then the connection can be immediately borrowed without a validation check (bc of the alive-bypass window).  This lead to a rare error where a connection that was borrowed, but never used, was returned to the pool in a disconnected state.  Because we set ``lastAccess`` to ``currentTime()`` in ``close()``, the connection was borrowed without testing, resulting in an error at the application level.

It is important that ``lastAccess`` reflect the last known time that we *know* that the connection was valid.

PR #581, opened 2016-02-13
small but important fixes

nitincchauhan:
1. close() must not call getNextException() in case of already closed connection ...separated check for (delegate != ClosedConnection.CLOSED_CONNECTION)
2. set lastAccess for readOnly connection too.
3. Init leak task (fills stack) before starting housekeeper
4. removed warning for multiple of seconds. as per doc driver should round it to nearest second anyway
5. removed re/set validationTimeout. consistent with other properties not reset in set methods.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -163,7 +163,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       try {
          long timeout = hardTimeout;
          do {
-            final PoolEntry poolEntry = connectionBag.borrow(MILLISECONDS.toNanos(timeout), clockSource.toNanos(now));
+            final PoolEntry poolEntry = connectionBag.borrow(timeout, MILLISECONDS);
             if (poolEntry == null) {
                break; // We timed out... break and throw exception
             }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -447,7 +447,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
             }, lifetime, MILLISECONDS));
          }
 
-         LOGGER.debug("{} - Added connection {}", poolName, poolEntry.connection);
+         LOGGER.debug("{} - Added connection {}", poolName, poolEntry.getConnectionName());
          return poolEntry;
       }
       catch (Exception e) {
@@ -485,14 +485,14 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
    private void abortActiveConnections(final ExecutorService assassinExecutor)
    {
       for (PoolEntry poolEntry : connectionBag.values(STATE_IN_USE)) {
+         Connection connection = poolEntry.close();
          try {
-            poolEntry.connection.abort(assassinExecutor);
+            connection.abort(assassinExecutor);
          }
          catch (Throwable e) {
-            quietlyCloseConnection(poolEntry.connection, "(connection aborted during shutdown)");
+            quietlyCloseConnection(connection, "(connection aborted during shutdown)");
          }
          finally {
-            poolEntry.close();
             if (connectionBag.remove(poolEntry)) {
                totalConnections.decrementAndGet();
             }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -173,7 +173,6 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
                timeout = hardTimeout - clockSource.elapsedMillis(startTime);
             }
             else {
-               poolEntry.lastBorrowed = now;
                metricsTracker.recordBorrowStats(poolEntry, startTime, now);
                return poolEntry.createProxyConnection(leakTask.schedule(poolEntry, houseKeepingExecutorService), now);
             }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -158,19 +158,20 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
    {
       suspendResumeLock.acquire();
       final long startTime = clockSource.currentTime();
+      long now = startTime;
 
       try {
          long timeout = hardTimeout;
          do {
-            final PoolEntry poolEntry = connectionBag.borrow(timeout, MILLISECONDS);
+            final PoolEntry poolEntry = connectionBag.borrow(clockSource.toNanos(timeout), clockSource.toNanos(now));
             if (poolEntry == null) {
                break; // We timed out... break and throw exception
             }
 
-            final long now = clockSource.currentTime();
             if (poolEntry.isMarkedEvicted() || (clockSource.elapsedMillis(poolEntry.lastAccessed, now) > ALIVE_BYPASS_WINDOW_MS && !isConnectionAlive(poolEntry.connection))) {
                closeConnection(poolEntry, "(connection is evicted or dead)"); // Throw away the dead connection (passed max age or failed alive test)
-               timeout = hardTimeout - clockSource.elapsedMillis(startTime);
+               now = clockSource.currentTime();
+               timeout = hardTimeout - clockSource.elapsedMillis(startTime, now);
             }
             else {
                metricsTracker.recordBorrowStats(poolEntry, startTime, now);
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -131,7 +131,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
          this.houseKeepingExecutorService = config.getScheduledExecutorService();
       }
 
-      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold(), houseKeepingExecutorService);
+      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold());
 
       this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
    }
@@ -174,7 +174,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
             }
             else {
                metricsTracker.recordBorrowStats(poolEntry, startTime);
-               return poolEntry.createProxyConnection(leakTask.start(poolEntry), now);
+               return poolEntry.createProxyConnection(leakTask.schedule(poolEntry, houseKeepingExecutorService), now);
             }
          } while (timeout > 0L);
       }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -131,9 +131,9 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
          this.houseKeepingExecutorService = config.getScheduledExecutorService();
       }
 
-      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold());
-
       this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
+
+      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold());
    }
 
    /**
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -163,7 +163,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       try {
          long timeout = hardTimeout;
          do {
-            final PoolEntry poolEntry = connectionBag.borrow(clockSource.toNanos(timeout), clockSource.toNanos(now));
+            final PoolEntry poolEntry = connectionBag.borrow(MILLISECONDS.toNanos(timeout), clockSource.toNanos(now));
             if (poolEntry == null) {
                break; // We timed out... break and throw exception
             }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -131,9 +131,9 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
          this.houseKeepingExecutorService = config.getScheduledExecutorService();
       }
 
-      this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
-
       this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold(), houseKeepingExecutorService);
+
+      this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
    }
 
    /**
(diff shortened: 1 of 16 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -131,9 +131,9 @@ public HikariPool(final HikariConfig config)
          this.houseKeepingExecutorService = config.getScheduledExecutorService();
       }
 
-      this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
+      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold());
 
-      this.leakTask = new ProxyLeakTask(config.getLeakDetectionThreshold(), houseKeepingExecutorService);
+      this.houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 0L, HOUSEKEEPING_PERIOD_MS, MILLISECONDS);
```
brettwooldridge: What is the reason for switching these?  Functionally they are the same, but it reads more cleanly if all of the ``this.houseKeepExecutorService`` initialization is grouped together (next to lines 124-132).

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -173,8 +173,9 @@ public final Connection getConnection(final long hardTimeout) throws SQLExceptio
                timeout = hardTimeout - clockSource.elapsedMillis(startTime);
             }
             else {
-               metricsTracker.recordBorrowStats(poolEntry, startTime);
-               return poolEntry.createProxyConnection(leakTask.start(poolEntry), now);
+               poolEntry.lastBorrowed = now;
+               metricsTracker.recordBorrowStats(poolEntry, startTime, now);
+               return poolEntry.createProxyConnection(leakTask.schedule(poolEntry, houseKeepingExecutorService), now);
```
brettwooldridge: ``lastBorrowed`` is only used for metrics.  By allowing the ``metricsTracker`` to set (or not set) ``lastBorrowed`` we completely avoid the assignment when metrics are disabled.  The [NopMetricsTrackerDelegate](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L642) does nothing, whereas the [MetricsTrackerDelegate](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L614) does set ``lastBorrowed``.  I am fine with lines 177-178, but 176 should be deleted and put back in the ``MetricsTrackerDelegate``.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -436,13 +436,10 @@ private int getAndSetNetworkTimeout(final Connection connection, final long time
...
                isNetworkTimeoutSupported = FALSE;
 
-               LOGGER.warn("{} - Unable to get/set network timeout for connection. ({})", poolName, e.getMessage());
+               LOGGER.warn("{} - Failed to get/set network timeout for connection. ({})", poolName, e.getMessage());
                if (validationTimeout < SECONDS.toMillis(1)) {
                   LOGGER.warn("{} - A validationTimeout of less than 1 second cannot be honored on drivers without setNetworkTimeout() support.", poolName);
                }
-               else if (validationTimeout % SECONDS.toMillis(1) != 0) {
-                  LOGGER.warn("{} - A validationTimeout with fractional second granularity cannot be honored on drivers without setNetworkTimeout() support.", poolName);
-               }
```
brettwooldridge: Keep lines 443-445.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -145,13 +145,14 @@ final SQLException checkException(final SQLException sqle)
...
       if (sqlState != null) {
-         final boolean isForceClose = sqlState.startsWith("08") || SQL_ERRORS.contains(sqlState);
-         if (isForceClose && delegate != ClosedConnection.CLOSED_CONNECTION) {
-            LOGGER.warn("{} - Connection {} marked as broken because of SQLSTATE({}), ErrorCode({})",
-                        poolEntry.getPoolName(), delegate, sqlState, sqle.getErrorCode(), sqle);
-            leakTask.cancel();
-            poolEntry.evict("(connection is broken)");
-            delegate = ClosedConnection.CLOSED_CONNECTION;
+         if (sqlState.startsWith("08") || SQL_ERRORS.contains(sqlState)) { // broken connection
+            if (delegate != ClosedConnection.CLOSED_CONNECTION) {
```
brettwooldridge: You might as well merge the conditional in 149 into 147.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -378,6 +379,7 @@ public void setAutoCommit(boolean autoCommit) throws SQLException
    {
       delegate.setAutoCommit(autoCommit);
       isAutoCommit = autoCommit;
+      isCommitStateDirty = false;
```
brettwooldridge: This is not a valid change.  The JavaDoc says:

> If this method is called during a transaction and the auto-commit mode is changed, the transaction is committed
> If setAutoCommit is called and the auto-commit mode is not changed, the call is a no-op.

If auto-commit started as ``false`` and the user performed some SQL, then ``isCommitStateDirty`` is ``true``.  If the user calls ``setAutoCommit(false)`` during the transaction, it is a no-op (auto-commit is still *false*).  Which means that ``isCommitStateDirty`` should *still* be ``true`` also.  This change would set it to ``false`` unconditionally.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -387,6 +389,7 @@ public void setReadOnly(boolean readOnly) throws SQLException
    {
       delegate.setReadOnly(readOnly);
       isReadOnly = readOnly;
+      isCommitStateDirty = false;
```
brettwooldridge: This change is valid.  The JavaDoc states that this method cannot be called during a transaction, therefore if the ``delegate.setReadOnly()`` call does not throw an exception, we can implicitly assume that the transaction was committed.  Therefore, it is same to imply that ``isCommitStateDirty`` is false.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -170,7 +171,7 @@ final void untrackStatement(final Statement statement)
 
    final void markCommitStateDirty()
    {
-      if (isAutoCommit) {
+      if (isAutoCommit || isReadOnly) {
```
brettwooldridge: This is more expensive.  ``markCommitStateDirty()`` may be called dozens or hundreds of times in a single connection use, so ``isReadOnly`` is evaluated many times.  However, ``close()`` is only called once per connection use, and therefore checking the ``isReadOnly`` flag there (where it is currently), is more efficient.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: lastBorrowed is used in toString() as well which gives misleading value for NopMetricsTrackerDelegate

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: why repeat java doc: 'If the JDBC driver does not support milliseconds, the JDBC driver will round the value up to the nearest second.'

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: at present, if auto commit false and connection is and stays read only and not returned within bypass window, it is validating before leasing out because lastAccss is not initialized in markCommitStateDirty

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: committed.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: committed.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: committed.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: wrt: expensive, I had created issue https://github.com/brettwooldridge/HikariCP/issues/576 for same reason.
IMO with narrower validation window and many stability improvements later, it is 'ok' to live with that 'rare' problem as you mentioned there till better solution.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
brettwooldridge: A few reasons.

1) Users do not know what APIs that HikariCP uses internally for validation.  We use ``isValid()`` or ``setQueryTimeout()``, but we should not expect that the user should know.

2) Because we also use ``setNetworkTimeout()`` when it is available, we can support a ``validationTimeout`` with fractional seconds, for example 1250ms.  If the user sets a ``validationTimeout`` of 1250ms, *and* the driver supports ``setNetworkTimeout()`` we can actually honor the 1250ms timeout even though we call ``isValid()`` with 2 seconds as a timeout.  If the driver *does not* support ``setNetworkTimeout()`` the we cannot honor the user's requested 1250ms timeout, and therefore the warning is useful.

If we log this warning, the user is less likely to open an issue that HikariCP is not honoring their ``validationTimeout``.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
brettwooldridge: If you think so, we can suppress ``lastBorrowed`` from the toString() when metrics are disabled.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: I think logging 'less than second' message is enough. I kept that already.
additional message... fractional second granularity ...is not required and is different from docs which may really confuse users.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
nitincchauhan: it would be cumbersome to add check in PoolEntry#toString() ...need reference to config etc
moreover, I do not see any side effect of initializing lastBorrowed in both cases this way.

nitincchauhan:
I looked for and found that issue 289

1. issue was really connection leak or mysql cache leak
   issues/289#issuecomment-87635381
2. As suspected, lastAccess was set in finally ! which was wrong in case of broken connection
   HikariCP/commit/cd457367f7668feaa5a9e9d201c7d8b677923265

following changes reduces (to almost none) chances of pool leasing out bad connection:
1. evicting connection from bag immediately on finding it broken
2. initialising lastAccess only after successful call to clearWarning
   that implies connection is alive till put back for recycle()
3. escape window is 500 now, not 1000 like before

With latest changes, I expect better stability and figures in chart too :)

nitincchauhan:
Besides less calls, it is now calculating time difference 'correctly' having start of time at the begining of operation

nitincchauhan:
Besides one less call to system time, it is giving 'accurate' time difference now having time set at the very beginning and very end of the operation

brettwooldridge:
Sometimes I think you are making changes just because you like making changes.  I am going to cherry-pick some of these, but quite a few of them I do not like, such as the 'accurate' time difference change.  It may be more accurate, but it is gross to look at.

ConcurrentBag is a generic class and having a method with signature that takes a "now" parameter is just gross.  Similar to the classes in java.util.concurrent that support timeout, all that should be passed is a duration and TimeUnit.

brettwooldridge:
Nitin, slow down.  Some of these changes are dangerous.  Particularly the one involving ``lastAccessed``.  It is not valid to always set ``lastAccessed`` to currentTime() in close().  That is what lead to broken connections being returned from the pool in at least one case.  ``lastAccessed`` should only be set when we are certain that there was actual wire-traffic between the driver and the database.  If the user borrows a connection, sleeps for 5 minutes, and calls close() without ever using the connection then the ``lastAccessed`` time should remain unchanged (ie. at least 5 minutes in the past).

I am more interested in stability over performance at this point in HikariCP's lifecycle.  Ideally the only changes occurring now should be, 1) to fix a specific user reported issue, or 2) the additional of new functionality.

nitincchauhan:
ok pasting those comments here again I wrote for lastAccessed change in this PR. Please take time and read.

6 days before:
-----------------
setting lastAccess after successful call to delegate.clearWarnings(); is better and more importantly safer too because its doc says:
...throws SQLException if a database access error occurs or this method is called on a closed connection

I suspect lastAccess was set somewhere else in close earlier in case of reported 'rare' escape
--------------

a day later:
--------------
I looked for and found that issue 289

issue was really connection leak or mysql cache leak issues/289#issuecomment-87635381
As suspected, lastAccess was set in finally ! which was wrong in case of broken connection HikariCP/commit/cd457367f7668feaa5a9e9d201c7d8b677923265
following changes reduces (to almost none) chances of pool leasing out bad connection:
1. evicting connection from bag immediately on finding it broken
2. initialising lastAccess only after successful call to clearWarning
that implies connection is alive till put back for recycle()
if not because of SQLException, connection will go through validation test
because lastAccess is not set.
unless connection is returned back and leased out within 500ms
3. escape window is 500 now, not 1000 like before

With latest changes, I expect better stability and figures in chart too :)

nitincchauhan:
I believe commits in this PR are BUG fixes or better fixes then applied before. As always made with intention to give more stability and consistency.

I spotted thing as incorrect or found better alternative only then I changed things with my limited knowledge, not because I like it :)

nitincchauhan:
@brettwooldridge I committed changes : ...retaining ConcurrentBag and accuracy of 'calculating time diff' as well
I compared with your latest dev repo and all I can say is: Please merge all of it for good. it is apparent you missed reading comment line of commits made for most,

nitincchauhan:
btw it has not run build for last two commits !

nitincchauhan:
at present, it is never updating lastAccessed in case: user does nothing but borrows and closes connection.
this PR along with last commit, takes care of such edge case as well

brettwooldridge:
2.4.4-RC2:
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode
 Cnt      Score     Error   Units
ConnectionBench.cycleCnnection  jdbc:stub             32  hikari  thrpt
16  17363.070 ± 245.380  ops/ms


With Your changes:
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode
 Cnt      Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub             32  hikari  thrpt
16  15905.404 ±  579.645  ops/ms

nitincchauhan:
I hope last commit betters benchmark as well.

tomcgn:
Hi Brett and Nitin, 
I use hikariCP in a jhipster app and experienced occasional downtimes of the entire webapp. The hikaricp default debug messages did not point to conrete reasons. I changed the timeout values as recommended in the documentation and various stackoverflow questions. Still, the app breakdowns occured several days after each restart. After trying the recent releases up to 2.4.5 without a change, I switched to Nitin's branch.

In http://stackoverflow.com/questions/32968530/hikaricp-connection-is-not-available, a user describes a situation that is similar; according to him, the cause are methods in REST Controllers, that use Spring JPA repositories. I added the @Transactional
annotations to the top level methods, but still, the app breaks down every now and then. 

Not sure, whether taking care of the stability in this case is on the hibernate or hikariCP side. But both, more explicit logging and hikariCP recovering from that situation without crashing the application is desirable. 
Let's see, if Nitin's version runs more stable. 

Best, 
Tom

```
2016-02-16 12:42:35 ERROR [http-nio-8080-exec-77] (DirectJDKLog.java:182):
Servlet.service() for servlet [dispatcherServlet] in context with path [] threw excepti [Request processing failed; nested exception is org.springframework.dao.DataAccessResourceFailureException: Could not open connection; nested exception is org.hibernate.exception.JDBCConnectionException: Could not open connection] with root cause
java.sql.SQLTransientConnectionException: HikariPool-0 - Connection is not available, request timed out after 30183ms.
at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:194) ~[HikariCP-2.4.4-RC1.jar!/:na]
at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:145) ~[HikariCP-2.4.4-RC1.jar!/:na]... ```

nitincchauhan:
@tomcgn 
This branch is awaiting review.
Please post this with configuration and full stack trace in [issues](https://github.com/brettwooldridge/HikariCP/issues). preferably with debug level logs.

nitincchauhan:
some other time :)

Issue #615, opened 2016-04-19
Make metrics prefix configurable


jnehlmeier:
Currently the pool name is used as metrics prefix with an added "pool" token.

We use InfluxDB to store & query metrics data and InfluxDB has the concept of tagging metric values so you can quickly query them using a SQL-like WHERE clause. Because metrics can only have a name these tags are computed from the metric name and thus our metrics have a pretty strict form:  `<measurement>.<tag-key>.<tag-value>......<field>`. In addition to tags from metric names there might be global tags like the webapp's context path to distinguish metrics per webapp.

Currently we need to workaround that HikariCP appends a "pool" token to the pool name when constructing metrics. Also we would like to add some information to the pool name that should not result in a tag for metrics.

We could make our metric name parsing algorithm more complex but I think it would be better if you would decouple the the metric name from the pool name and let developers configure both at their will (the default can be as it is now). So ideally HikariCP should create metrics using something like `MetricRegistry.name(hikariConfig.getPoolMetricsPrefix(), "Wait")`

brettwooldridge:
@jnehlmeier What is an example of an InfluxDB metric name under your scheme?  When considering DropWizard metrics there is a naming convention that should be followed:
```
Every metric in a registry has a unique name, which is just a dotted-name string like
"things.count" or "com.example.Thing.latency"
```
Presumably you have something pulling metrics from DropWizard and inserting them into InfluxDB.  Isn't that the better location to perform your application specific translation?

jnehlmeier:
@brettwooldridge Of course we have a working solution but its more complex then it needs to just because we can not freely configure the metric prefix used by HikariCP. In addition I think you have to open up metric naming anyways because Metrics 4.0 will not only support String names for metrics but also an instance of class MetricName which supports tags (to avoid overly long metric names as in Metrics 3.x just to encode some data in them). But MetricName is immutable so if people want to add tags to the metrics of HikariCP you would need to add some API for it (see: https://github.com/brettwooldridge/HikariCP/pull/624#issuecomment-215986597) 

Anyways in our reporting to InfluxDB we generally distinguish two kinds of metrics: single valued (counter, gauge) and multi valued (timer, histogram, meter). We do this because in InfluxDB we want to combined single valued metrics if they belong to the same measurement. Some examples:

  - Timer: `handler.host.1.vm.1.customer.1234.name.ComputeHandler.execution-timings`
[...]

PR #624, opened 2016-04-30
Configurable Metrics prefix

EdgeCaseBerg:
Hi! I saw [issue 615](https://github.com/brettwooldridge/HikariCP/issues/615) and figure'd it would be something pretty easy to do in a morning. 
All the tests still pass, and I've written the update so that it wouldn't break anyone's metrics 
that are relying on the hardcoded "pool" being added to the pool name as the base of the
metrics name.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -54,6 +54,7 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long IDLE_TIMEOUT = MINUTES.toMillis(10);
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
    private static final int DEFAULT_POOL_SIZE = 10;
+   private static final String METRICS_PREFIX = "pool";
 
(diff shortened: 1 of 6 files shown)


jnehlmeier:
Seems to look good. 

With Metrics 4.x in the future HikariConfig would also need an API to provide metric tags because in Metrics 4.x (currently their master branch) their metrics name will be represented by an instance of `MetricName` (https://github.com/dropwizard/metrics/blob/master/metrics-core/src/main/java/io/dropwizard/metrics/MetricName.java) which is immutable. This class supports attaching custom tags to a metric so that metric names do not have to be relatively long just because tags are encoded directly in the metric name like I described in my issue.

So with Metrics 4.x you could do:
```
Map<String, String> tags = ....;
tags.put("customer", "example.com");
tags.put("host", "cluster-host-01");
tags.put("jvm-instance", "1");

// with "metricPrefix" just being, e.g. "connection-pool" or "hikaricp".
registry.register(new MetricName(metricPrefix + ".activeConnections", tags), someGauge);
```

Issue #650, opened 2016-06-16
Not recovering from Amazon Redshift disconnects


xstevens:
We are seeing an issue where Hikari was not recovering from a Redshift disconnect. It produces a stack trace like this:

```
! Causing: java.sql.SQLException: [Amazon](500150) Error setting/closing connection: Not Connected.
! at com.amazon.redshift.client.PGClient.isConnected(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.redshift.client.PGClient.prepareStatement(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.redshift.dataengine.PGQueryExecutor.<init>(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.redshift.dataengine.PGDataEngine.prepare(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.jdbc.common.SPreparedStatement.<init>(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.jdbc.jdbc41.S41PreparedStatement.<init>(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.redshift.core.jdbc41.PGJDBC41PreparedStatement.<init>(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.redshift.core.jdbc41.PGJDBC41ObjectFactory.createPreparedStatement(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.amazon.jdbc.common.SConnection.prepareStatement(Unknown Source) ~[aws-redshift-jdbc41-1.1.10.1010.jar:RedshiftJDBC_1.1.10.1010]
! at com.zaxxer.hikari.pool.ProxyConnection.prepareStatement(ProxyConnection.java:310) ~[HikariCP-2.4.7-SNAPSHOT.jar:na]
[...]

PR #651, opened 2016-06-16
Fixing Redshift specific disconnect error

xstevens:
Fixes: https://github.com/brettwooldridge/HikariCP/issues/650


brettwooldridge:
If I can ask a favor, before I merge this pull can you check Redshift for additional disconnection codes that should be registered as well?

xstevens:
@brettwooldridge Sure no problem. I'll take a look. I did a bit of searching and HY000 seems like a pretty generic code for some databases, so I'm a little worried that this may have other side effects. I'm not super familiar with SQLSTATE codes.

johnou:
At least mysql docs state that error indicates no DNS resolution for the given endpoint. 
Error 2005 (HY000): Unknown MySQL Server Host
http://dev.mysql.com/doc/refman/5.7/en/error-messages-client.html#error_cr_unknown_host

johnou:
@xstevens guess we can close this in favour of https://github.com/brettwooldridge/HikariCP/commit/21eedd550258f6a33bc0f4b22579366afe7f4e62 ?

xstevens:
@johnou Yes, thanks!

Issue #547, opened 2016-01-15
QUESTION: Why is the minimum value for connectionTimeout limited to 1000?


bysse:
Hi, we're evaluating this connection pool and i have a question as to why the minimum value of connectionTimeout is limited to 1000 ms?
> This property controls the maximum number of milliseconds that a client (that's you) will wait for a  connection from the pool. If this time is exceeded without a connection becoming available, a SQLException will be thrown. 1000ms is the minimum value. Default: 30000 (30 seconds)

If we have an issue with a single shard we will fill up all available threads on the server waiting for a connection rather than just drop the request straight away. Any thoughts on this?

bysse:
After finding the forums I came across this answer:

> The issue with setting the connectionTimeout too low (for example 10's of milliseconds) is that the connectionTimeout doubles as both the time that the user is willing to wait for a connection from the pool and the time that HikariCP itself is willing to wait when creating new connections to add to the pool.  A timeout in the 10's of milliseconds might be insufficient in many environment to establish and authenticate a new connection.

- https://groups.google.com/forum/#!searchin/hikari-cp/"connectionTimeout"/hikari-cp/ohaKBZma3R0/RPxqIkJ3DRQJ

Which unfortunately makes HikariCP unusable for our  needs.

brettwooldridge:
@bysse What minimum would be sufficient for your use case?

brettwooldridge:
@bysse We can certainly consider a lower floor, but 125ms would be the absolute minimum.

Both Windows and Linux have a default scheduler quantum of 20ms.  If 16 threads are running on a 4-core CPU, a single thread may have to wait up to 80ms just to be allowed to run.  If the pool has a vacancy due to, for example, the retirement of a connection at maxLifetime, this leaves precious little time to establish a connection to fill the slot without returning a *spurious* failure to the client.

If careful consideration is not taken to ensure the CPU and scheduler are not oversaturated, running at a 125ms timeout, puts your application tier at risk of spurious failures even if the pool has available connections.

bysse:
@brettwooldridge Thanks for the response. I would say that the problem as I see it is that the timeout for connecting, and the timeout waiting for a free connection from a full pool is the same. However, since we already have a thin layer that is limiting the number of simultaneous connections to the database we're not affected by the later. I think my initial thoughts was a bit skewed by our solution :)

We do not need to go as low as 125 ms but a smaller connectionTimeout of say 250 - 500 ms would be better.

[...]

PR #654, opened 2016-06-21
Tidyup

nitincchauhan:
also fix:
https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#close()
states
"Calling the method close on a Connection object that is already closed is a no-op."
but HikariCP throws exception

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -243,8 +243,8 @@ public class HikariConfig implements HikariConfigMXBean
    @Override
    public void setValidationTimeout(long validationTimeoutMs)
    {
-      if (validationTimeoutMs < 250) {
-         throw new IllegalArgumentException("validationTimeout cannot be less than 250ms");
+      if (validationTimeoutMs < 1000) {
+         throw new IllegalArgumentException("validationTimeout cannot be less than 1000ms");
       }
 
       this.validationTimeout = validationTimeoutMs;
@@ -820,8 +820,8 @@ public class HikariConfig implements HikariConfigMXBean
          connectionTimeout = CONNECTION_TIMEOUT;
       }
 
-      if (validationTimeout < 250) {
-         LOGGER.warn("{} - validationTimeout is less than 250ms, setting to {}ms.", poolName, VALIDATION_TIMEOUT);
+      if (validationTimeout < 1000) {
+         LOGGER.warn("{} - validationTimeout is less than 1000ms, setting to {}ms.", poolName, VALIDATION_TIMEOUT);
          validationTimeout = VALIDATION_TIMEOUT;
       }
 
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -104,6 +104,8 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       this.totalConnections = new AtomicInteger();
       this.suspendResumeLock = config.isAllowPoolSuspension() ? new SuspendResumeLock() : SuspendResumeLock.FAUX_LOCK;
 
+      checkFailFast();
+
       if (config.getMetricsTrackerFactory() != null) {
          setMetricsTrackerFactory(config.getMetricsTrackerFactory());
       }
@@ -114,8 +116,6 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
       setHealthCheckRegistry(config.getHealthCheckRegistry());
 
       registerMBeans(this);
-
-      checkFailFast();
 
       ThreadFactory threadFactory = config.getThreadFactory();
       this.addConnectionExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), poolName + " connection adder", threadFactory, new ThreadPoolExecutor.DiscardPolicy());
@@ -216,11 +216,10 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
 
          softEvictConnections();
 
-         if (addConnectionExecutor != null) {
-            addConnectionExecutor.shutdown();
-            addConnectionExecutor.awaitTermination(5L, SECONDS);
-         }
-         if (config.getScheduledExecutorService() == null && houseKeepingExecutorService != null) {
+         addConnectionExecutor.shutdown();
+         addConnectionExecutor.awaitTermination(5L, SECONDS);
+
+         if (config.getScheduledExecutorService() == null) {
             houseKeepingExecutorService.shutdown();
             houseKeepingExecutorService.awaitTermination(5L, SECONDS);
          }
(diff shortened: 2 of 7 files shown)


nitincchauhan:
CI failure is :

java.lang.NoSuchMethodError: org.hamcrest.Matcher.describeMismatch(Ljava/lang/Object;Lorg/hamcrest/Description;)V
	at com.zaxxer.hikari.metrics.prometheus.HikariCPCollectorTest.connection1(HikariCPCollectorTest.java:54)

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -455,7 +455,7 @@ private static Connection getClosedConnection()
             public Object invoke(Object proxy, Method method, Object[] args) throws Throwable
             {
                final String methodName = method.getName();
-               if ("abort".equals(methodName)) {
+               if ("close".equals(methodName) || "abort".equals(methodName)) {
```
brettwooldridge: What's going on here?  Multiple-close is already handled -- see [``testDoubleClose()``](https://github.com/brettwooldridge/HikariCP/blob/dev/src/test/java/com/zaxxer/hikari/pool/TestConnections.java#L220).  This proxy only handles methods not intercepted by ``ProxyConnection``.  The ``close()`` method is already intercepted and is pointless here.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -445,9 +439,6 @@ private int getAndSetNetworkTimeout(final Connection connection, final long time
                if (validationTimeout < SECONDS.toMillis(1)) {
                   LOGGER.warn("{} - A validationTimeout of less than 1 second cannot be honored on drivers without setNetworkTimeout() support.", poolName);
                }
-               else if (validationTimeout % SECONDS.toMillis(1) != 0) {
-                  LOGGER.warn("{} - A validationTimeout with fractional second granularity cannot be honored on drivers without setNetworkTimeout() support.", poolName);
-               }
```
brettwooldridge: Why was this removed?  This is intended to log a warning in cases where the user configures a ``validationTimeout`` with fractional second granularity, for example, ``2250ms`` but does not have a driver that supports ``setNetworkTimeout()``.

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
nitincchauhan: agreed and undone

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: This is lack of (reading) documentation that leaks in source with checks and messages.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -441,13 +435,7 @@ private int getAndSetNetworkTimeout(final Connection connection, final long time
...
 
-               LOGGER.info("{} - Driver does not support get/set network timeout for connections. ({})", poolName, e.getMessage());
-               if (validationTimeout < SECONDS.toMillis(1)) {
-                  LOGGER.warn("{} - A validationTimeout of less than 1 second cannot be honored on drivers without setNetworkTimeout() support.", poolName);
-               }
-               else if (validationTimeout % SECONDS.toMillis(1) != 0) {
-                  LOGGER.warn("{} - A validationTimeout with fractional second granularity cannot be honored on drivers without setNetworkTimeout() support.", poolName);
-               }
+               LOGGER.info("{} - Failed to get/set network timeout for connection. ({})", poolName, e.getMessage());
             }
```
nitincchauhan: Pool logs failure, e.getMessage() has actual reason.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -519,16 +507,15 @@ private void createNetworkTimeoutExecutor(final DataSource dataSource, final Str
...
-    * @param connectionTimeout the timeout in milliseconds
     */
-   private void setLoginTimeout(final DataSource dataSource, final long connectionTimeout)
+   private void setLoginTimeout(final DataSource dataSource)
    {
       if (connectionTimeout != Integer.MAX_VALUE) {
          try {
-            dataSource.setLoginTimeout((int) MILLISECONDS.toSeconds(Math.max(1000L, connectionTimeout)));
+            dataSource.setLoginTimeout(Math.max(1, (int) MILLISECONDS.toSeconds(500L + connectionTimeout)));
          }
```
nitincchauhan: as for setNetworkTimeout() rounding off to 'nearest' second

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -243,8 +243,8 @@ public long getValidationTimeout()
    @Override
    public void setValidationTimeout(long validationTimeoutMs)
    {
-      if (validationTimeoutMs < 250) {
-         throw new IllegalArgumentException("validationTimeout cannot be less than 250ms");
+      if (validationTimeoutMs < 1000) {
+         throw new IllegalArgumentException("validationTimeout cannot be less than 1000ms");
       }
```
nitincchauhan: for setNetworkTimeout() as in docs, accepts ms but driver should round off to 'nearest' second.
here safe guarding less than 1 second values, faulty driver may round it to ZERO and treat as 'no timeout'

src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -229,16 +224,15 @@ public final void close() throws SQLException
...
             }
 
             if (dirtyBits != 0) {
                poolEntry.resetConnectionState(this, dirtyBits);
-               lastAccess = clockSource.currentTime();
             }
 
             delegate.clearWarnings();
+            lastAccess = clockSource.currentTime();
          }
```
nitincchauhan: setting lastAccess here.. only after 'successful' any sql calls before and lastly clearWarnings() ...closest before recycling

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -493,14 +490,14 @@ public void run() {
    private void abortActiveConnections(final ExecutorService assassinExecutor)
    {
       for (PoolEntry poolEntry : connectionBag.values(STATE_IN_USE)) {
+         Connection connection = poolEntry.close();
          try {
```
nitincchauhan: ...cancels leak task first. this pattern should be followed 'consistently'
proxyConnection.cancelLeakTask() looks like redundent / patchwork and can be removed with refactoring

src/main/java/com/zaxxer/hikari/HikariConfig.java
brettwooldridge: You can't do this, this was specifically changed for a reason:

https://github.com/brettwooldridge/HikariCP/issues/547

Please check the "Blame" history of a file before changing particular lines.  The blame for this line would show this bug specifically.

nitincchauhan: that issue was with connectionTimeout. linking it with validationTimeout is wrong.

brettwooldridge: That is incorrect, you need to read the entire thread.  Having a ``validationTimeout`` that is higher than the ``connectionTimeout`` leads to violations of the ``connectionTimeout`` contract.  The ``violationTimeout`` absolutely must be able to be set to an equal or lesser value than ``connectionTimeout``.

I am going to cherry-pick the changes in this pull request rather than discussing back and forth on every change.

brettwooldridge:
Cherry-picked.

PR #656, opened 2016-06-24
Rollback failure when both the isAutoCommit and isIsolateInternalQueries are false.

kenny319:
The autoCommit should be changed  before checkDiverSupport, otherwise once both the isAutoCommit and isIsolateInternalQueries are false the rollback operation on a true connection will fail.


brettwooldridge:
Cherry-picked.

PR #680, opened 2016-07-17
Implement List directly instead of extending ArrayList.

johnou:
Closes https://github.com/brettwooldridge/HikariCP/issues/672 .

PR #681, opened 2016-07-17
Fix “NoSuchMethodError: org.hamcrest.Matcher.describeMismatch” in Hik…

PascalSchumacher:
…ariCPCollectorTest#connectionClosed by using mockito-core instead of mockito-all and declaring the hamcrest dependencies before the mockito and junit dependencies.

See http://stackoverflow.com/questions/7869711/getting-nosuchmethoderror-org-hamcrest-matcher-describemismatch-when-running for details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,12 +99,10 @@
          <version>${commons.csv.version}</version>
          <scope>test</scope>
       </dependency>
-      <dependency>
-         <groupId>org.mockito</groupId>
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
Closed, https://github.com/brettwooldridge/HikariCP/pull/680/commits/b45e9fc6c60029e0f58969d9561c35ef7104a212 is a better solution for this problem.

PR #690, opened 2016-08-04
Fix raise-conditon with MySQL replication driver - set readOnly and autoCommit before setting timeouts for validation

nilo85:
This is cause when using MySQL replication driver, the result when default is readOnly, the master connection keeps the validation timeout.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -332,6 +332,10 @@ abstract class PoolBase
     */
    private void setupConnection(final Connection connection) throws SQLException
    {
+      checkDriverSupport(connection);
+      connection.setReadOnly(isReadOnly);
+      connection.setAutoCommit(isAutoCommit);
+
       if (networkTimeout == UNINITIALIZED) {
          networkTimeout = getAndSetNetworkTimeout(connection, validationTimeout);
       }
@@ -339,10 +343,6 @@ abstract class PoolBase
          setNetworkTimeout(connection, validationTimeout);
       }
 
-      checkDriverSupport(connection);
-
-      connection.setReadOnly(isReadOnly);
-      connection.setAutoCommit(isAutoCommit);
 
       if (transactionIsolation != defaultTransactionIsolation) {
          connection.setTransactionIsolation(transactionIsolation);


src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -332,17 +332,17 @@ Connection newConnection() throws Exception
     */
    private void setupConnection(final Connection connection) throws SQLException
    {
+      checkDriverSupport(connection);
```
nilo85: Not sure if this should be executed before or after timeout is set, I guess now it might allow longer execution in checkDriverSupport SQL. Maybe move this down again?

nilo85:
Example code to show this problem:
`import com.zaxxer.hikari.HikariConfig;
import com.zaxxer.hikari.HikariDataSource;

import java.sql.Connection;
import java.sql.SQLException;

import static java.util.concurrent.TimeUnit.SECONDS;


public class ExampleCase {



    public static HikariConfig createConfiguration() {

        HikariConfig c = new HikariConfig();

        c.setUsername("the_username");
        c.setPassword("the_password");
        c.setRegisterMbeans(true);

        c.setValidationTimeout(SECONDS.toMillis(1));
        c.setConnectionTimeout(SECONDS.toMillis(5));
        c.setIdleTimeout(SECONDS.toMillis(600));
        c.setMaxLifetime(SECONDS.toMillis(720));
        c.setDriverClassName("com.mysql.jdbc.Driver");

        return c;
    }

    public static void main (String[] args) throws SQLException {

        HikariConfig c = createConfiguration();

        c.setPoolName("user_pool_name");
        c.setMinimumIdle(1);
        c.setMaximumPoolSize(1);
        c.setJdbcUrl("jdbc:mysql:replication://master,slave/database?allowMasterDownConnections=true&connectTimeout=500&socketTimeout=6000");

        c.setReadOnly(true);
        c.setConnectionTimeout(5000);
        c.setLeakDetectionThreshold(0);

        HikariDataSource hikariDataSource = new HikariDataSource(c);

        Connection connection = hikariDataSource.getConnection();

        connection.setReadOnly(true);
        connection.setAutoCommit(false);
        // 6.000
        System.out.println(connection.getNetworkTimeout());
        connection.createStatement().execute("SELECT SLEEP(5)");
        connection.commit();

        connection.close();

        connection = hikariDataSource.getConnection();

        connection.setReadOnly(false);
        connection.setAutoCommit(false);
        // 1.000
        System.out.println(connection.getNetworkTimeout());

        //BANG, Socket read timeout
        connection.createStatement().execute("SELECT SLEEP(5)");
        connection.commit();


    }
}`

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: checkDriverSupport() may issue commit/rollback :) keep it after setting networkTimeout

brettwooldridge:
@nilo85 Huh?  This can never work.  By setting ``socketTimeout=6000`` in the database URL you are saying that no socket can sit idle for longer than 6sec before it is terminated.

If you insist on setting a global ``socketTImeout`` value in the URL, it must be &gt;=``idleTimeout `` and greater than your longest running query.

nilo85:
Well. Thats good feedback regarding idleTimeout, however, the sad composed behaviour with MySQL replication deiver is still broken.

HikariCP does sets connection timeout during validation, however, it has not set readOnly yet, so it will set it to the replication drivers master connection.

When HikariCP later resets the timeout to the intended one for use, it has set readOnly=true and therefore updates the slave connection, resulting in that master has timeout of validation, and slave has intended timeout.

nilo85:
@brettwooldridge actually your statement about socket timeout is incorrect. Socket timeout does not affect idle connections, only applies once you do a blocking read, and MySQL driver seems to do this when executing a query, not having a thread always reading data.

Please reevaluate pull request 

http://docs.oracle.com/javase/7/docs/api/java/net/Socket.html#setSoTimeout%28int%29

brettwooldridge:
@nilo85 I understand that MySQL is strange, but the network timeout is there to protect the pool (and applications) from a network partition event.  There are many other databases besides MySQL, and the guarantees that the pool provides should not be compromised to compensate for an implementation quirk of one driver vendor.

If interaction with the database, for example ``setReadOnly()`` or ``setAutoCommit()``, occurs at the time of a network partition, and outside of a network timeout, the pool could potentially hang until the OS level TCP timeout occurs (many minutes later).  So ``setReadOnly()`` *cannot* not be moved above the setting of the timeout.  It is a quirk of MySQL that it uses ``setReadOnly()`` to switch to the replication server (when ``allowMasterDownConnections=true``).

Why start the pool read-only?  According to MySQL's ReplicationConnection.setReadOnly() code the master connection can't be set to read-only anyway, unless there are no slaves.

brettwooldridge:
@nilo85 There are so many issues with the way that MySQL has chosen to implement that feature.  Basically none of the calls in ``setupConnection()`` have any meaning because they are not applied to both connections, but instead only the "current" one.

nilo85:
@brettwooldridge Ok thanks for explaining. I agree, MySQL is not the only one available, and not my first choice anyway =)

The reason to set readOnly=false per default is to survive master failure. 

Hikari will not be able to create new connections in the pool if master is down and readOnly is not set prior to validation.

We have now went back to default readOnly=false and we investigate removing this buggy replication driver altogether.

PR #693, opened 2016-08-09
mysql driver's bug on isValid()

wantedlzx:
Hi, if we use mysql driver, in some condition, there will exist huge of "MySQL Statement Cancellation Timer" threads.
How to reproduce:
1. get a connection and execute statement.setQueryTimeout(xxx) to make sure cancelTimmer will start.
2. make sure that server has closed the connection.
3. use isValid() to check whether a connection is alive or not
4. mysql JDBC4Connection call pingInternal(), if failed, will do abortInternal(), and in this method, it will set isClosed true.
5. when close this connection, ConnectionImpl will call realClose(), but if isClosed == true, it will return and cancelTimer is not cancelled.

I do something to fix this bug. When we find datasoure or jdbcURL or dsclassname contains "mysql", we will directly use pingInternal().
pingInternal() is a void method, it throw exception rather than return false, so we catch this exception and do nothing except return false.
(Mysql isValid() catch this exception and set isClosed true. Further details see #JDBC4Connection.isValid(int timeout)). 

All the tests still pass.

# refs : https://bugs.mysql.com/bug.php?id=82500

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -20,6 +20,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 import java.lang.management.ManagementFactory;
+import java.lang.reflect.Method;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLTransientConnectionException;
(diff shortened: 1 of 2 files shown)


brettwooldridge:
I appreciate the pull request, but the place this should be fixed is MySQL.  We're tired of hacking around their bugs.  The workaround is to use a validation query instead of isValid().

johnou:
You could try using the validation query.. 

/* ping */ SELECT 1

PR #697, opened 2016-08-13
Fallback to random pool number in security sandbox

garricko:
Prevent failures when using default pool name if the security manager is restricting access to system properties.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -40,7 +40,6 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 
-import java.util.concurrent.atomic.AtomicInteger;
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -55,7 +54,6 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long IDLE_TIMEOUT = MINUTES.toMillis(10);
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
    private static final int DEFAULT_POOL_SIZE = 10;
-   private static final AtomicInteger poolNumberCounter = new AtomicInteger(1);
 
    private static boolean unitTest = false;
 


src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -880,11 +882,17 @@ protected void loadProperties(String propertyFileName)
 
    private int generatePoolNumber()
    {
-      // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
-      synchronized (System.getProperties()) {
-         final int next = Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1;
-         System.setProperty("com.zaxxer.hikari.pool_number", String.valueOf(next));
-         return next;
+      try {
```
johnou: Instead of catching AccessControlException why not something like this..?

```java
 Integer next = AccessController.doPrivileged(new PrivilegedAction<Integer>() {
                @Override
                public Integer run() {
                    // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
         synchronized (System.getProperties()) {
            final int next = Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1;
            System.setProperty("com.zaxxer.hikari.pool_number", String.valueOf(next));
            return next;
         }
                }
            });
```
If null fallback to static atomic integer.

garricko: Yes, it would probably be better to wrap this whole block into a doPrivileged(). I also don't like that if requires read access to *all* properties just for the purpose of synchronization, while the synchronization seems a little suspect in that I don't think there is a guarantee it will always be the same instance.

brettwooldridge:
Is this really a thing, why should HikariCP support execution in such a restricted environment?

johnou:
Yes, security managers are really starting to take off, I would suggest moving all system property access to a dedicated class eg. SystemPropertyUtil and invoke PrivilegedAction if there is a security manager defined. I am happy to work on this in a few weeks when I get back from holiday if no one else picks it up.

garricko:
@brettwooldridge: It seems HikariCP already supports sandboxing fairly well. Generally reading system properties isn't an issue because all libraries do it, so allowing read access is a given. The problem with this code is it requires PropertyPermission("*", "read,write"), which, given the global nature of System allows for all kinds of mischief.

@johnou: After giving it a bit more thought, I think it might be better to just allow this code to fail with the static counter fallback. Wrapping it in doPrivileged() poses other risks because you might be allowing abuse of privileges granted only to HikariCP (doPrivileged runs will all of the callers permissions, not just the property related ones). I updated the pull request to use a counter.

acdickinson:
Is this pull request going to be merged and released anytime soon?

billoneil:
You can easily work around this by setting the pool name explicitly which would never call `generatePoolNumber()` . Is this not sufficient enough for more restricted environments?

See https://github.com/brettwooldridge/HikariCP/blob/5e05adcd27ceeaa4b6b5a23930505f99ff3b0644/src/main/java/com/zaxxer/hikari/HikariConfig.java#L744

Are security managers becoming popular server side or only mobile?

garricko:
I did end up working around this in my database library (https://github.com/susom/database/commit/6757a3645646a14c65ee668b3fd3090ad7759f05). Using the security manager on the server side adds a meaningful level of protection in my view, but it is challenging for reasons like this. Making libraries require less authority, and better tolerate the lack of authority is a helpful step.

brettwooldridge:
@garricko @billoneil @acdickinson This pull request is still open, though needs rebasing if it is going to merge.  Thoughts on its necessity?  Is it required, or are available workarounds sufficient?

billoneil:
I have no use case for this.

garricko:
Is there a concern with providing some reasonable fallback? I was able to work around it, but I wasted time trying to figure out why it was failing, so the next person might as well. It seems like better and more desirable behavior for the library to handle the error.

brettwooldridge:
@garricko I would accept a pull request that, in the case of rejection by a SecurityManager, falls back to a random 4 character alpha-numeric string, prefixed by ``HikariPool-``, and logs an INFO level message that the fallback occurred due to SecurityManager restrictions.

johnou:
@garricko you got this covered?

garricko:
@brettwooldridge I implemented it the way you suggested, but I messed up the rebase here, so I started over with clean PR #988 and am closing this one.

PR #702, opened 2016-08-16
replace System.out.println with logger

diguage:


brettwooldridge:
@diguage JavassistProxyFactory is only used at build-time, is this change necessary or does it prevent output from appearing during the build?

diguage:
Sorry. I just started to learn the source, and did not understand the usage of `JavassistProxyFactory`.

PR #705, opened 2016-08-27
Programmatically connection initializer

krupt:
I use JDBC feature for SQL types in Oracle 11g. So, I need to register mappings (typeName -> JavaClass) for every connection once. My use-case:
```
HikariConfig hikariConfig = new HikariConfig();
...
hikariConfig.setConnectionInitializer(connection -> {
    try {
        Map<String, Class<?>> typeMap = connection.getTypeMap();
        for (Class<?> classToRegister : findOracleTypeClasses()) {
            if (SQLData.class.isAssignableFrom(classToRegister)) {
                try {
                    SQLData sqlData = (SQLData) classToRegister.newInstance();
                    String sqlTypeName = sqlData.getSQLTypeName();
                    typeMap.put(sqlTypeName, classToRegister);
                    log.info("SQLType ({}) registered as {}", sqlTypeName, classToRegister);
                } catch (ReflectiveOperationException e) {
                    throw new ConnectionInitializationException("Persistent " + classToRegister + " doesn't have default constructor", e);
                } catch (SQLException e) {
                    throw new ConnectionInitializationException("Getting SQLTypeName of persistent " + classToRegister + " failed", e);
                }
            }
        }
    } catch (SQLException e) {
        log.error("Couldn't initialize connection", e);
        throw new ConnectionInitializationException(e);
    }
});
```

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -22,11 +22,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
+import java.sql.Connection;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
+import java.util.function.Consumer;
 
 import javax.naming.InitialContext;
 import javax.naming.NamingException;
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@krupt HikariCP [used to have](https://github.com/brettwooldridge/HikariCP/blob/2.3.0/hikaricp-common/src/main/java/com/zaxxer/hikari/AbstractHikariConfig.java#L197) a programmable initialization point.  It was removed.

You [can read the explanation and discussion here](https://groups.google.com/d/msg/hikari-cp/9n-N5W0FIOI/2EHZefJw3jcJ).

If you want to execute code for every new connection, once, then do so in a portable way that allows you to switch connection pools with only a configuration change rather than a code change.  See [this gist](https://gist.github.com/brettwooldridge/dfd21934416fc2a3c865) for one possible pattern -- there are others as well.

PR #723, opened 2016-09-17
Unwrap chained exceptions.

johnou:
Calling sqle.getNextException() will return the same exception as sqle is never reassigned in the loop.


johnou:
Thanks!

johnou:
@brettwooldridge unfortunately this is not complete, stay tuned for follow up merge request.

PR #725, opened 2016-09-18
Enable to create HikariConfig from Map properties

amanteaux:
A small enhancement for the `HikariConfig` API: since the `Properties` class already implements `Map<Object, Object>`, there is almost nothing to change.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -16,6 +16,10 @@
 
 package com.zaxxer.hikari;
 
+import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
@@ -40,11 +44,6 @@ import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.zaxxer.hikari.metrics.MetricsTrackerFactory;
 import com.zaxxer.hikari.util.PropertyElf;
-
-import static java.util.concurrent.TimeUnit.MINUTES;
-import static java.util.concurrent.TimeUnit.SECONDS;
-
-import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 
 public class HikariConfig implements HikariConfigMXBean
 {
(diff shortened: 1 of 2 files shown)


brettwooldridge:
What are the merits of this change?  Even though it seems like a harmless change, because it is a widening of the type, it consititutes a break of binary compatibility (see [Evolving Java Based APIs](http://wiki.eclipse.org/Evolving_Java-based_APIs)).

amanteaux:
Indeed, I had not thought about the binary compatibility. It should be ok now.

My use case is that I am configuring HikariCP with the [Config](https://github.com/typesafehub/config) library. So when I fetch Hikari properties from my configuration file, I have a `Map<String, String>` which I need to convert into a `Properties` object to configure Hikari. If the Hikari API also accepts properties as `Map`, then I will not need to do this conversion.

johnou:
Conversion isn't that painful..?

```java
Properties propertiesConfig = new Properties();
propertiesConfig.putAll(mapConfig);
```

amanteaux:
Sure the conversion is trivial.
However, when you want to check if a `String` is empty, what do you do?
- `if(string == null || string.equals(""))`
- or `Strings.isNullOrEmpty(string)` or the Apache equivalent `StringUtils.isEmpty(string)` ?

As a developer, we try to avoid repetition as much as possible following the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle.

brettwooldridge:
@amanteaux I was initially skeptical because of binary compatibility, but in a general sense I agree with @johnou.  HikariCP tries to expose the absolute minimum API required.  For example, some pools  have offered convenience methods such as:
 * ``setIdleConnectionTestPeriodInSeconds(int secs)``
 * ``setIdleConnectionTestPeriodInMinutes(int mins)``
 * ``setIdleConnectionTestPeriod(long period, TimeUnit unit)``

But we reject that kind of widening of the API.  We offer only *one value type* per-setter.  Our time signatures take milliseconds, it is up to the caller to convert if desired:
```java
config.setIdleTimeout(TimeUnit.MINUTES.toMillis(5));
```
Similarly, we take ``Properties`` where best practices dictates ``String`` as a key.  Of course, because ``Properties`` extends ``Hashtable`` (which implements ``Map``) it is possible for users to violate the sematics, but at least the semantics are clear.  Taking a ``Map<?, ?>`` is far too broad and provides no indication to the user regarding the expectation.

One more (unnecessary) method is one more method to document, provide test coverage for, and provide support when users do stupid things.  We are already drowning in trivial questions from newbie developers without adding more ways for them to get tangled.

amanteaux:
@brettwooldridge Ok I understand, thank you for detailed response.

PR #767, opened 2016-11-23
added HikariDataSourceConfigurable

iobestar:
My use case:
I want to be able to configure custom data source created by Hikari from specified dsClassName. Currently, I am not able to do that because PropertyElf sets properties to late.

If this is not good way to solve my issue, I am open for suggestions.


lfbayer:
This is not something that HikariCP will implement. Your use-case can be satisfied by calling setDataSource on HikariConfig directly with a pre-initialized DataSource.

PR #772, opened 2016-11-29
Added logging to the connection-pool initialization.

kny78:
As the code is now, it will only be logged to debug if the initialization of the connection is failing. 
If we set initializationFailTimeout to Long.MAX_VALUE, and disable debug, then we will not be notified that the connection-initialization is hanging.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -515,6 +515,7 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
             if (!connection.getAutoCommit()) {
                connection.commit();
             }
+            LOGGER.info("Initialization of Connection is complete!");
             return;
          }
          catch (ConnectionSetupException e) {
@@ -525,6 +526,8 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
             throwable = t;
             quietlySleep(1000L);
          }
+         long remainingMillisUntilTimeout = config.getInitializationFailTimeout() - clockSource.elapsedMillis(startTime);
+         LOGGER.warn("Could not aquire connection during initialization. Will retry for {} milliseconds more.", remainingMillisUntilTimeout);
       } while (clockSource.elapsedMillis(startTime) < config.getInitializationFailTimeout());
 
       throw new PoolInitializationException(throwable);


brettwooldridge:
I will have to consider this.  We used to log connection attempt failures at info level, but users complained that it filled up their logs. Believe it or not, some users allow their application to "pend" on the pool initialization for hours or days, and do not want HikariCP logging every attempt.

polpot78:
That sounds crazy!
But if they want to do this, they should maybe just filter out log messages?

But for us it is very important to get this warning if the initialization is failing. The alternative is that we will not understand what is wrong, and we will not get back up'n'running in a timely manner....

andersha:
Even if it is cumbersome the logging in question could be made into a configuration parameter to HikariConfig since we have conflicting needs; some users want to idle silently forever and others want to know about the connection timeout asap (fail fast). Controlling this through logger configuration is problematic since you can silence other important log messages from the same code.

brettwooldridge:
@andersha If the database goes down, the pool will quickly drain, and callers to ``getConnection()`` will receive SQLTransientConnectionExceptions whose ``getNextException()`` method returns the underlying driver exception.  How is this not sufficient in your case?

andersha:
Disclaimer: It is nearly one year since I made the comment, so my recollection might not be 100%. 

I do not think the issue was related to the database going down while running (which I think most users will handle fine) but the fact that we silently do nothing if we cannot connect to it when starting up. We use Hikari as part of the java version of play framework and when we start a component it sometimes just freeze until it eventually times out with a console error that it cannot connect to database. Usually this is a configuration issue on the database or network side. 

Since I am a fan of fail-fast strategies a log line from the application which again would propagate to our centralised log-server where we could trap it and trigger an alarm immediately would be very helpful, but then we need that log line.

Should it be a WARN, or should it just be logged on DEBUG/INFO(?). Maybe it could be configurable if you want to see that log or not? Of that I am not sure..

brettwooldridge:
@andersha There is the [initializationFailTimeout](https://github.com/brettwooldridge/HikariCP#infrequently-used), which should fire immediately (by default) upon startup if the connection cannot be established, and [log something at ERROR-level](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L546).

I guess that doesn't address @polpot78 's use-case where they set ``initializationFailTimeout`` to ``Long.MAX_VALUE``, and expect to see error messages.  I *still* don't quite understand *that* use-case...

andersha:
@brettwooldridge Fair point, we could control this using the various timeouts. BTW, I did not notice until you pointed it out that the PR code change is inside the initialization part and not the connect part. I agree that this is a marginal use case.

lfbayer:
I'm closing this pull request as there hasn't been movement on it for almost 6 years now.

Issue #774, opened 2016-11-29
Number of Connections vs Thread Count


samhendley:
I have been really impressed with Hikari in general, much better than the other pools we have tried. 

We have been trying to characterize the behavior of the pool in regards to how many connections it makes to the database. In our case we notice that the pool hits our maxPoolSize limit of 100 connections when the processing is very fast. I did an experiment where I tried different thread counts vs. simulated operation time vs. idle time between operations to try to understand how many connections are made to the database. Full Test code is here https://gist.github.com/samhendley/d51b8ebba59e348ee10d13e6c16d2820 but core of the test is this loop run by many threads.

```java
while (running.get()) {
    final Connection connection = hikariDataSource.getConnection();
    if (operationTime > 0) {
        Thread.sleep(operationTime);
    }
    connection.close();
    if (idleTime > 0) {
        Thread.sleep(idleTime);
    }
}
```

What we find is that if the operationTime is short and the idle time is short (simulating a busy and fast process) we will quickly spawn our max of 100 connections even with only 8 threads. 

![image](https://cloud.githubusercontent.com/assets/36385/20720634/f05782a0-b62d-11e6-8899-87061d7e059b.png)

This was a bit of a surprise to me. I was expecting the pool to be end up maxing out at around the same number of threads we have but it looks like there must be some "housekeeping" I wasn't aware of that lets the pool grow too large. I know I could change the maxPoolSize to match the thread count but that can be tricky to "plumb" and if we try to be too exact we can create deadlock type failures. Notice that when the either the operations or idleTime are "slow" the pool stays at about the expected size. We only noticed this when we had optimized the performance of our process so all of it's calls fell into the <10ms range.

So two questions: 

* Can you explain what is happening after the connection.close() so that it isn't immediately ready for reuse. (is the housekeeping thread running the connectionQuery?)

* Any settings other than maxPoolSize I can change to keep this growth in check?

brettwooldridge:
@samhendley Your issue was an inspiration for a [new write-up](https://github.com/brettwooldridge/HikariCP/blob/dev/documents/Welcome-To-The-Jungle.md) to be published when v2.6.0 is released.

samhendley:
Awesome write up and thank you for the help. If you can, you might want to change the color of "waiting threads" or "total connections". They were really difficult to tell apart on my monitor and phone so it took a while to understand what they are showing me.

[...]

PR #776, opened 2016-11-29
Avoid creating too many connections in bursty workloads 

samhendley:
See Issue #774 for backstory. Turns out in a bursty mode where the actual work being done is short it is pretty easy to hit maxPoolSize connections. Once connections are made they are quickly cycled through so they don't appear to age out and the pool will stay at max size while there is a consistent set of work to be done.

See javadoc in middle of PoolEntryCreator for in-depth description of fix. 

Notes
- RaceCondition test was accidently disabled, didn't actually use the pool and had no assertions to check that it was doing anything
- RampUp test was racy and would fail if unlucky
- Main change ended up being pretty simple but I can't 100% convince myself that it is perfect. I agree it is better to make more connections than too few.
- Last commit to switch getPendingQueue() to use the atomic.get() is optional but I think it is more accurate because in a heavily trafficed pool all of the threads are waking up and checking all of the objects (only to be disappointed) most of the time

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -77,7 +77,8 @@ public class HikariPool extends PoolBase implements HikariPoolMXBean, IBagStateL
    private final long ALIVE_BYPASS_WINDOW_MS = Long.getLong("com.zaxxer.hikari.aliveBypassWindowMs", MILLISECONDS.toMillis(500));
    private final long HOUSEKEEPING_PERIOD_MS = Long.getLong("com.zaxxer.hikari.housekeeping.periodMs", SECONDS.toMillis(30));
 
-   private final PoolEntryCreator POOL_ENTRY_CREATOR = new PoolEntryCreator();
+   private final PoolEntryCreator POOL_FILLER_ENTRY_CREATOR = new PoolEntryCreator(false);
+   private final PoolEntryCreator POOL_USER_ENTRY_CREATOR = new PoolEntryCreator(true);
    private final ThreadPoolExecutor addConnectionExecutor;
    private final ThreadPoolExecutor closeConnectionExecutor;
    private final ScheduledThreadPoolExecutor houseKeepingExecutorService;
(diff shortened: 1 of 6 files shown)


brettwooldridge:
@samhendley Sorry, I've been hammered at work this week.  I'll take a look at this over the weekend.  At first glance it looks good.

brettwooldridge:
Merged and modified in commit f129797.  All tests are passing, including new ``ConnectionPoolSizeVsThreadsTest``.

samhendley:
Another thought I have is that we could add a "minTimeBetweenConnections" to throttle how fast we will attempt to connect to the database. Would help mitigate "thundering herd" problem where a sudden burst of traffic causes us to create lots of database connections rather than satisfying any requests and then having too many left over connections.

samhendley:
The test I originally created didn't really cover the bursty workload I was most interested in and also didn't notice if extra connections were spawned after the main part of the test was over. If we have alot of threads that all need to do a short chunk of work (and short means much less than the connection acquisition time)  it is easy for all of the work to be done before we have even created more than 1 or 2 connections. Without a check before we create a new connection we can end up creating an extra connection per original thread. 

Best way I've found to think of it is imagine I have 50 threads needing to do a single 1 ms db call. Imagine also that a new connection takes 100 ms to fully initialize. Starting from an empty pool, all 50 threads will look for a connection, all will see an empty pool then request a connection be made and continue waiting on bag. The connection thread will make the first connection in 100ms and insert into bag, a random thread will get connection, do 1ms of work and return. All 49 threads will complete their work in the time it takes the connection thread to initialize the second connection. By that point there is no more work to do and the connection thread can ignore creating the other 48 connections.

I think this also slightly speeded up the implementation since the relatively expensive getPendingQueue() and getQueue() calls are out of the "caller thread" and only done on the connection allocation thread (which should be rare). I think it should be ok to enqueue "excess" pool creation requests since in any steady state heavy load the pool will quickly reach the correct size and connections will not be needed to be created. If there isn't heavy load there won't be enough requests to cause any harm.

brettwooldridge:
@samhendley Thanks for the updated test and code.  Can you rebate your changes on the head of dev; GitHub is reporting conflicts?

Issue #769, opened 2016-11-25
New metric to measure time to establish connections.


stottpaul:
Could we have a metric that measures time to establish physical DB connections, this will help us to take a view on the cost of establishing/recycling connections, which will influence decisions as to how we configure the rest of the pool.

samhendley:
I think we can probably consider this a dupe of #518. If we add a historgram type metric it would cover both the count and how long it is taking

brettwooldridge:
Fixed by 1807776.

Issue #518, opened 2015-12-09
metrics: total borrowed/created connections


abiacco:
I'm testing a move from dbcp2 to hikari. Thanks for writing another alternative for pooling.
In dbcp2 i've been monitoring my pools via JMX for borrowedCount (the amount of times connections have been borrowed/used) and createdCount (the amount of times a connection is created).
These are running counts, and i do deltas on the counts for my monitoring.
I'm wondering if we can get something like this in Hikari.
Thanks.

brettwooldridge:
@abiacco What does the delta of these two values tell you?  I.e. how is that metric informative?

brettwooldridge:
@abiacco You can get *one* of those values, the number of times connections have been used.  If you enable [Dropwizard Metrics](https://github.com/brettwooldridge/HikariCP/wiki/Dropwizard-Metrics), the "pool.Usage" metric is a ``com.codahale.metrics.Histogram``, which has a ``count`` value indicating the number of borrow/return pairs.

We don't track the number of connection creation events.  I am still curious how the delta of these two values is useful for monitoring.

samhendley:
+1 for number of creation events. By monitoring that counter I am able to determine that we are creating and dropping more connections than I expect.  It is a hugely useful counter to have.

samhendley:
Are you opposed to this being added? I am working on a PR that will fix the pool overfill issue and this is really the metric I need to verify that the fix is working correctly

brettwooldridge:
Send a pull and we'll take a look.

brettwooldridge:
@abiacco @samhendley In HikariCP v2.6.0 there is a new DropWizard/Prometheus metric for connection creation.  In DropWizard, the metric is called ``<poolname>.pool.ConnectionCreation``. In Prometheus, the metric is ``hikaricp_connection_creation_millis``.

PR #778, opened 2016-12-03
#518 and #769 add a "Connect" histogram metric 

samhendley:
Can be applied with PR #776 or separately.


brettwooldridge:
Merged with modifications as 346a2f5.

Issue #360, opened 2015-07-31
Is 2.4.0 binary compatible with 2.3.x stream?


dotta:
(I've failed to find a mailing list where I could ask the question. Please, feel free to close this ticket and redirect me to it if one exists.)

Hi,

I'm wondering if the 2.4.0 release is binary compatible with 2.3.x. And, more in general, if you follow [semantic versioning](http://semver.org/).

Thanks!

dotta:
Ah, actually, it looks like I already have an answer, as I'm seeing the following problem after upgrading to 2.4.0

    java.lang.NoSuchMethodError: com.zaxxer.hikari.HikariConfig.setReadOnly(Z)V

Therefore, 2.4.0 is definitely not binary compatible with 2.3.x.

Still, I'm interested to know if patch releases are guaranteed (or not) to be binary compatible.

dotta:
My bad. I just noticed that the project where I tried to bump the HikariCP version uses a module that depends on `HikariCP-java6`. Hence, the linkage error I reported above is not surprising.

Hence, my initial question about semantic versioning still stands. Thanks again, and sorry for the noise.

johdah:
I got the same error when I tried to update to HikariCP 2.4 in a Play 2.4 project.

dotta:
@johdah What db module where you using? play-jdbc? play-slick?

If it was play-slick, then that's expected, since the exception in my [above comment](https://github.com/brettwooldridge/HikariCP/issues/360#issuecomment-126607673) happened after bumping the version of hikaricp in play-slick.

johdah:
@dotta I use play-jdbc.

dotta:
@johdah Thanks for the info. Then, since play-jdbc uses the hikaricp jar for java8, it seems that 2.4.0 is not binary compatible with the 2.3.x stream.

johdah:
@dotta To clarify for myself, you are talking about HikariCP-2.3 and 2.4 right? It's a little bit hard to follow sometimes since the same version numbers is used when talking about Play :confused: 

I don't really need Hikari-2.4 since I think my problems are in Play so I guess I am fine with being stuck on 2.3.x until the Play team upgrades.

dotta:
@johdah Yes, I meant HikariCP-2.4.0 seems not to be binary compatible with HikariCP-2.3.x stream. Sorry for the confusion.

[...]

PR #793, opened 2016-12-23
Change ScheduledThreadPoolExecutor to ScheduledExecutorService

lukelast:
This change is fully backwards compatible.

Using the interface allows support for many more types of executors. This is needed for wrapping, extending, instrumenting, etc. a `ScheduledExecutorService`.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -25,7 +25,7 @@ import java.lang.reflect.Modifier;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 
 import javax.naming.InitialContext;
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@lukelast Unfortunately, seemingly minor things can break in unexpected ways.  Take for example, HikariCP v2.4.0...

In v2.4.0 we changed the signature of ``HikariConfig.setReadOnly()`` from:
```java
void setReadOnly(boolean readOnly);

to

void setReadOnly(Boolean readOnly);
```
We wanted to leverage it as a 'tri-state' value internally.  The internal default would be ``null``, and therefore we could know whether the user explicitly assigned a readOnly value or not.

We assumed this could not possibly break users, because they would invoke the method as:
```java
config.setReadOnly(true);

or 

config.setReadOnly(false);
```
...and Java auto-boxing would make it transparent.  Unfortunately, this did not account for code previously compiled against HikariCP; we assumed auto-boxing would occur at runtime, but it does not and the seemingly harmless signature change [broke Play/Slick](https://github.com/brettwooldridge/HikariCP/issues/360).

We have committed to Sematic Versioning where by *any* backwards incompatible change in the ABI (Application Binary Interface) requires a change in the major version number (from 2.x.x to 3.x.x, in this case).

In the case of your change, any existing caller to ``HikariConfig.getScheduledExecutorService()`` is expecting to receive a ``ScheduledThreadPoolExecutor`` instance, not a ``ScheduledExecutorService`` instance.

We have already [deprecated a method](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/HikariConfig.java#L461) and [introduced its replacement](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/HikariConfig.java#L433) in the upcoming v2.5.2 release.  What you can do to get this pull request accepted is to deprecate the existing method (rather than removing it), and introduce the new one.  The deprecated setter method should issue a warning level log similar to the one referenced above.

lukelast:
OK that's a good point thanks. I will make the change.

brettwooldridge:
@lukelast I've already done it in commit @907da6a.

PR #796, opened 2016-12-31
Performance improvement for constrained cases threads > pool size

geneopenminder:
Performance improvement for constrained cases when number of threads much more than pool size.

ConcurrentBag will not signal another waiter after free resource if another thread already has stolen it.
This reduce context switches but make pool a little bit more unfair.

Added   @sun.misc.Contended for PoolEntry state. This reduse possible false sharing.

This changes increase performance of pool for 5-30%.

**Some benchmark results on Core i7 4 cores (8 with HT) with -XX:-RestrictContended JVM param and maxPoolSize = 8**

**_100 threads / maxPoolSize = 8_**

 _JMH 1.17.2 (released 23 days ago)
 VM version: JDK 1.8.0_92, VM 25.92-b14
 VM invoker: C:\Program Files\Java\jre1.8.0_92\bin\java.exe
 VM options: -server -XX:-RestrictContended -XX:+AggressiveOpts -XX:+UseFastAccessorMethods -Xms1096m -Xmx1096m
 Warmup: 3 iterations, 1 s each
 Measurement: 6 iterations, 1 s each
 Timeout: 10 min per iteration
 Threads: 100 threads, will synchronize iterations
 Benchmark mode: Throughput, ops/time
 Benchmark: com.zaxxer.hikari.benchmark.ConnectionBench.cycleCnnection
 Parameters: (jdbcUrl = jdbc:stub, maxPoolSize = 8, pool = hikari)_

2.5.2 version
 Run complete. Total time: 00:10:35
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt       Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48   20748,432 ▒ 1282,185  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  131634,185 ▒ 5463,157  ops/ms

**this new version**
 Run complete. Total time: 00:10:48
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt       Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48   **27079,295** ▒ **1077,770**  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  136371,239 ▒ 8264,813  ops/ms

**_64 threads / maxPoolSize = 8_**

2.5.2 version
 Run complete. Total time: 00:06:38

Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt       Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48   20348,462 ▒ 1739,307  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  125352,538 ▒ 2392,916  ops/ms

**this new version**
 Run complete. Total time: 00:06:46

Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt       Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48   **26591,769** ▒  **997,099**  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  132026,574 ▒ 3027,153  ops/ms

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
@@ -44,7 +44,11 @@ final class PoolEntry implements IConcurrentBagEntry
    Connection connection;
    long lastAccessed;
    long lastBorrowed;
+
+   //should be used with -XX:-RestrictContended JVM arg
+   @sun.misc.Contended
    private volatile int state;
+
    private volatile boolean evict;
 
    private volatile ScheduledFuture<?> endOfLife;
--- a/src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
+++ b/src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
@@ -194,7 +194,10 @@ public class ConcurrentBag<T extends IConcurrentBagEntry> implements AutoCloseab
          threadLocalList.add(weakThreadLocals ? new WeakReference<>(bagEntry) : bagEntry);
       }
 
-      synchronizer.signal();
+      //connection can be stolen already by another thread, so call signal() cause unnecessary thread unparking and context switch
+      if (bagEntry.getState() == STATE_NOT_IN_USE) {
+         synchronizer.signal();
+      }
    }
 
    /**
@@ -210,7 +213,10 @@ public class ConcurrentBag<T extends IConcurrentBagEntry> implements AutoCloseab
       }
 
       sharedList.add(bagEntry);
-      synchronizer.signal();
+
+      if (bagEntry.getState() == STATE_NOT_IN_USE) {
+         synchronizer.signal();
+      }
    }
 
    /**
@@ -301,7 +307,9 @@ public class ConcurrentBag<T extends IConcurrentBagEntry> implements AutoCloseab
    public void unreserve(final T bagEntry)
    {
       if (bagEntry.compareAndSet(STATE_RESERVED, STATE_NOT_IN_USE)) {
-         synchronizer.signal();
+         if (bagEntry.getState() == STATE_NOT_IN_USE) {
+            synchronizer.signal();
+         }
       }
       else {
          LOGGER.warn("Attempt to relinquish an object to the bag that was not reserved: {}", bagEntry);


src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
```diff
@@ -301,7 +307,9 @@ public boolean reserve(final T bagEntry)
    public void unreserve(final T bagEntry)
    {
       if (bagEntry.compareAndSet(STATE_RESERVED, STATE_NOT_IN_USE)) {
-         synchronizer.signal();
+         if (bagEntry.getState() == STATE_NOT_IN_USE) {
```
johnou: surely redundant checking the state here since it was just set..?

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
```diff
@@ -44,7 +44,11 @@
    Connection connection;
    long lastAccessed;
    long lastBorrowed;
+
+   //should be used with -XX:-RestrictContended JVM arg
+   @sun.misc.Contended
```
johnou: not sure it's a good idea to use non-public JVM specific API..?

johnou: Can you post JMH benchmark results with and without this annotation?

src/main/java/com/zaxxer/hikari/util/ConcurrentBag.java
geneopenminder: unreserve() method rare case in real lifecycle of connection. but requite() and add() should recheck state after free and it will improve performance.

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
geneopenminder: Maybe not, but unsafe, for example, used by many other libs and it's OK.
I've tested many times with contented annotation and without it. The difference is about ~2-5%.
But for case when threads = cpu cores this will be mandatory condition - reduce false sharing.
Only thing than save from it is random location of objects in memory because of they are added asynchronously and with random delays. So JVM locate them randomly in the heap often. But sometimes they can be located sequentally in memory. 
If not sun.misc.Contended than int var1,var2,vr3,var4 .... etc. 

The killer feature is recheck object state after set it free.

brettwooldridge:
@geneopenminder I appreciate where your head is at.  Unfortunately, from my perspective the jury is still out.

These are my runs with 2.5.2-SNAPSHOT and with 2.5.2-SNAPSHOT + @Contended, on my iMac Late 2013 3.4 GHz Intel Core i5 (4 cores, no hyperthreading):

## 2.5.2-SNAPSHOT

#### 4 Threads, 32 connections
*./benchmark.sh medium -t 4 -p pool=hikari*
```
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt      Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub             32  hikari  thrpt   48  28215.919 ± 1054.978  ops/ms
StatementBench.cycleStatement   jdbc:stub             32  hikari  thrpt   48  52347.473 ±  478.313  ops/ms
```

#### 32 Threads, 8 connections
*./benchmark.sh medium -t 32 -p pool=hikari -p maxPoolSize=8*
```
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt      Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48  17272.044 ±  657.641  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  78533.873 ± 7504.881  ops/ms
```

## 2.5.2-SNAPSHOT + @Contended (-XX:-RestrictContended)

#### 4 Threads, 32 connections
*./benchmark.sh medium -t 4 -p pool=hikari*
```
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt      Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub             32  hikari  thrpt   48  27589.197 ± 1372.921  ops/ms
StatementBench.cycleStatement   jdbc:stub             32  hikari  thrpt   48  58174.097 ± 7847.880  ops/ms
```

#### 32 Threads, 8 connections
*./benchmark.sh medium -t 32 -p pool=hikari -p maxPoolSize=8*
```
Benchmark                       (jdbcUrl)  (maxPoolSize)  (pool)   Mode  Cnt      Score      Error   Units
ConnectionBench.cycleCnnection  jdbc:stub              8  hikari  thrpt   48  17924.362 ±  654.138  ops/ms
StatementBench.cycleStatement   jdbc:stub              8  hikari  thrpt   48  85961.096 ± 6129.560  ops/ms
```

Conclusion: Both the contended and non-contended runs are within the margin of error of each other.  I wasn't patient enough to run ``long`` instead of ``medium``.

brettwooldridge:
@geneopenminder I'll benchmark on i7 as well.  I'm more interested now in seeing if I can get ``PoolEntry.state`` and ``PoolEntry.lastAccessed`` into the same contention pair with the optional "group" parameter to ``@Contended``.  Basically these are always accessed as a pair (even though slightly seperated by code).  This involves converting ``lastAccessed`` from a *long* to an *int*, because cache-lines are only 64-bytes on Intel and both members need to fit into the same cache-line.  ``lastAccessed`` doesn't null full precision (from epoch), just a relative time to compare to the previous access; 32-bit precision allows ~6 weeks of miliisecond ticks, which is sufficient.  I'm playing with that on a branch now to see what the effect is.

geneopenminder:
@brettwooldridge Did you test my changes with recheck object state after set it free.
Thanks,
Evgeniy.

johnou:
@geneopenminder you may want to rebase and rerun the benchmarks?

geneopenminder:
@johnou I will try to rerun on this week.

PR #842, opened 2017-03-17
Log better message when connection was evicted or dead

wvuong:
This minor change in the logging lets us know which condition (was evicted or is dead) caused the connection to be closed.  The "dead" condition is interesting to us because that means the JDBC4 connection.isValid() failed and we can better pin the blame on the network or database server.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -76,6 +76,9 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    private static final int POOL_SUSPENDED = 1;
    private static final int POOL_SHUTDOWN = 2;
 
+   private static final String EVICTED_CONNECTION_MESSAGE = "(connection was evicted)";
+   private static final String DEAD_CONNECTION_MESSAGE = "(connection is dead)";
+
    private volatile int poolState;
 
    private final long ALIVE_BYPASS_WINDOW_MS = Long.getLong("com.zaxxer.hikari.aliveBypassWindowMs", MILLISECONDS.toMillis(500));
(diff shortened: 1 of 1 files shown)


brettwooldridge:
Good point.  Definintely do not want to call ``isConnectionAlive()`` unless absolutely necessary.  Evictions/dead connections are rare enough that paying the cost of re-evaluation *inside* the original conditional would be better.

johnou:
@brettwooldridge while I am not as experienced as you with byte code optimisation could I trouble you to explain why there are two try blocks? I was thinking something like this might be "ideal"

```java
public Connection getConnection(final long hardTimeout) throws SQLException
{
   suspendResumeLock.acquire();
   final long startTime = currentTime();
   PoolEntry poolEntry = null;
   try {
      long timeout = hardTimeout;
      do {
         poolEntry = connectionBag.borrow(timeout, MILLISECONDS);
         if (poolEntry == null) {
            break; // We timed out... break and throw exception
         }

         final long now = currentTime();
         if (poolEntry.isMarkedEvicted() || (elapsedMillis(poolEntry.lastAccessed, now) > ALIVE_BYPASS_WINDOW_MS && !isConnectionAlive(poolEntry.connection))) {
            closeConnection(poolEntry, poolEntry.isMarkedEvicted() ? EVICTED_CONNECTION_MESSAGE : DEAD_CONNECTION_MESSAGE); // Throw away the dead connection (passed max age or failed alive test)
            timeout = hardTimeout - elapsedMillis(startTime);
         }
         else {
            metricsTracker.recordBorrowStats(poolEntry, startTime);
            return poolEntry.createProxyConnection(leakTask.schedule(poolEntry), now);
         }
      } while (timeout > 0L);

      metricsTracker.recordBorrowTimeoutStats(startTime);
   }
   catch (InterruptedException e) {
      if (poolEntry != null) {
         poolEntry.recycle(startTime);
      }
      Thread.currentThread().interrupt();
      throw new SQLException(poolName + " - Interrupted during connection acquisition", e);
   }
   finally {
      suspendResumeLock.release();
   }

   throw createTimeoutException(startTime);
}
```

johnou:
@brettwooldridge as there has not been further contact from wvuong I went ahead and made the changes as per our discussion https://github.com/brettwooldridge/HikariCP/pull/977

wvuong:
Hey, sorry for ghosting this - closing mine because #977 is doing it better.

PR #851, opened 2017-03-26
Fix prometheus metrics to support multiple hikaricp pools in single jvm

ankurbhakta:


brettwooldridge:
In general, I'm fine with this change.  The question is: *Is this going to break existing users monitoring via Prometheus under the current scheme?*

brettwooldridge:
Bump.

brettwooldridge:
@ankurbhakta Can you tell me what the impact of this change is?  If it is likely to break existing Prometheus users, it will need to go into the next major release of HikariCP, with a bump in the most significant digit of the version number, due to [semantic versioning](http://semver.org).

cenkakin:
@brettwooldridge we also faced this problem. Prometheus raises an exception when two collectors with same name are registered. I don't think the fix will affect current users, since it doesn't change any existing label or value. It just registers the collectors statically and continues to add labels for each hikari config. This pr can be considered as a bugfix.

brettwooldridge:
@cenkakin This change needs to be rebased onto the current head.  The original author of the change is unresponsive, so if you want a this change you can clone his repo and submit a new pull request.  I will close this request once that is received.

bluehallu:
This fix is no longer viable given that support for custom registries was added. I have a fix that involves keeping static maps that track all collectors that were registered on any registry but it looks really ugly. I believe a bad API on the Prometheus Java client is the culprit. 

I'll try to get a PR to them that allows querying for existing collectors on a registry and then fix this code using that capability which will make the bugfix a lot neater. 

How do you feel about donating the Prometheus classes to their repo altogether? I think they're in a much better position to maintain it and their repo already acts as a collection of bridges for all sorts of libraries anyway!

brettwooldridge:
I would love it if Prometheus owned that code.

cemo:
@brettwooldridge we will revert the change introduced by #876 as we agreed author of the commit since it is causing some ugly design problems. @cenkakin will take care.

cemo:
@brettwooldridge you can close this PR.  
@ankurbhakta please check #890

PR #876, opened 2017-04-26
Added support for non-default Prometheus CollectorRegistry

llucherini:
Hey.
This PR introduces a new constructor in `PrometheusMetricsTracker` and a constructor overload in `PrometheusMetricsTrackerFactory` to enable configuration of whatever Promethes `CollectorRegistry` one might want to use. It's completely backwards compatible and shouldn't affect any existing codebase relying on this.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
@@ -34,17 +34,20 @@ class PrometheusMetricsTracker implements IMetricsTracker
    private final Summary eaSummary;
    private final Summary ebSummary;
    private final Summary ecSummary;
-   private final Collector collector;
 
-   PrometheusMetricsTracker(String poolName, Collector collector)
+   private final Collector collector;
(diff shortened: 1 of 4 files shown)


cemo:
Hi @llucherini,

What was your use case to adding custom registration support? Since this current buggy form does not support multiple datasource and causing errors we need to fix this. However custom registration makes the code pretty much ugly to fix.

cemo:
Please see #851

cemo:
BTW It is worth to mention CollectorRegistr from Prometheus javadoc:

> The majority of users should use the {@link #defaultRegistry}, rather than instantiating their own.
> Creating a registry other than the default is primarily useful for unittests, or pushing a subset of metrics to the <a href="https://github.com/prometheus/pushgateway">Pushgateway</a> from batch jobs.

So I offer to revert this change which is causing ugly implementations to fix current buggy implementation.

llucherini:
Hey. The use case is exactly the one stated in Prometheus javadocs, unit testing.
For us this is a nice to have, but it's not extremely important or vital, so feel free to revert it.

cemo:
thanks @llucherini.

We will revert this change in favor or supporting multiple datasource at once.

PR #881, opened 2017-05-03
Fixed #878 Searching for driverClassName in both hikari classes class…

### LINKED ISSUE #878: Support for Driver classes present in Thread Context Class Loader
#### Environment
```

Getting "Failed to load class of driverClassName {}" when hikari jar is present in parent classloader(tomcat shared lib) and the driver class is present in Thread context class loader(individual webapps) when the pool is initialized.

Looks like the problem is in setDriverClassName of HikariConfig, it is expecting the driver class to be present in the same classloader from which hikari classes are loaded.


HikariCP version: x.x.x
```

udayshnk:
Fix for issue https://github.com/brettwooldridge/HikariCP/issues/878

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -314,13 +314,32 @@ public class HikariConfig implements HikariConfigMXBean
 
    public void setDriverClassName(String driverClassName)
    {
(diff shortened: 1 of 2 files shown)


udayshnk:
@brettwooldridge Have updated the pull request, can you please review the latest code and release the 2.4.12 branch if everything is all right.

PR #890, opened 2017-05-11
Fix prometheus metrics to support multiple hikaricp pools

cenkakin:
I reverted commit #876 and rebased #851 with minor naming changes. In addition to these, I changed 'connection timeout counter' name's suffix from 'count' to 'total' which all prometheus counters use. That might be breaking change for people who use that metric.


cemo:
@brettwooldridge we have simplified implementation by reverting custom registry support. Custom registry support is even not default modules. By the main goal of this PR is addressing missing support multi datasource. In case a multi datasource and Prometheus enabled configuration, Hikari raises an exception. Please merge this one before releasing reverted code. There is no compatibility issue since the reverted code is not released yet.

bluehallu:
I can confirm this is the same solution I was going to PR

mmimica:
This doesn't really fix it. I have corrected the test.

PR #904, opened 2017-05-28
refact by Java 8 feature

diguage:


johnou:
-1 may introduce performance regression and I see no need to introduce a stream here.

brettwooldridge:
@diguage @johnou We *do* use streams in a couple of the ``ConcurrentBag`` methods, but this particular one is called in some of the hotter code paths, so I'd rather not touch it.

@diguage If you can run the [benchmarks](https://github.com/brettwooldridge/HikariCP-benchmark), and collect a before/after comparison showing no impact, I will consider re-opening this request.

PR #905, opened 2017-05-29
Mask password in JDBC URL

dtaniwaki:
I use HikariCP in my project and set JDBC URL with password property. I found the password in JDBC URL is not masked while directly set password is. I think it's good to mask it as well. Could you consider to merge this change?


brettwooldridge:
Compile failures.

dtaniwaki:
I fixed it and forcibly pushed the commit.

dtaniwaki:
Hi @brettwooldridge I made it fixed. Could you consider to merge it?

PR #927, opened 2017-07-03
Fix test

mmimica:
The test is a false positive.
Related to https://github.com/brettwooldridge/HikariCP/pull/890.

PR #940, opened 2017-07-27
Fix Prometheus metric collector for multiple data config

cenkakin:
Hello, 

In PR #890, I intended to support multiple data source for prometheus metrics. But, it turns out it has missing parts and its tests are misleading shown in #927 . In this PR, I aimed to fix its missing parts and test. I would be grateful, if you can check it anytime you are available @mmimica, @brettwooldridge. 

Thanks in advance,

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
@@ -18,38 +18,46 @@ package com.zaxxer.hikari.metrics.prometheus;
 
 import com.zaxxer.hikari.metrics.PoolStats;
 import io.prometheus.client.Collector;
-
+import io.prometheus.client.GaugeMetricFamily;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 class HikariCPCollector extends Collector {
-   private final PoolStats poolStats;
-   private final List<String> labelNames;
-   private final List<String> labelValues;
 
-   HikariCPCollector(String poolName, PoolStats poolStats) {
-      this.poolStats = poolStats;
-      this.labelNames = Collections.singletonList("pool");
-      this.labelValues = Collections.singletonList(poolName);
-   }
+   private static final List<String> LABEL_NAMES = Collections.singletonList("pool");
+
+   private final Map<String, PoolStats> poolStatsMap = new ConcurrentHashMap<>();
 
    @Override
    public List<MetricFamilySamples> collect() {
       return Arrays.asList(
-         createSample("hikaricp_active_connections", "Active connections", poolStats.getActiveConnections()),
-         createSample("hikaricp_idle_connections", "Idle connections", poolStats.getIdleConnections()),
-         createSample("hikaricp_pending_threads", "Pending threads", poolStats.getPendingThreads()),
-         createSample("hikaricp_connections", "The number of current connections", poolStats.getTotalConnections())
+         createGauge("hikaricp_active_connections", "Active connections",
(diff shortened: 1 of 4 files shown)


src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
```diff
@@ -18,38 +18,46 @@
...
+            PoolStats::getTotalConnections)
       );
    }
 
-   private MetricFamilySamples createSample(String name, String helpMessage, double value)
-   {
-      List<MetricFamilySamples.Sample> samples = Collections.singletonList(
-         new MetricFamilySamples.Sample(name, labelNames, labelValues, value)
-      );
+   public HikariCPCollector add(String name, PoolStats poolStats) {
```
cenkakin: that's my mistake. I fixed it.

chkal:
Hey there! I discovered this pull request as I was about to provide some feedback about the current implementation of the Prometheus. Therefore, I just reviewed this PR. Here are my thoughts:

* The PR is definitely a step into the right direction. Especially `HikariCPCollector` is much cleaner.
* I think we will need something like `PrometheusMetricsTrackerFactory.close()` to do proper deregistration of all collectors. Neither the code in `master` nor the code in the PR seems to address this.
* In my application I'm not using the default registry `CollectorRegistry.defaultRegistry` but a custom one that I instantiate myself. This is currently not supported. IMHO it would make sense to supply the `CollectorRegistry` in the constructor of `PrometheusMetricsTrackerFactory` (and of cause provide a default constructor which uses `CollectorRegistry.defaultRegistry`). But this will mean that the static fields in `PrometheusMetricsTracker` won't work any more. However, I think it is important to support custom registries.

I hope this helps!?! Let me know what you think.

brettwooldridge:
Thanks for the feedback!  Both deregistration and non-default registries seem like crucial features.  Can you open an issue in the issue tracker?

cenkakin:
@chkal thank you for review. 

In the past, you can register your custom registry for collecting hikari metrics. But, it caused some design problems to fix the issue that using Prometheus metric tracker in multiple hikari data sources. Also prometheus itself does not recommend using custom registry to collect data. So, because of these, we agreed to revert adding custom registry (See PR #851)

Deregistration collectors in PrometheusMetricsTracker also can be problematic, due to the similar reasons with custom registry. On the other hand, I agree that both features should be supported in the future. This PR focuses on making current implementation run in multiple hikari data sources.

chkal:
@brettwooldridge 
Sure, I'll create an issue for tracking this.

@cenkakin 
Thanks for referencing the previous discussion in #851. I also ran into similar problems while working on the Prometheus Hibernate module which I contributed to the Prometheus Java Client project. However, I still think that there could be a way to support custom registries without using static metric fields to enforce uniqueness. But we can continue the discussion in the issue I create to track this. And I'm also optimistic about providing proper support deregistration.

However, please note that my comments shouldn't block this PR from getting merged. As I mentioned earlier, this PR is definitely a step into the right direction. So I'm +1 for merging it.

cenkakin:
@brettwooldridge what do you think? Will you merge the PR? I guess @chkal needs this to implement the improvements he mentioned.

adilfulara:
@brettwooldridge when do you think the this PR will be released ? 

I see the following error in 2.6.3 and my hope is that PR will fix the error.

`Collector already registered that provides name: hikaricp_active_connections`

brettwooldridge:
@adilfulara can you build the HikariCP dev branch and verify that it fixes the issue?

adilfulara:
@brettwooldridge ty for the quick response . Yes `2.6.4-SNAPSHOT` build from `dev` branch does fix the issue for me. 👍

adilfulara:
@brettwooldridge Gentle 🔔  😸

brettwooldridge:
This change is in the v2.7.1 release cut last week.

Issue #843, opened 2017-03-18
[question] Why HikariCP doesn't forcefully close abandoned connections?


mohsen:
```
#### Environment
HikariCP version: HikariCP-2.4.6.jar
JDK version     : 1.8.0_121
Database        : MariaDB 10.1.21
Driver version  : mysql-connector-java-5.1.38.jar
-----------------------------------------------------------------------------------------
```
I have a web application with the following configuration. I'm going out of connection on average every two weeks in a medium-traffic website.

    <bean id="internalDataSource" class="com.mysql.jdbc.jdbc2.optional.MysqlDataSource">
        <property name="user" value="${db.username}" />
        <property name="password" value="${db.password}" />
        <property name="URL" value="${db.url}" />
        <property name="cachePrepStmts" value="true" />
        <property name="prepStmtCacheSize" value="250" />
        <property name="prepStmtCacheSqlLimit" value="2048" />
        <property name="autoSlowLog" value="false" />
        <property name="logSlowQueries" value="true" />
        <property name="slowQueryThresholdMillis" value="10000" />
        <property name="dumpQueriesOnException" value="true" />
    </bean>

    <bean id="dataSource" class="com.zaxxer.hikari.HikariDataSource" destroy-method="close">
        <property name="dataSource" ref="internalDataSource" />
        <property name="metricRegistry" ref="metricRegistry" />
        <property name="maxLifetime" value="1800000" /> 
        <property name="minimumIdle" value="1" />
        <property name="maximumPoolSize" value="10" />
        <property name="connectionTimeout" value="30000" />
        <property name="poolName" value="myConnPool" />
        <property name="leakDetectionThreshold" value="45000" />
        <property name="registerMbeans" value="true" />
    </bean>

Here is the related exception part:
```
    Caused by: java.sql.SQLTransientConnectionException: myConnPool - Connection is not available, request timed out after 15765ms.
        at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:196)
        at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:147)
        at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:99)
        at org.hibernate.engine.jdbc.connections.internal.DatasourceConnectionProviderImpl.getConnection(DatasourceConnectionProviderImpl.java:122)
[...]

PR #947, opened 2017-08-02
Add configuration option to force closing a leaked connection

abrenk:
This change introduces a "leakDetectionForceClose" configuration property
to control if a connection should be forcibly closed after the leak
detection threshold.

This allows the pool to recover and makes it more robust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,12 @@ message is logged indicating a possible connection leak.  A value of 0 means lea
 is disabled.  Lowest acceptable value for enabling leak detection is 2000 (2 seconds).
 *Default: 0*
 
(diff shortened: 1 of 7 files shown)


jnehlmeier:
Please see https://github.com/brettwooldridge/HikariCP/issues/843#issuecomment-287615297

I don't think this PR will be accepted, which is actually a good thing. The application code should be fixed instead. And if a connection leak makes it into production then an important test case is missing which should be added immediately.

brettwooldridge:
Yes, this "feature" has been discussed many times.

https://groups.google.com/d/msg/hikari-cp/zBKo96xoHXI/80hSknD7zi4J

PR #984, opened 2017-09-20
Fixed ClassNotFoundException when com.codahale.metrics.MetricRegistry is not in the classpath

gavlyukovskiy:
ClassNotFoundException is thrown during `DataSource#getConnection()` when `io.dropwizard.metrics:metrics-core`  is not available in the class path.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -272,10 +272,10 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
    public void setMetricRegistry(Object metricRegistry)
    {
-      if (metricRegistry instanceof MetricRegistry) {
+      if (metricRegistry.getClass().getName().equals("com.codahale.metrics.MetricRegistry")) {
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@gavlyukovskiy Thanks for the contribution. 🥇

johnou:
@gavlyukovskiy at this point wouldn't it be safer to check for the fqdn? com.codahale.metrics.MetricRegistry and also remove the import?

gavlyukovskiy:
I've tried this first, but then I get an issue because micrometer returns not `MeterRegistry`, but subclass - `CompositeMeterRegistry`. 
In general I'd like to do `if (ClassUtils.isPresent("io.micrometer.core.instrument.MeterRegistry") && MeterRegistry.class.isAssignableFrom(metricRegistry.getClass()))` to handle subclasses, but simply forgot to do so.
If @brettwooldridge is ok with this I'm happy to do a new pull request.

gavlyukovskiy:
BTW you can not remove the import because you have to do a cast.

brettwooldridge:
@gavlyukovskiy HikariCP v2.7.2 containing this fix has been released.

PR #985, opened 2017-09-21
Add custom collectorRegistry

PuGuanyu:
In many ocations, we need a custom registry to show our metrics, so I made this PR, hope to be merged.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
@@ -18,19 +18,24 @@ package com.zaxxer.hikari.metrics.prometheus;
 
 import com.zaxxer.hikari.metrics.PoolStats;
 import io.prometheus.client.Collector;
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.GaugeMetricFamily;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
-class HikariCPCollector extends Collector {
+public class HikariCPCollector extends Collector {
 
    private static final List<String> LABEL_NAMES = Collections.singletonList("pool");
 
    private final Map<String, PoolStats> poolStatsMap = new ConcurrentHashMap<>();
+
+   private static final Set<CollectorRegistry> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
 
    @Override
    public List<MetricFamilySamples> collect() {
@@ -52,12 +57,33 @@ class HikariCPCollector extends Collector {
    }
 
    private GaugeMetricFamily createGauge(String metric, String help,
-      Function<PoolStats, Integer> metricValueFunction) {
+                                         Function<PoolStats, Integer> metricValueFunction) {
       GaugeMetricFamily metricFamily = new GaugeMetricFamily(metric, help, LABEL_NAMES);
       poolStatsMap.forEach((k, v) -> metricFamily.addMetric(
          Collections.singletonList(k),
          metricValueFunction.apply(v)
       ));
       return metricFamily;
+   }
+
+   @Override
+   public <T extends Collector> T register() {
+      if (registered.contains(CollectorRegistry.defaultRegistry)) {
+         return (T) this;
+      } else {
+         registered.add(CollectorRegistry.defaultRegistry);
+         return super.register();
+      }
+
+   }
+
+   @Override
+   public <T extends Collector> T register(CollectorRegistry registry) {
+      if (registered.contains(registry)) {
+         return (T) this;
+      } else {
+         registered.add(registry);
+         return super.register(registry);
+      }
    }
 }
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
@@ -23,6 +23,7 @@ import io.prometheus.client.GaugeMetricFamily;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -35,7 +36,7 @@ public class HikariCPCollector extends Collector {
 
    private final Map<String, PoolStats> poolStatsMap = new ConcurrentHashMap<>();
 
-   private static final Set<CollectorRegistry> registered = Collections.newSetFromMap(new ConcurrentHashMap<>());
+   private static final Set<CollectorRegistry> registered = new HashSet<>();
 
    @Override
    public List<MetricFamilySamples> collect() {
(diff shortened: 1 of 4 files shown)


brettwooldridge:
Thank you for the pull request.  I understand the use-case and will work with you on a solution.  The issue with the proposed solution is the use of a static member.  Several pools may be deployed within the same VM, and they should not be restricted to using the same registry.

PuGuanyu:
yeah, you are right.It would be better if we provide a constructor with a registry, I will make some change~

PuGuanyu:
I've just done the changes. I added a set to avoid duplicated registry(exception occurred).

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTrackerFactory.java
```diff
@@ -19,29 +19,89 @@
...
+    * If a CollectorRegistry is provided, use provided CollectorRegistry and {@link globalCollector} as default
+    */
+   public PrometheusMetricsTrackerFactory(CollectorRegistry registry) {
+      if (globalCollector == null) {
+         synchronized (PrometheusMetricsTrackerFactory.class) {
+            if (globalCollector == null) {
+               globalCollector = new HikariCPCollector();
+            }
+         }
+      }
```
brettwooldridge: [Broken Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java) pattern.

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
```diff
@@ -17,41 +17,52 @@
...
    private final Summary.Child elapsedCreationSummaryChild;
 
-   PrometheusMetricsTracker(String poolName) {
+   PrometheusMetricsTracker(String poolName, CollectorRegistry registry) {
+      if (!registered.contains(registry)) {
+         CONNECTION_TIMEOUT_COUNTER.register(registry);
+         ELAPSED_ACQUIRED_SUMMARY.register(registry);
+         ELAPSED_BORROWED_SUMMARY.register(registry);
+         ELAPSED_CREATION_SUMMARY.register(registry);
+         registered.add(registry);
```
brettwooldridge: Not thread-safe, despite use of a wrapped ``ConcurrentHashMap``.  ``registered.contains()`` and ``registered.add()`` are not collectively atomic.

src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
```diff
@@ -52,12 +57,33 @@ protected HikariCPCollector add(String name, PoolStats poolStats) {
...
    }
+
+   @Override
+   public <T extends Collector> T register() {
+      if (registered.contains(CollectorRegistry.defaultRegistry)) {
+         return (T) this;
+      } else {
+         registered.add(CollectorRegistry.defaultRegistry);
+         return super.register();
+      }
```
brettwooldridge, line 79: Not thread-safe, despite use of a wrapped ConcurrentHashMap. registered.contains() and registered.add() are not collectively atomic.

src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
```diff
@@ -52,12 +57,33 @@ protected HikariCPCollector add(String name, PoolStats poolStats) {
...
+   }
+
+   @Override
+   public <T extends Collector> T register(CollectorRegistry registry) {
+      if (registered.contains(registry)) {
+         return (T) this;
+      } else {
+         registered.add(registry);
+         return super.register(registry);
+      }
```
brettwooldridge, line 91: Not thread-safe, despite use of a wrapped ConcurrentHashMap. registered.contains() and registered.add() are not collectively atomic.

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTrackerFactory.java
```diff
@@ -19,29 +19,89 @@
...
    }
 
    /**
     * initialize and register collector if it isn't initialized yet
     */
    private HikariCPCollector getCollector() {
       if (collector == null) {
-         collector = new HikariCPCollector().register();
+         if (globalCollector == null) {
+            collector = globalCollector = new HikariCPCollector();
```
brettwooldridge: Not atomic.  Two threads can reach line 100.

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
brettwooldridge: These functions are rarely called.  Just use a simple set and synchronize the method.

PuGuanyu:
ok, I will fix these problem tonight😁

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTrackerFactory.java
PuGuanyu: I resolved this problem by adding volatile keyword for the filed.

PuGuanyu:
@brettwooldridge Hi, I've done some changes, please review~

src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
```diff
@@ -18,20 +18,26 @@
...
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
-class HikariCPCollector extends Collector {
+public class HikariCPCollector extends Collector {
```
brettwooldridge: Why should ``HikariCPCollector`` be ``public``?  This is considered an internal class.

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTrackerFactory.java
```diff
@@ -19,29 +19,91 @@
...
    private HikariCPCollector getCollector() {
       if (collector == null) {
-         collector = new HikariCPCollector().register();
+         if (globalCollector == null) {
+            synchronized (PrometheusMetricsTrackerFactory.class) {
+               if (globalCollector == null) {
+                  globalCollector = new HikariCPCollector();
+               }
+            }
+         }
```
brettwooldridge: This code is now identical to lines 65-71.  Refactor into a shared ``getGlobalCollector()`` method.

src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTrackerFactory.java
```diff
@@ -19,29 +19,91 @@
...
+      this.collector = globalCollector.register(registry);
+   }
+
+   /**
+    * If both a CollectorRegistry and a Collector are provided, use provided CollectorRegistry and Collector
+    */
+   public PrometheusMetricsTrackerFactory(CollectorRegistry registry, HikariCPCollector collector) {
+      this.registry = registry;
+      this.collector = collector.register(registry);
+   }
```
brettwooldridge: Why is this constructor necessary?  In particular, if ``HikariCPCollector`` is internal then this constructor is never used.

src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
PuGuanyu: If we use only one collector and provides custom CollectorRegistry, why not make users are able to use this HikariCPCollector as a default implemention? So that users can use this or write another implemention extends HikariCPCollector.

PuGuanyu: @brettwooldridge

PuGuanyu:
@brettwooldridge I understand we should keep it simple, stupid, but It doesn't matter that we provide customization abilities~

PuGuanyu:
@brettwooldridge Well, long time no see~

PR #989, opened 2017-09-25
Removed some maturity wrinkles :)

nitincchauhan:
1. isConnectionAlive() if connection is invalid/broken, avoid delay
caused by calls (and probably more exception) on connection to be
closed.
2. getConection() removed dead code. poolEntry is always null if
connectionBag.borrow() throws InterruptedException.
probably metricsTracker.recordBorrowTimeoutStats(startTime);
and      metricsTracker.recordConnectionTimeout() are same/ should be
one call?

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -152,38 +152,38 @@ abstract class PoolBase
    boolean isConnectionAlive(final Connection connection)
    {
       try {
-         try {
-            setNetworkTimeout(connection, validationTimeout);
+         setNetworkTimeout(connection, validationTimeout);
 
-            final long validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
+         final int validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
 
-            if (isUseJdbc4Validation) {
-               return connection.isValid((int) validationSeconds);
-            }
-
+         boolean valid = true;
+         if (isUseJdbc4Validation) {
+            valid = connection.isValid(validationSeconds);
+         }
+         else {
             try (Statement statement = connection.createStatement()) {
-               if (isNetworkTimeoutSupported != TRUE) {
-                  setQueryTimeout(statement, (int) validationSeconds);
+               if (isNetworkTimeoutSupported == FALSE) {
+                  setQueryTimeout(statement, validationSeconds);
                }
 
                statement.execute(config.getConnectionTestQuery());
-            }
-         }
-         finally {
-            setNetworkTimeout(connection, networkTimeout);
+             }
+          }
 
-            if (isIsolateInternalQueries && !isAutoCommit) {
-               connection.rollback();
-            }
-         }
+          if (valid) {
+             if (isIsolateInternalQueries && !isAutoCommit) {
+                connection.rollback();
+             }
 
-         return true;
+             setNetworkTimeout(connection, networkTimeout);
+             return true;
+          }
       }
       catch (Exception e) {
          lastConnectionFailure.set(e);
          LOGGER.warn("{} - Failed to validate connection {} ({})", poolName, connection, e.getMessage());
-         return false;
       }
+      return false;
    }
 
    Throwable getLastConnectionFailure()
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -181,11 +181,9 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
          } while (timeout > 0L);
 
          metricsTracker.recordBorrowTimeoutStats(startTime);
+         metricsTracker.recordConnectionTimeout();
       }
       catch (InterruptedException e) {
-         if (poolEntry != null) {
-            poolEntry.recycle(startTime);
-         }
          Thread.currentThread().interrupt();
          throw new SQLException(poolName + " - Interrupted during connection acquisition", e);
       }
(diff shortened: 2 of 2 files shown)


brettwooldridge:
``connectionBag.borrow()`` is not the only code that can throw ``InterruptedException``.

brettwooldridge:
``recordBorrowTimeoutStats(startTime)`` and ``recordConnectionTimeout()`` cannot be merged, because ``recordConnectionTimeout()`` is a public API (``IMetricsTracker.recordConnectionTimeout()``), while ``recordBorrowTimeoutStats()`` is an internal method that eventually calls ``IMetricsTracker.recordConnectionAcquiredNanos()``.

nitincchauhan:
If you comment out catch block there, only connectionBag.borrow() complains to catch InterruptedException

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -152,38 +152,38 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
...
-               connection.rollback();
-            }
-         }
+          if (valid) {
+             if (isIsolateInternalQueries && !isAutoCommit) {
+                connection.rollback();
+             }
 
-         return true;
+             setNetworkTimeout(connection, networkTimeout);
```
brettwooldridge: This absolutely must be in a ``finally``.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -152,38 +152,38 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
...
-
-            final long validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
+         setNetworkTimeout(connection, validationTimeout);
 
-            if (isUseJdbc4Validation) {
-               return connection.isValid((int) validationSeconds);
-            }
+         final int validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
 
+         boolean valid = true;
```
brettwooldridge: This is an additional local variable allocation...

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -152,38 +152,38 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
...
+         if (isUseJdbc4Validation) {
+            valid = connection.isValid(validationSeconds);
```
brettwooldridge: And assignment...

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -152,38 +152,38 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
...
-         finally {
-            setNetworkTimeout(connection, networkTimeout);
+             }
+          }
 
-            if (isIsolateInternalQueries && !isAutoCommit) {
-               connection.rollback();
-            }
-         }
+          if (valid) {
```
brettwooldridge: And extra conditional.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: not really. if connection is invalid or broken, it gets closed immediately on returning false

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: required for flow without finally

brettwooldridge:
@nitincchauhan You're right re:``InterruptedException``.  I was thinking it was an unchecked Exception (``RuntimeException`` subclass).

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: this is to avoid rollback and resetting networktimeout for dead / invalid connection

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
nitincchauhan: worth for check to avoid method calls to dead / invalid connections later before returning false

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
brettwooldridge: No.  If the timeout is not reset, the ``close()`` call itself may fail prematurely.

nitincchauhan: before closing this you should have waited for reply or drill down code in closeConnection()

brettwooldridge: Trust me, I did.  The timeout still needs to be reset in isConnectionAlive(), because the close request goes into a queue, and there is no way of knowing how many milliseconds (or seconds) it may sit there before being serviced.

nitincchauhan: ok. to close the dead / broken connection, if the queue takes longer to execute quietlyCloseConnection then connection would be closed anyway in previously set network timeout.

nitincchauhan: for dead / broken connection only call that probably succeeds is close(). all other methods would throw

nitincchauhan: the aim here is to close connection and as per doc, setNetworkTimeout will close connection on timeout. validationTimeout in this case.
attempting to close already broken / invalid connection results in more exceptions here.

brettwooldridge: Just because the query times out does not mean the connection is actually dead (disconnected). It just means that the database did not respond within the user’s configured timeout.  Of course, will still evict the connection.

This change is an unimportant optimization that carries risks.  The existing code has been running without reported problems for quite a long time, which is worth a lot.  

Additionally, the setNetworkTimeout() code paths in drivers like MySQL have already proven to be brittle, and several drivers had only just implemented support (primarily  because of HikariCP) and those code paths are largely untested.

Changing the code buys us very little as far as I can tell, and in fact I find it less readable and more difficult to reason about.

brettwooldridge:
@nitincchauhan HikariCP v2.7.2 containing the yummy parts of this pull request has been released.

PR #991, opened 2017-10-02
Split copyState to copyStateTo and copyStateFrom

### LINKED ISSUE #892: Document HikariConfig.copyState(HikariConfig other)
#### Environment
```
HikariCP version: 2.6.1
```

Please add Javadoc to `HikariConfig.copyState(HikariConfig other)`. I keep on mistakenly thinking the method means "copy the state of `other` (into `this`)" when in fact it does the opposite: it copies `this` into `other`.

Renaming the method to `copyStateTo` or `copyStateInto` would also help.

FredDeschenes:
Fixes #892 and should prevent further confusion.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -1043,7 +1043,26 @@ public class HikariConfig implements HikariConfigMXBean
       }
    }
 
+   /**
+    * Deprecated, use {@link #copyStateTo(HikariConfig)}.
+    * <p>
+    * Copies the state of {@code this} into {@code other}.
+    *</p>
+    *
+    * @param other Other {@link HikariConfig} to copy the state to.
+    */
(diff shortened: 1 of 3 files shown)


brettwooldridge:
@FredDeschenes Thanks for the pull request.  I *do* have one request; acknowledging that it's nitpicky.  Originally ``copyState()`` was only used internally, though accidently marked ``public``.  Unfortunately, it seems to have *leaked* into external usage.  Which is *ok*...

However, given that the only internal consumer uses ``copyStateTo()``, I'd prefer to kill ``copyStateFrom()``, because the user can just as easily reverse the order and use ``copyStateTo()`` instead, and I'd rather not carry around "convenience" methods that are not actually used by the pool.

Thanks.

FredDeschenes:
That should do it.

Sorry for the double commit btw, I always forget that Github doesn't "Markdown-ize" commit messages (although it does for the details...).

PR #1009, opened 2017-10-31
ProxyConnection always resets the underlying connection

BrentDouglas:
When closing transactions with uncommited statements the proxy can
return the actual connection in an invalid state back to the pool.
Moving the connection reset into a finally block ensures that it is
always returned in a valid state.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -236,18 +236,21 @@ public abstract class ProxyConnection implements Connection
          leakTask.cancel();
 
          try {
-            if (isCommitStateDirty && !isAutoCommit) {
-               delegate.rollback();
-               lastAccess = currentTime();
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@BrentDouglas I'm still unclear how this actually solves the issue.

Inside of ``resetConnectionState()`` there are potentially six individual driver calls, any one of which could throw an exception that prevents the rest from executing.  In which case the partially reset connection is returned to the pool.

To be clear, I am not actually sure there *is* an issue in the first place.  Under what circumstance can the ``rollback()`` fail that also does not result in a SQL State that causes an ejection from the pool?  Similarly, I would expect that any of the invocations in ``resetConnectionState()`` cannot fail in a manner that would avoid eviction.  I could be wrong, and if so I would like a concrete example.

BrentDouglas:
Thats a good point about the other calls. Honestly, I'm not sure what I was thinking when I opened this. I had been debugging an issue that looked like a request was reusing a connection that had been opened and returned uncommited in a different request though I can't see how I came to the conclusion that this would help.

PR #1016, opened 2017-11-10
Update for Micrometer.rc3

checketts:
I apologize @brettwooldridge I just found out the latest rc removed one more method. It has stabilized now so we don't expect any further breakages. Right @jkschneider?


jkschneider:
The pieces Hikari is using are stabilized anyway, yes. But I would suggest removing `percentilesHistogram` entirely from core libraries like this. It can be enabled later selectively by users via meter filters, and we shouldn't be sending extra time series to backends that can't make use of them.

Issue #928, opened 2017-07-04
HikariJNDIFactory should not throw a NamingException when !"javax.sql.DataSource".equals(ref.getClassName())


h-thurow:
The JavaDoc of ObjectFactory.getObjectInstance(Object, Name, Context, Hashtable<?,?>) : Object specifies this behaviour:

"An object factory should only throw an exception if it is sure that it is the only intended factory and that no other object factories should be tried. If this factory cannot create an object using the arguments supplied, it should return null."

So if one configured multiple ObjectFactories to be asked if they can create a specific type by setting  java.naming.factory.object to a list of object factories HikariJNDIFactory will prevent other factories to be consulted.

org.apache.commons.dbcp2.BasicDataSourceFactory does it this way:

        if (!"javax.sql.DataSource".equals(ref.getClassName())) {
            return null;
        }

brettwooldridge:
Can you please submit a pull request?

PR #1022, opened 2017-11-17
Issue #928: HikariJNDIFactory should not throw a NamingException when…

h-thurow:
Finally I found the time to propose this change to solve issue #928.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
+++ b/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
@@ -57,7 +57,9 @@ public class HikariJNDIFactory implements ObjectFactory
          }
          return createDataSource(properties, nameCtx);
       }
-      return null;
+      else {
+         return null;
+      }
    }
 
    private DataSource createDataSource(final Properties properties, final Context context) throws NamingException
--- a/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
+++ b/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
@@ -40,7 +40,8 @@ import com.zaxxer.hikari.util.PropertyElf;
 public class HikariJNDIFactory implements ObjectFactory
 {
    @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
```diff
@@ -40,31 +40,26 @@
 public class HikariJNDIFactory implements ObjectFactory
 {
    @Override
-   synchronized public Object getObjectInstance(Object obj, Name name, Context nameCtx, Hashtable<?, ?> environment) throws Exception
-   {
+   synchronized public Object getObjectInstance(Object obj, Name name, Context nameCtx, Hashtable<?, ?> environment) throws Exception {
```
brettwooldridge: HikariCP's code style is open brace on a new line after method declaration, and open brace on the same line inside in method.

src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
```diff
@@ -40,31 +40,26 @@
...
+            String type = element.getType();
+            if (type.startsWith("dataSource.") || hikariPropSet.contains(type)) {
+               properties.setProperty(type, element.getContent().toString());
+            }
          }
+         return createDataSource(properties, nameCtx);
+      }
+      else {
+         return null;
       }
```
brettwooldridge: Eliminate the ``else`` and simply return null after the ``if``.

src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
h-thurow: DONE

src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
h-thurow: DONE I am not a friend of multiple returns in a method because one have to read through all the method body to finally find out that some statements are never reached. The if-else block would have made it a little bit clearer.

PR #1026, opened 2017-11-21
Added a support for SQL server's specific isolation level (Snapshot)

znikola:
Added a constant that represents the SQL server's snapshot isolation level and added a case with the constant.

Configuring the isolation level with Spring Boot 2's application.yml works.


brettwooldridge:
@znikola Thanks.  This change will be included in the next release.

PR #1041, opened 2017-12-10
Log statement passing along wrong number of arguments

KoenDG:
Hello to all.

I was checking some of the source code and this error popped up in Intellij.

This log statements is passing along the wrong number of parameters.
When passing along a `Throwable` parameters, the calls from the LOGGER class only ever have 2 arguments: the already formatted String, and the Throwable. As can be seen in the source code of the LOGGER object.

I copied the permalinks from version 1.7.25:

The code that was actually being called: https://github.com/qos-ch/slf4j/blob/a81440ea36676f3b090f5a4ccba70d17c7b80f6e/slf4j-api/src/main/java/org/slf4j/Logger.java#L255

The code being called after this change: https://github.com/qos-ch/slf4j/blob/a81440ea36676f3b090f5a4ccba70d17c7b80f6e/slf4j-api/src/main/java/org/slf4j/Logger.java#L280

I used a String.format() to fix it, not sure if that's optimal.

Also, this one file had unused imports, removed them.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
@@ -17,11 +17,8 @@
 package com.zaxxer.hikari.metrics.prometheus;
 
 import com.zaxxer.hikari.metrics.IMetricsTracker;
-import io.prometheus.client.Collector;
-import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Summary;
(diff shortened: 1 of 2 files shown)


brettwooldridge:
I believe that your IDE is mistaken.  The point of slf4j methods that accept (String, Object, Object) or (String, Object...) is to *avoid* actually performing the format operation if DEBUG logging is disabled.  Your code unconditionally formats the message.

 Slf4j checks the type of the last parameter at runtime to determine if it is an instanceof Exception.  Maybe this confuses the IDE, but rest assured that slf4j is not confused.

KoenDG:
Didn't know that last part. Ah well, life goes on :P

PR #964, opened 2017-08-29
Allow data source to be reinitialized at runtime

tomreay:
As described here: https://github.com/brettwooldridge/HikariCP/issues/763

Allows a user to reinitialize the data source if they have, for instance, changes the user/password.

PR #1083, opened 2018-02-09
Fix init of HikariDataSource, if jdbcUrl was specified not though pro…

GrigorievNick:
…perties, but with setter. In case of DataSourceClassName specified

In case initialisation dataSource with jdbcUrl and DataSourceClassName:
PoolBase ignore data that was set to HicariConfig, not as properties, but as `setX` method.
So enrich properties with HicariConfig getter methods.


johnou:
Duplicate of https://github.com/brettwooldridge/HikariCP/pull/964?

PR #1091, opened 2018-02-19
Deprecate setMetricRegistry in favour of setMetricsTrackerFactory 

NersesAM:
Following up on my yesterday's PR #1089. 
Even though it is ok, I think the fact of accepting an Object and then doing tricks to find out wether the library class is in classpath and later on casting the object to correct type are not very nice. 
This is especially not great in light of [Change in dropwizard metrics package name in upcoming v5](https://github.com/dropwizard/metrics/pull/1255). For some time we will need to support both versions. This will add even more complexity in that part of the code. 

That's why I propose to deprecate/delete setMetricRegistry in favour of setMetricsTrackerFactory. Also by providing overloaded static factory methods in the MetricsTrackerFactory interface we won't need to do type checks at all. Users can just use `setMetricsTrackerFactory(from(metricRegistry))` instead of `setMetricRegistry(metricRegistry)` allowing to greatly simplify code around registering metricRegsitry .
This is preliminary PR just to get your feedback. Can update it with complete removal.
WDYT?

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/MetricsTrackerFactory.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/MetricsTrackerFactory.java
@@ -16,6 +16,11 @@
 
 package com.zaxxer.hikari.metrics;
 
+import com.codahale.metrics.MetricRegistry;
+import com.zaxxer.hikari.metrics.dropwizard.CodahaleMetricsTrackerFactory;
+import com.zaxxer.hikari.metrics.micrometer.MicrometerMetricsTrackerFactory;
+import io.micrometer.core.instrument.MeterRegistry;
+
 public interface MetricsTrackerFactory
 {
    /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/zaxxer/hikari/metrics/MetricsTrackerFactory.java
```diff
@@ -26,4 +31,13 @@
     * @return a IMetricsTracker implementation instance
     */
    IMetricsTracker create(String poolName, PoolStats poolStats);
+
+   static MetricsTrackerFactory from(MetricRegistry metricRegistry) {
```
johnou: I don't think MetricsTrackerFactory should contain references to implementation specific code.

NersesAM: I didn't want to create MetricsTrackerFactoryFactory :)
This seems not right at first but in JDK itself starting from 9 there are lots of examples like this. For example [List.of(...)](https://github.com/netroby/jdk9-dev/blob/master/jdk/src/java.base/share/classes/java/util/List.java#L856-L860)

Do you want me to revert this file so that users have to do `setMetricsTrackerFactory(new CodahaleMetricsTrackerFactory(metricRegistry))` instead?

NersesAM: There is no risk of CNFE here if that's your concern. This method cannot be invoked without creating MetricRegistry first. Which means the class will already be loaded by JVM before reaching this point. 
When deprecating corresponding metricsTracker implementation this factory method can be deprecated at the same time.

johnou: More concerned about it being too tightly coupled.

PR #1100, opened 2018-02-23
Remove hard-coded percentiles for Micrometer metrics

shakuzen:
See following comment from Micrometer maintainer https://github.com/micrometer-metrics/micrometer/issues/91#issuecomment-365780694. The hard-coded client-side percentile in HikariCP will make an extra time series whether the user wants it or not. Ultimately, to configure the percentiles or histogram they want, they will have to use a `MeterFilter`. I've added documentation for how they can do that.

If you don't want the public constant or documentation changes, feel free to take just the commit that removes hard-coded percentiles.

/cc @jkschneider @checketts

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/micrometer/MicrometerMetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/micrometer/MicrometerMetricsTracker.java
@@ -9,18 +9,42 @@ import io.micrometer.core.instrument.Timer;
 
 import java.util.concurrent.TimeUnit;
 
+/**
+ * {@link IMetricsTracker Metrics tracker} for Micrometer.
+ * HikariCP metrics can be configured in your application by applying a
+ * {@link io.micrometer.core.instrument.config.MeterFilter MeterFilter} to metrics starting with
+ * {@link #HIKARI_METRIC_NAME_PREFIX}. For example, to configure client-side calculated percentiles:
+ *
+ * <blockquote><pre>
+ *     new MeterFilter() {
+ *       &#064;Override
+ *       public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
+ *         if (id.getName().startsWith(MicrometerMetricsTracker.HIKARI_METRIC_NAME_PREFIX)) {
+ *           return DistributionStatisticConfig.builder()
+ *               .percentiles(0.5, 0.95)
+ *               .build()
+ *               .merge(config);
+ *            }
+ *         return config;
+ *         }
+ *      };
+ * </pre></blockquote>
+ */
 public class MicrometerMetricsTracker implements IMetricsTracker
 {
-   private static final String METRIC_CATEGORY = "pool";
-   private static final String METRIC_NAME_WAIT = "hikaricp.connections.acquire";
-   private static final String METRIC_NAME_USAGE = "hikaricp.connections.usage";
-   private static final String METRIC_NAME_CONNECT = "hikaricp.connections.creation";
+   /** Prefix used for all HikariCP metric names. */
+   public static final String HIKARI_METRIC_NAME_PREFIX = "hikaricp";
 
-   private static final String METRIC_NAME_TIMEOUT_RATE = "hikaricp.connections.timeout";
-   private static final String METRIC_NAME_TOTAL_CONNECTIONS = "hikaricp.connections";
-   private static final String METRIC_NAME_IDLE_CONNECTIONS = "hikaricp.connections.idle";
-   private static final String METRIC_NAME_ACTIVE_CONNECTIONS = "hikaricp.connections.active";
-   private static final String METRIC_NAME_PENDING_CONNECTIONS = "hikaricp.connections.pending";
+   private static final String METRIC_CATEGORY = "pool";
+   private static final String METRIC_NAME_WAIT = HIKARI_METRIC_NAME_PREFIX + ".connections.acquire";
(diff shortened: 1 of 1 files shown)


shakuzen:
@brettwooldridge any feedback about this pull request?

brettwooldridge:
Does this change break existing users?

jkschneider:
If anybody relies on the percentile shipped by Hikari, they can still configure it with `management.mertrics.distribution.percentiles.hikari=0.95` (property-driven meter [filters](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#_per_meter_properties)).

brettwooldridge:
I assume all existing users rely on those defaults. In that case, HikariCP need to default to existing values, and allow them to be overridden by the property.

shakuzen:
I can certainly understand the desire to not break existing users and it is possible there may be some users relying on the hard-coded percentile. It would have been best if the Micrometer support in HikariCP could have been somehow marked as incubating until Micrometer 1.0 GA was released and the support here was finalized. After all, it wasn't based on the GA version of Micrometer until recently (and after the last HikariCP release). As they say, hindsight is 20/20.
Also, not that two wrongs make a right, but https://github.com/brettwooldridge/HikariCP/commit/4e0fcb11d1c2244a3efd809d5e037a2e7ac8d9eb looks like it dropped the median quantile without user complaint.

I guess removal of the hard-coded percentile can be scheduled for HikariCP 3.0, where it would acceptable to have a breaking change.

@jkschneider to disable the current percentile, I tried the following property filter in a Spring Boot app but it doesn't seem to do the trick.

```yml
management:
  metrics:
    distribution:
      percentiles:
        hikaricp: []
```

The equivalent Java configuration does seem to work, though:

```java
@Bean
MeterFilter disableHikariQuantileMeterFilter() {
	return new MeterFilter() {
		@Override
		public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticConfig config) {
			if (id.getName().startsWith("hikaricp")) {
				return DistributionStatisticConfig.builder()
						.percentiles()
						.build();
			}
			return config;
		}
	};
}
```

billoneil:
I believe @brettwooldridge meant you can add a HikariCP property / flag to disable the metrics you don't want. They should stay defaulted to on though until possibly a major version bump.

shakuzen:
> I believe @brettwooldridge meant you can add a HikariCP property / flag to disable the metrics you don't want.

@billoneil That could work, but the point of this pull request is to move all such configuration out of HikariCP, which better aligns with Micrometer's configuration model. User's can consistently and cross-cuttingly configure any and all of their Micrometer `Meter`s with `MeterFilter`s. Putting something in HikariCP to work around this pesky hard-coded percentile seems to subvert that and make users have to configure things in two ways. It may, however, be unavoidable until a major version bump, as you say.

brettwooldridge:
@shakuzen @billoneil The current commitment for Micrometer is  "Alpha-level support", as stated in the change log for v2.7.0.

I am willing and ready to cut v3.0.0 as the next release.  So, if we're going to make a breaking change, now would be the time to do it.

Having said that, HikariCP strives for an "out-of-the-box" experience that is as simple as possible, with defaults that make sense for majority of users.  Maybe this is simply not possible with Micrometers, I don't know.

What I *would* like you to do is simply ponder and consider whether this is the case, or whether there is something clever that can be done, for example detecting whether a MeterFilter is associated and if not injecting one with reasonable defaults.  I honestly have no idea, I'm just spitballing here.  Just come back to me with what you consider the absolute best solution (even if it is what is currently in this pull request)...

brettwooldridge:
@shakuzen I am prepared to merge this change.  Is there any documentation required to guide existing micrometers users in how to configure micrometers correctly in HikariCP?

jkschneider:
@brettwooldridge Missed this earlier comment... Here is [the doc](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#_per_meter_properties) on how to add percentiles back to Hikari timers with a property in Spring Boot.

shakuzen:
@brettwooldridge Sorry for my delay in responding.
In addition to the Spring Boot-specific documentation @jkschneider shared, there is the generic [Micrometer documentation on configuring this](http://micrometer.io/docs/concepts#_configuring_distribution_statistics).

If we want to give a more opinionated, out-of-the-box default experience, we would have to add logic to determine the specific metrics backend (based on the type of `MeterRegistry`), and then we could enable percentile histogram if the backend supports it or configure default client-side percentiles otherwise. If the `MeterRegistry` we're passed is a `CompositeMeterRegistry` we would need to apply the logic to each of the underlying `MeterRegistry`.

As for backing off if a user registers a `MeterFilter` that configures histograms or hard-coded percentiles to Hikari meters (or wants to turn these both off), I think we have an ordering issue. Filters are applied in the order they're registered. HikariCP being a library can't control the order in which its code is run relative to others in the application. If Hikari's filter were registered before a user filter, their filter would have to undo what Hikari's did, if it isn't what they wanted. If Hikari's filter were registered after a user filter, we could check for histograms/percentiles being configured and back off, but how would a user configure no histograms and no percentiles since this is Micrometer's default configuration? I'm not sure there is a way to detect another meter filter is going to configure histograms/percentiles for Hikari meters and back off.

@jkschneider any thoughts?

jkschneider:
I think in general folks should just look at max for alerts and max+average together for some notion of what's normative. Even high percentile values fail to convey meaningful information about end user experience, so I'd say looking at them is the exception rather than the rule.

The famous slide from one of Gil Tene's that demonstrates this:

image

PR #1108, opened 2018-03-08
Maintain strong reference to PoolStats for Micrometer gauges

shakuzen:
Micrometer's gauges were breaking after the `PoolStats` object they weakly referenced was garbage collected. This change holds a strong reference to the `PoolStats` object in `MicrometerMetricsTracker` to avoid this issue.

See http://micrometer.io/docs/concepts#_why_is_my_gauge_reporting_nan_or_disappearing

I noticed that the HikariCP gauges were showing `NaN` for their value sometime after application startup. The recently added documentation in Micrometer linked above helped me track down the issue.


brettwooldridge:
I’ll try to put out a release today.

Issue #1128, opened 2018-04-06
IdleTimeout is unreliable


javrasya:
#### Environment
```
HikariCP version: 2.7.2
JDK version     : 1.8.0_152
Database        : PostgreSQL with ELB
Driver version  : 42.2.2
```

-----------------------------------------------------------------------------------------
#### Description

We spotted some reliability issues with `idleTimeout`. In some cases, the idle connections are not evicted right after they breached the `idleTimout`. In our case, we have idle session timeout is enabled on ELB which is running ahead of PostgreSQL. ELB is dropping the connection after the idle timeout. Yet the connection pool on client side thinks the connection can still be used. 

We have checked the latest version of `HikariCP` but the issue still remains. **We are preparing a pull request for this as well.**

Here is a stack trace I can share after this happens;

```
org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.
    at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:333)
    at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
    at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
    at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:155)
    at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:132)
    at org.postgresql.jdbc.PgConnection.isValid(PgConnection.java:1364)
    at com.zaxxer.hikari.pool.PoolBase.isConnectionAlive(PoolBase.java:160)
    at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:173)
    at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:149)
    at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:85)
    at org.skife.jdbi.v2.DataSourceConnectionFactory.openConnection(DataSourceConnectionFactory.java:34)
    at org.skife.jdbi.v2.DBI.open(DBI.java:211)
```

brettwooldridge:
I'm not sure what you mean by "unreliable".  But the documentation clearly states:

> Whether a connection is retired as idle or not is subject to a maximum variation of +30 seconds, and average variation of +15 seconds.

javrasya:
[...]

PR #1129, opened 2018-04-06
Make idleTimeout reliable

javrasya:
Regarding the issue #1128 , IdleTimeout doesn't work as expected when there is no connection needs to be removed. This toRemoved logic is preventing the idle connections have been there for more than idleTimeout to be refreshed. This unreliability becomes a problem when there is also an idleSessionTimeout on the server side. The server is dropping the idle connections whereas hikari is not. Thus   the application picks a connection from the pool which is actually dropped by the server. The idle connection should be evicted, closed and replaced with the newly created one by housekeeping immediately. To avoid this problem, this buggy toRemoved logic is removed.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -772,14 +772,11 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
                logPoolState("Before cleanup ");
                afterPrefix = "After cleanup  ";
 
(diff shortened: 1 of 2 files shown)


javrasya:
Thank you

brettwooldridge:
@javrasya You're welcome, Ahmet.  Trust me, even with a ``maxLifetime`` as short as one or two minutes, you will find HikariCP both performant and extremely stable.  Connection replacement occurs in the background, and your application will never feel the effects.

javrasya:
Then we will use that one. Thank you @brettwooldridge

PR #1136, opened 2018-04-16
log loaded driver when using jdbcUrl (Closes #1018)

### LINKED ISSUE #1018: Print driver when Hikari chooses it
#### Environment
```
HikariCP version: 2.7.2
JDK version     : 1.8.0_131-b11
Database        :  Percona
Driver version  : mysql-connector-java-8.0.8-dmr-bin.jar
```
I am using the jdbcURL without driverClassName. We have run into some misleading errors making us think it was not using the correct driver, however it was, we just couldn't see it.  I have tried many ways to print the driver that Hikari is using when it determines it itself with no luck. How can I get and print the driver Hikari chooses to be able to confirm it is using what I think it is using?

goekay:
This PR closes issue #1018.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
@@ -101,6 +101,7 @@ public final class DriverDataSource implements DataSource
       try {
          if (driver == null) {
             driver = DriverManager.getDriver(jdbcUrl);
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@johnou Good catch.  That's what I get for merging at 4AM.

goekay:
my bad. i could just leave the value of jdbcUrl out actually.

PR #1159, opened 2018-05-18
Log uncaught error while acquiring connection

HeartSaVioR:
Default ThreadPoolExecutor implementation just lets thread being killed without any information.

This patch introduces new class "ListenableThreadPoolExecutor" which can set listener to handle
pre/post execution which ThreadPoolExecutor basically provides the lifecycle of running task.

This patch adds listener to connection adder/closer to log uncaught error before thread is killed.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -482,9 +482,22 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
          return poolEntry;
       }
-      catch (Exception e) {
+      catch (SQLException e) {
          if (poolState == POOL_NORMAL) { // we check POOL_NORMAL to avoid a flood of messages if shutdown() is running concurrently
-            LOGGER.debug("{} - Cannot acquire connection from data source", poolName, (e instanceof ConnectionSetupException ? e.getCause() : e));
+            LOGGER.debug("{} - Cannot acquire connection from data source", poolName, e);
+         }
+         return null;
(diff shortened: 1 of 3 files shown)


HeartSaVioR:
@brettwooldridge
Please take a look. Thanks in advance!

Btw, if the patch is not proposing preferred way, we also have some other alternatives:

1. Catch `Throwable` in `PoolEntryCreator.call` and log, and only rethrow `Exception`.
2. Catch `Throwable` in `HikariPool.createPoolEntry`. We would want to clone method for supporting `checkFailFast()`, cause we may want to propagate Error to upstream immediately in that case.

brettwooldridge:
@HeartSaVioR I would prefer the second approach.  In the case of ``HikariPool.createPoolEntry``, I would prefer two catches, first ``SQLException``, then ``Throwable``.  The ``Throwable`` case should log and then wrap the exception in a ``ConnectionSetupException`` exception.  [This check](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L554) in ``checkFailFast()`` should handle the immediate propagation upstream.

HeartSaVioR:
@brettwooldridge 
Thanks for reviewing! I applied your suggestion. Please take a look again.

HeartSaVioR:
@brettwooldridge Kindly reminder.

HeartSaVioR:
@brettwooldridge Ping. Could you take a look if you really don't mind? I'd like to see it as whether softly rejected or just forgotten one for you.

brettwooldridge:
Merged.

PR #1182, opened 2018-06-18
allow to pass in a prometheus CollectorRegistry (fixes #1181)

### LINKED ISSUE #1181: Prometheus metrics are always registered to default registry
This is not bug, but a missing feature :)

I want to expose Hikari Metrics to Prometheus in a Spring Boot application. `PrometheusMetricsTrackerFactory` always registers the metrics do the default collector registry (`CollectorRegistry.defaultRegistry`). 

The static default registry is a pain for testing and modularity, though, so in our Spring Boot application we created our own `CollectorRegistry` instance that we use to register all our metrics. 

It would be nice if `PrometheusMetricsTrackerFactory`  provided a way to pass in a `CollectorRegistry` instance to have more control.

thombergs:
Added a constructor to `PrometheusMetricsTrackerFactory` that allows to pass in a `CollectorRegistry` instance.

Removed the `static` modifier from the collector field since with the new constructor you may theoretically have more than one instance of the factory using a different `CollectorRegistry` instance each.

fixes #1181

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/PrometheusMetricsTracker.java
@@ -17,6 +17,7 @@
 package com.zaxxer.hikari.metrics.prometheus;
 
 import com.zaxxer.hikari.metrics.IMetricsTracker;
+import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.Counter;
 import io.prometheus.client.Summary;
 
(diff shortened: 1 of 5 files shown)


thombergs:
I just noticed that PR #985 already addresses this issue, however this solves it differently.

brettwooldridge:
@thombergs Have you looked at #985 ?  What are the pros/cons of these changes by comparison?

thombergs:
@brettwooldridge I just had a little deeper look at #985. 

#985 introduced a static "global" list of `HikariCPCollector` instances which is accessed by static methods.

I think this solution here is a little easier on maintenance since it removes any uses of static `HikariCPCollector` instances and thus removes the need to centrally manage a list of instances.

brettwooldridge:
@thombergs Ok, fair enough.  Next question.  Is this in any way a breaking change for existing users?  Or is it backwards compatible?

thombergs:
@brettwooldridge this is backwards compatible. 

`PrometheusMetricsTrackerFactory` is the only public class that was touched, I simply added another constructor.

You can still use the default constructor:

```java
config.setMetricsTrackerFactory(new PrometheusMetricsTrackerFactory());
```

and it will have the same effect as it had before (the metrics will be registered to prometheus' `defaultRegistry`). This is also asserted in the unit test `PrometheusMetricsTrackerFactoryTest`.

You can opt-in to use the following instead, to register the metrics to a custom registry:

```java
CollectorRegistry myRegistry = ...;
config.setMetricsTrackerFactory(new PrometheusMetricsTrackerFactory(myRegistry));
```

Issue #899, opened 2017-05-25
Thoughts on Integration with RDS IAM authentication


jklukas:
Has there been any discussion of how HikariCP would interact with [IAM database authentication](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html) for RDS?

Essentially, that scheme requires applications to fetch a limited lifetime token and pass that in as the `password` when connecting to a database instance.

To use that scheme with HikariCP, there would need to be a way to fetch the token every time a new connection needs to be added to the pool.

I assume there's no way to achieve this in HikariCP as it exists today, but let me know if I'm wrong.

If it's true that this is not currently possible, would a reasonable solution be to provide an interface that users could implement for dynamically fetching a password, and then some new configuration parameter to pass in a class name instead of a static password?

Looking for thoughts on other approaches, too.

DonMcNamara:
This issue seems related: https://github.com/brettwooldridge/HikariCP/issues/763

marcioos:
I created a PR that might help with that: #1196

bjhartin:
Could this be merged?  We have the same need, AWS Aurora and RDS IAM tokens.

thiagolocatelli:
Any idea when is this going to me merged? Also looking forward to using this.

beccagaspard:
I was able to get RDS IAM auth working with Hikari (for MySQL at least) with the new-ish [AWS IAM Credential Plugin](https://mariadb.com/kb/en/mariadb-connector-j-250-release-notes) provided in the MariaDB Connector/J driver, no extra code required! 🤓    

Connection string should look something like this:
`jdbc:mariadb://host/db?credentialType=AWS-IAM&useSsl&serverSslCert=/somepath/rds-combined-ca-bundle.pem`

I also set `maxLifetime=600000` to match the 10 minute caching the driver uses.  

Most examples I found online seem to use custom coded implementations, so figured I would share in case this helps someone else trying to implement something similar.

brettwooldridge:
HikariCP v7.0.0 will include support for a ``HikariCredentialsProvider`` interface allowing for a user-supplied implementation.

Issue #763, opened 2016-11-17
Change driver properties at runtime


riba2101:
Hi there,

we are currently working on some security enhancements and noticed a shortcoming in HikariCP and were kinda wondered if this is by design?

So, the issue: you can't change (driver)properties once a pool is created? 

From what we see, there is only one option: close/shutdown the old one and create a new one in his place.
The problem here would be Hibernate and others who would wrap the DataSource

Some options are to monitor the HikariConfig changes, wrap Hikari in another DataSource(questionable) or add a ConnectionManager.

If we have overseen something, please let us know.

brettwooldridge:
We've just never had this request.  I don't know of any technical barrier to implementing this.

brettwooldridge:
@riba2101 Pull requests welcome.  You would need to create a new DataSource instance, initialize it, atomically swap it in; I suggest changing [the DataSource](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L77) member to an ``AtomicReference`` to facilitate that.

johntdyer:
Would this allow a user to change MySQL password at runtime ?   We're looking for a way to use vault for leased db credentials and don't want to have to restart service to enact updated credentials

riba2101:
This is the exact same problem we had. What we ended up doing was create a datasource that knows how to communicate with vault and supplied it to Hikari.

However I will supply a PR that would ease up vault integration

cowwoc:
I just ran across this. Ideally, changing the configuration should affect subsequent calls to `getConnection()`. Baring that, changing the configuration after the pool is already created should throw an exception. Currently, the code fails (to behave as expected) silently.

cksdnjs4113:
I have fixed this problem by replacing the property value. The getConnection(String username, String password) is kinda dummy method in hikari since it will be eventually called getConnection() method. This means that even though we change username/password in hikariconfig it will still call the property file that was created at initializeDataSource() method. 

This property is private, however, it doesn't mean that we cannot change the value at runtime. By using java reflection, I was able to change the property value, which allowed us to have dynamic passwords for the db. This is the code I used:

//Get HikariDataSource from the bean
HikariDataSource hikariDataSource = (HikariDataSource) apiContext.getBean("opsapids");


HikariPool hikariPool = (HikariPool) FieldUtils.readField(hikariDataSource, "fastPathPool", Boolean.TRUE);

[...]

PR #1196, opened 2018-07-04
Make HikariConfig able to retrieve passwords dynamically from a password supplier

marcioos:
This patch makes `HikariConfig` able to retrieve passwords from a provided password supplier at runtime when `getPassword` is called. This is useful for scenarios where expiring database passwords are dynamically generated at runtime (e.g. RDS access tokens) and one wants to avoid having to create a new `DataSource` or scheduling a task that periodically updates the password via `HikariConfigMXBean`.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -39,6 +39,7 @@ import java.util.TreeSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
 
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
@@ -97,6 +98,7 @@ public class HikariConfig implements HikariConfigMXBean
    private Object metricRegistry;
    private Object healthCheckRegistry;
    private Properties healthCheckProperties;
+   private Supplier<String> passwordSupplier;
 
    private volatile boolean sealed;
 
@@ -269,12 +271,13 @@ public class HikariConfig implements HikariConfigMXBean
    }
 
    /**
-    * Get the default password to use for DataSource.getConnection(username, password) calls.
+    * Get the password to use for DataSource.getConnection(username, password) calls.
+    * If {@link HikariConfig#passwordSupplier} is present, the password will be retrieved from it.
     * @return the password
     */
    public String getPassword()
    {
-      return password;
+      return passwordSupplier != null ? passwordSupplier.get() : password;
    }
 
    /**
@@ -284,7 +287,22 @@ public class HikariConfig implements HikariConfigMXBean
    @Override
    public void setPassword(String password)
    {
+      if (passwordSupplier != null) {
+         throw new IllegalStateException("cannot use setPasswordSupplier() and setPassword() together");
+      }
       this.password = password;
+   }
+
+   /**
+    * Set a password {@link Supplier} to be used by {@link HikariConfig#getPassword()} for dynamically acquiring a password.
+    * @param passwordSupplier the password {@link Supplier}
+    */
+   public void setPasswordSupplier(Supplier<String> passwordSupplier)
+   {
+      if (password != null) {
+         throw new IllegalStateException("cannot use setPassword() and setPasswordSupplier() together");
+      }
+      this.passwordSupplier = passwordSupplier;
    }
 
    /**
--- a/src/test/java/com/zaxxer/hikari/HikariConfigTest.java
+++ b/src/test/java/com/zaxxer/hikari/HikariConfigTest.java
@@ -1,0 +1,51 @@
+package com.zaxxer.hikari;
+
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+public class HikariConfigTest {
+
+   @Test
+   public void getPasswordReturnsPasswordFromSupplier() {
+      HikariConfig config = new HikariConfig();
+      config.setPasswordSupplier(() -> "supplied-password");
+
+      String password = config.getPassword();
+
+      assertThat(password, is(equalTo("supplied-password")));
+   }
+
+   @Test
+   public void getPasswordReturnsDefaultPassword() {
+      HikariConfig config = new HikariConfig();
+      config.setPassword("default-password");
+
+      String password = config.getPassword();
+
+      assertThat(password, is(equalTo("default-password")));
+   }
+
+   @Test(expected = IllegalStateException.class)
+   public void setPasswordThrowsExceptionWhenSupplierWasProvided() {
+      HikariConfig config = new HikariConfig();
+      config.setPasswordSupplier(() -> "supplied-password");
+
+      config.setPassword("default-password");
+
+      fail("Should have thrown exception");
+   }
+
+   @Test(expected = IllegalStateException.class)
+   public void setPasswordSupplierThrowsExceptionWhenDefaultPasswordWasProvided() {
+      HikariConfig config = new HikariConfig();
+      config.setPassword("default-password");
+
+      config.setPasswordSupplier(() -> "supplied-password");
+
+      fail("Should have thrown exception");
+   }
+}


src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -284,9 +287,24 @@ public String getPassword()
    @Override
    public void setPassword(String password)
    {
+      if (passwordSupplier != null) {
```
billoneil: If accepted maybe it would make sense to always use a supplier and just do
```java
this.passwordSupplier = () -> password;
```

brettwooldridge: It seems a nice idea at first, but I think replacing the ``Supplier`` might be unexpected.  I think I prefer the explicit failure so the developer immediately understands they have an ambiguous configuration.

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -284,9 +287,24 @@ public String getPassword()
...
+         throw new IllegalStateException("cannot use setPasswordSupplier() and setPassword() together");
+      }
       this.password = password;
    }
 
+   /**
+    * Set a password {@link Supplier} to be used by {@link HikariConfig#getPassword()} for dynamically acquiring a password.
+    * @param passwordSupplier the password {@link Supplier}
+    */
+   public void setPasswordSupplier(Supplier<String> passwordSupplier)
```
brettwooldridge, line 300: There should be a "getter" as well.

flosell:
Found this PR while trying to integrate AWS Secrets Manager with auto-rotating credentials. It would help a lot. 

In the meantime, for those looking for a workaround, configuring Hikari with a custom `DataSource` that handles dynamic credentials seems to work.

brettwooldridge:
@marcioos With the addition of a "getter" for the password supplier, I am prepared to merge this pull request.

ericstewart:
I like where this is going because I too am using Hikari and need to support AWS SecretsManager based password rotation. Be aware, however, that [AWS describes several scenarios for rotation](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-customizing.html). If changing only the password (which the changes above seem to support) then a small amount of downtime as the password change is applied is possible. 

Other options include rotating between two sets of credentials so that the username and password are changing together, so the username would need to change as well. However, having two getters that could fetch and external data opens up a race condition, so perhaps there is an option to obtain a pair?

src/main/java/com/zaxxer/hikari/HikariConfig.java
johnou: For what use-case?

2ndalpha:
Any plans to proceed with this PR?

fsauer65:
we need this too for use with RDS IAM tokens

brettwooldridge:
@2ndalpha @fsauer65 @ericstewart @flosell I have requested minor changes to this pull request, but the original submitter has not followed up.  If one of you would like to clone his repo, address the issues, and submit a new pull request, after review I will be willing to merge it.

c-a-m:
would you be open to a username supplier as well?

c-a-m:
> However, having two getters that could fetch and external data opens up a race condition, so perhaps there is an option to obtain a pair?

Would a supplier of Map.Enty<String, String> be preferred over a supplier of a custom pojo UsernamePassword?

2ndalpha:
@brettwooldridge the more I think about it the more I see that this PR is unnecessary.
Initially I was having trouble understanding how I can change the password at the runtime.
Correct solution:
```java
dataSource.getHikariConfigMXBean().setPassword("newPassword");
```

Using supplier makes sense if you want to roll passwords lazily.
That's not a good idea with RDS IAM.
It's expensive to fetch the new password, especially in ECS (there is a bug, need to assume a role before fetching the password: https://forums.aws.amazon.com/thread.jspa?threadID=291106)
it's better to fetch new password every 14 minutes and pass it to Hikari.

brettwooldridge:
@2ndalpha In the past we had an [open issue](https://github.com/brettwooldridge/HikariCP/issues/763#issuecomment-344402945) related to rolling the RDS password, and my recommended solution then was exactly as you propose; use the MBean.

So, I guess the question for everybody else is: if that solution is insufficient, why?

Anyone who knows the history of HikariCP, knows that I am *extremely* reluctant to add additional axis to the configuration.  Keeping the "surface area" of the configuration API as simple as possible is a *major* touchstone of this project.  Hence the quote at the top of the project page:

&nbsp;&nbsp;&nbsp;**"Simplicity is prerequisite for reliability."**
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;- *Edsger Dijkstra*

ericstewart:
I don’t know if anyone has tried this (mentioned in November) but it could be a solution for those using RDS without requiring modifications here. Has anyone attempted?

https://aws.amazon.com/blogs/security/use-aws-secrets-manager-client-side-caching-libraries-to-improve-the-availability-and-latency-of-using-your-secrets/

fsauer65:
I can live with using the MBean. Since I use it with RDS in an Akka application I simply used an Akka actor to schedule a refresh every 14 minutes. Works like a charm

fsauer65:
I was wrong about this working like a charm. I do refresh the tokens every 10 minutes, yet I do occasionally see auth exceptions, very strange... Here's what I'm doing:

```
 val system = ActorSystem("RDS")

  case object RefreshToken

  class TokenRefreshActor(ds: HikariDataSource, url: String, user: String) extends Actor {
    import concurrent.duration._
    import context.dispatcher

    override def preStart(): Unit = {
      super.preStart()
      context.system.scheduler.schedule(10 minutes, 10 minutes, self, RefreshToken)
    }

    def receive : Receive = {
      case RefreshToken =>
        val cfg = ds.getHikariConfigMXBean
        cfg.setPassword(getRDSCredentials(url,user))
        log.info(s"refreshed iam token for ${cfg.getPoolName}")
    }
  }

  val GetHostNameFromUrl = "jdbc:postgresql://([^,/:]*).*".r
  def getRDSCredentials(endpoint: String, userName: String, port: Int = 5432, regionName: String = "us-east-1"): String = {

    val generator: RdsIamAuthTokenGenerator = RdsIamAuthTokenGenerator.builder()
      .credentials(new DefaultAWSCredentialsProviderChain())
      .region(regionName)
      .build()

    val hostname = endpoint match {
      case GetHostNameFromUrl(hostName) => hostName
      case _ => sys.error(s"No parseable hostname in $endpoint")
    }

    generator.getAuthToken(GetIamAuthTokenRequest.builder.hostname(hostname).port(port).userName(userName).build)
  }

  def connectionPool(url: String, size: Int, name: String ): HikariDataSource = {
    val isIAM = rdsConfig.getBoolean("iamAuth")
    val user  = rdsConfig.getString("user")
    val pool = new HikariDataSource()
    pool.setPoolName(name)
    pool.setMaximumPoolSize(size)
    pool.setMinimumIdle(size)
    pool.setJdbcUrl(url)
    pool.setUsername(user)
    pool.setPassword(if (isIAM) getRDSCredentials(url,user) else rdsConfig.getString("password"))
    if (isIAM) system.actorOf(Props(classOf[TokenRefreshActor], pool, url, user))
    pool
  }
```
I see the log messages about the token being refreshed, so that actor is definitely firing every 10 minutes... I have no idea what I'm doing wrong here. Any ideas?

2ndalpha:
Documentation for `hikariConfigMXBean.setPassword` states
"Altering this at runtime only works for DataSource-based connections, not Driver-class or JDBC URL-based connections."

I'm using following Kotlin code to create the datasource for RDS IAM use:
```kotlin
val dataSource = HikariDataSource()
dataSource.dataSourceClassName = "org.postgresql.ds.PGSimpleDataSource"
dataSource.username = username
dataSource.password = generateAuthToken()
dataSource.addDataSourceProperty("url", "jdbc:postgresql://$host/$database")
dataSource.addDataSourceProperty("ssl", "true")
```

fsauer65:
Thank you!!! I completely missed that...

dannyv-dev:
Hi,

I'm currently working on a project that uses AWS Aurora (MySQL flavor) and uses AWS Secrets Manager service to rotate passwords every so often. My current implementation to integrate both services with HikariCP is to have a wrapper around the HikariDataSource, and on getConnection() check for SQLExceptions with a 1045 error code. This error code represents a DBAcesssDenied in MySQL. When this exception occurs, I will fetch the new username/password from AWS, and then through the HikariConfigMXBean change those properties at runtime.

That said, I've manage to get this working, however, I'm not completely satisfied. The main reason is because when I try to acquire a new connection with expired/wrong credentials the HikariDataSource#getConnection method will keep retrying with those bad credentials until a ConnectionTimeout occurs.

I'm fairly new to HikariCP, and from my current understanding, it seems like there is no way to avoid this loop right ? Do any of you guys have any suggestions apart from simply reducing the connection timeout ? 

Let me know your thoughts,

Thanks!

Jan-M:
Great to see this was already opened, what is missing to get this merged and made available?

josephnoall:
This effort would be really helpful for my team, but it doesn't quite do enough to settle the issue of dynamic credentials completely.  We are currently wrapping the underlying DataSource in order to intercept the getConnection() invocations, and it would be great for HikariCP to support dynamic credentials natively.  This PR doesn't quite address our problem, however, because we need both the username and the password to be dynamically resolved (we alternate between two application users so that we can roll them more seamlessly).  It seems like @brettwooldridge is reluctant to add more surface area to the Hikari API, so to me it would be short-sighted to support a password supplier without addressing usernames. I would like to see some kind of CredentialSupplier (so that both the password and username can be provided together, guaranteeing that they are not out of sync).  I'm curious whether anyone else also has this problem?

marcioos:
Given the considerations from @brettwooldridge and my lack of time to work on the suggested improvements, I'm thinking about just closing this pull request. Also, I'm not sure if this patch has everything needed to make the feature work correctly. I haven't checked if it is necessary to recreate the datasource or keep refreshing the ConfigMXBean.

martabelfm:
Any plans to merge it into the master?

beccagaspard:
I was working on getting RDS IAM working with Hikari and came across this thread.  I was able to get it working (for MySQL at least) with no code changes using the [AWS IAM Credential Plugin](https://mariadb.com/kb/en/mariadb-connector-j-250-release-notes/) provided in the MariaDB Connector/J driver which handles generating, caching and refreshing the auth token - I put a little more detail [here](https://github.com/brettwooldridge/HikariCP/issues/899#issuecomment-713091912).  

But I also saw that you can implement your own MariaDB [`CredentialPlugin`](https://github.com/mariadb-corporation/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/credential/CredentialPlugin.java) - this might be an option for others upthread who are trying to pull from AWS Secrets Manager, etc - you should be able to do something similar to the [`AwsIamCredentialPlugin`](https://github.com/mariadb-corporation/mariadb-connector-j/blob/master/src/main/java/org/mariadb/jdbc/credential/aws/AwsIamCredentialPlugin.java).  I tested with my own dummy plugin and it seems to work fine - just make sure to register your implementation in `META-INF/services`.  This will only work if you are using MariaDB driver, so not very helpful to the Postgres people - but I figured I would comment in case it helps anyone!

PR #1250, opened 2018-10-01
Fix proxy classes generation reproducibility

Monichev:
Order of interfaces could be different from build to build. It results in different binary CRC for same source code.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/JavassistProxyFactory.java
+++ b/src/main/java/com/zaxxer/hikari/util/JavassistProxyFactory.java
@@ -23,7 +23,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.Statement;
 import java.util.ArrayList;
-import java.util.Arrays;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


src/main/java/com/zaxxer/hikari/util/JavassistProxyFactory.java
```diff
@@ -222,13 +224,13 @@ private static boolean isDefaultMethod(Class<?> intf, CtClass intfCt, CtMethod i
          paramTypes.add(toJavaClass(pt));
       }
 
-      return intf.getDeclaredMethod(intfMethod.getName(), paramTypes.toArray(new Class[paramTypes.size()])).toString().contains("default ");
+      return intf.getDeclaredMethod(intfMethod.getName(), paramTypes.toArray(new Class[0])).toString().contains("default ");
    }
 
    private static Set<Class<?>> getAllInterfaces(Class<?> clazz)
    {
       Set<Class<?>> interfaces = new HashSet<>();
```
johnou: Instead of all the comparator fluff above couldn't we just use a linked hash map here?

Monichev: Yes, this is definitely a better solution.

johnou (approved):

PR #1271, opened 2018-10-31
Add support to ProxyResultSet in order to optimize the findColumn loo…

ppearcy:
…kup on Oracle/SQLServer. Postgres has had a similar optimization for many years

First off, thanks for great lib. My performance testing showed a 200x improvement vs a "safe" tomcat pool setup for the get/close connection case. 

I hope it's not presumptuous to toss this your way without opening an issue, but I figure code is easier to discuss. I'm moving to use HikariCP on a system that supports SQLServer and Oracle. 
Both of which have an inefficient findColumn method that I've had to work around for years and has caused real world perf issues. Postgres has had an optimized method for 10+ years: 
https://github.com/pgjdbc/pgjdbc/blob/265f22b28fd7b5511e30a28ad959d5645e9722cd/pgjdbc/src/main/java/org/postgresql/jdbc/PgResultSet.java#L2624

This ancient artifact discusses the issue and workarounds that were taken:
https://hibernate.atlassian.net/browse/HB-1072

I'm curious if you'd be amenable to including this in HikaricCp. I understand why the answer would be no, as maybe it should be the job of the driver not to be silly with this lookup, but this will likely give some further bragging rights on Oracle/MSSQL :) 

If so, happy to write test cases and make any modifications. 

Things I weren't sure about is if it really made sense to force the optimizeFindColumn boolean down everywhere, but seemed to be the least icky pattern available. Also, made this configurable (or will make it) as it could change some behavior in nuanced ways and there is no reason to ever enable for postgres. 

Thanks!!!


brettwooldridge:
You are correct that I cannot integrate this into HikariCP on principle.  However, at least in the case of SQL Server, the mssql-jdbc driver is now open source (here on GitHub) and I have contributed many performance related fixes in the past 2-3 years.

PR #1272, opened 2018-10-31
Expose the stacktrace created when connection is aquired during leak detection

ppearcy:
Hi Again!
  A small quality of life improvement to allow external users to get to the connection creation stacktrace. This is useful as I have leak detection hooks in my testing framework and when a leaked connection is detected being able to pull the stack trace right from the connection object is very convenient. 

Open to any thoughts or getting this rejected, but figured I'd give it a shot. 

If you are open to this, will add test case and address any feedback.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -184,6 +184,17 @@ public abstract class ProxyConnection implements Connection
       leakTask.cancel();
    }
 
+   /**
+    * Returns the stacktrace location of where the connection was acquired. If leak detection is not endabled, will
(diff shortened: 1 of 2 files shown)


src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
```diff
@@ -184,6 +184,17 @@ void cancelLeakTask()
       leakTask.cancel();
    }
 
+   /**
+    * Returns the stacktrace location of where the connection was acquired. If leak detection is not endabled, will
```
johnou, line 188: endabled => enabled

johnou:
how do the leak detection hooks work? could the existing logging be improved instead?

PR #1283, opened 2018-11-26
Master

ice563102472:
合并主分支代码


lfbayer:
There are too many unnecessary formatting changes. As well as a Main class that probably shouldn't be included. If you want this to be reviewed, please clean it up first.

PR #1306, opened 2018-12-28
update the version for pom demo

taojintianxia:
since HikariCP released the v 3.2.0 , why not show the up to date version.


brettwooldridge:
Fixed.

PR #1315, opened 2019-01-06
Fixes #1012 Local MBean name part should be key-property list

### LINKED ISSUE #1012: Local MBean name part should be key-property list
#### Environment
```
HikariCP version: 2.7.2
JDK version     : 1.8.0_151
```

Same problem as https://stackoverflow.com/questions/42427332/jmxterm-client-hikaricp-beans-instancenotfoundexception-custom-mbean-name

Currently my jmxterm shows names:

```
com.zaxxer.hikari:type=Pool (my-pool)
com.zaxxer.hikari:type=PoolConfig (my-pool)
```

This might be valid with other clients but I haven't found a way to provide this name to jmxterm with any quoting. This would be more consistent with JMX docs on Object Names:

```
com.zaxxer.hikari:type=Pool,name=my-pool
com.zaxxer.hikari:type=PoolConfig,name=my-pool
```

wlfshmn:
This will break client code that depend on the current JMX bean naming, which is an issue if we consider bean naming to be part of the API.

If the renaming is a problem, we could register the same beans twice, but at the cost of clutter.

This allows less fragile querying of JMX beans:

Current

* Single object query: ```new ObjectName("com.zaxxer.hikari:type=Pool (testMBeanReporting)")```
* All objects: ```new ObjectName("com.zaxxer.hikari:type=*")```

New

* Single object query: ```new ObjectName("com.zaxxer.hikari:type=Pool,name=testMBeanReporting")```
* All objects: ```new ObjectName("com.zaxxer.hikari:type=*,name=*")``` or, less explicitly, ```new ObjectName("com.zaxxer.hikari:*")``` 

Additionally, some additional queries become easier to do:

* All beans for a pool: ```new ObjectName("com.zaxxer.hikari:type=*,name=testMBeanReporting")```
* All Pool beans for all pools: ```new ObjectName("com.zaxxer.hikari:type=Pool,name=*")```

The properties based query API also works as intended. (Used in the unit test)

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -274,8 +274,8 @@ abstract class PoolBase
       try {
          final MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
 
-         final ObjectName beanConfigName = new ObjectName("com.zaxxer.hikari:type=PoolConfig (" + poolName + ")");
-         final ObjectName beanPoolName = new ObjectName("com.zaxxer.hikari:type=Pool (" + poolName + ")");
+         final ObjectName beanConfigName = new ObjectName("com.zaxxer.hikari:type=PoolConfig,name=" + poolName);
+         final ObjectName beanPoolName = new ObjectName("com.zaxxer.hikari:type=Pool,name=" + poolName);
          if (register) {
             if (!mBeanServer.isRegistered(beanConfigName)) {
                mBeanServer.registerMBean(config, beanConfigName);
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@wlfshmn I like it, but unfortunately there are probably thousands of applications that depend on the existing MBean name.  I think the only solution will be double registering (yeah, I know).

carstenosu:
Hey guys, this issue has recently come to my attention cause I also would like to use MBean names as proposed above. Instead of double registering, are you opposed to the idea of a flag that would use the new names versus the old? Default that flag to the old so that it doesn't break existing users. Something like this...

HikariConfig.java
        
        private boolean isRegisterMbeans;
        private boolean isKeyPropertyMbeanName;`
   

And in PoolBase.java

         final String configObjectName = "com.zaxxer.hikari:type=PoolConfig (" + poolName + ")";
         final String poolObjectName = "com.zaxxer.hikari:type=Pool (" + poolName + ")";
         final String configPropObjectName = "com.zaxxer.hikari:type=PoolConfig,name=" + poolName;
         final String poolPropObjectName = "com.zaxxer.hikari:type=Pool,name=" + poolName;

         final ObjectName beanConfigName = new ObjectName( config.isKeyPropertyMbeanName() ? configPropObjectName : configObjectName);
         final ObjectName beanPoolName = new ObjectName( config.isKeyPropertyMbeanName() ? poolPropObjectName : poolObjectName);`

PR #1316, opened 2019-01-08
Timeout failures has to be logged on higher than DEBUG level

semenodm:


brettwooldridge:
@semenodm We used to log this at warn level, but a large number in the community complained that this generated too much noise in the logs.  An exception *is* thrown to the user, and presumably the user *is* catching SQLException, so the opportunity to log at a higher level is available.  Frameworks such as Hibernate and Play, just to name two, already do this.

PR #1320, opened 2019-01-25
Fix Readme file with the correct minIdle parameter name

NipunaPrashan:
Fix the minIdle parameter name in README file reffering https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/HikariConfig.java#L71


lfbayer:
I believe the original documentation is correct since the properties use the name from the getter/setter pair, which in this case is `minimumIdle`.

Issue #1337, opened 2019-03-12
isolation level fix for all values accept string includes TRANSACTION_SQL_SERVER_SNAPSHOT_ISOLATION_LEVEL


enesacikoglu:
I am using spring boot with hikariCP but
spring.datasource.hikari.transaction-isolation: SQL_SERVER_SNAPSHOT_ISOLATION_LEVEl
not working,when I used 4096 it works.So I changed it to use string for all transaction-isolation levels.

-----------------------------------------------------------------------------------------
*I will extend solution of this issue https://github.com/brettwooldridge/HikariCP/issues/1024

PR #1338, opened 2019-03-12
#1337 isolation level fix for all values accept string

enesacikoglu:


src/main/java/com/zaxxer/hikari/util/UtilityElf.java
```diff
@@ -165,30 +164,14 @@ public static ThreadPoolExecutor createThreadPoolExecutor(final BlockingQueue<Ru
...
-               case Connection.TRANSACTION_SERIALIZABLE:
-               case Connection.TRANSACTION_NONE:
-               case SQL_SERVER_SNAPSHOT_ISOLATION_LEVEL: // a specific isolation level for SQL server only
-                  return level;
-               default:
-                  throw new IllegalArgumentException();
-             }
-         }
-         catch (Exception e) {
+            final String upperCaseIsolationLevelName = Optional.of(transactionIsolationName.toUpperCase(Locale.ENGLISH))
```
turbanoff: Why `Optional` is used here? It's absolutely redundant

enesacikoglu: Right,please check last commit.

enesacikoglu:
Hello,
Waiting for review.
Thanks.

sadikkuzu:
Congrats @enesacikoglu

brettwooldridge:
@enesacikoglu It looks like this change actually causes a regression.  Previous versions of HikariCP could accept an integer value, such as 4096, as the isolation level.  After this change, only strings of the form "TRANSACTION*" are supported, therefore existing users who were using integer values will now be broken.

It is my fault for not catching this in my review of the code.

enesacikoglu:
> @enesacikoglu It looks like this change actually causes a regression. Previous versions of HikariCP could accept an integer value, such as 4096, as the isolation level. After this change, only strings of the form "TRANSACTION*" are supported, therefore existing users who were using integer values will now be broken.
> 
> It is my fault for not catching this in my review of the code.

Hello @brettwooldridge ,

Actually, I was aware of doing this change because it seems to be not configured correctly. It must be the string instead of number because client can not understand which isolation level refers to that number.

I know , it is a kind of breaking change but I would not want to add a workaround solution inside code for only 4096.

If it is a fault, it is not only your fault.

Finally , we can document this change as a breaking change for the client before to build a new version. I hope it will work.

Thanks
Enes

PR #1378, opened 2019-05-20
Delete redundancy logic.

huangzhendong:
As you see, the delete code is a redundancy logic, so i delete it to make the code to be more clean.


lfbayer:
It is technically possible to have multiple member variables with the same spelling but with different case. This change would potentially break that case. Although unlikely, I don't think it is worth merging the change.

Closing this pull request.

Issue #1212, opened 2018-07-29
I think it is necessary to check the connection status in the post-processing logic of "HikariPool #isConnective Alive".


marcusmoon:
#### Environment
```
HikariCP version: 3.2.0
JDK version     : 1.8.0_141
Database        : MySQL
Driver version  : mysql connector 5.1.46
```

HikariCP is calling "Connection # isValid (timeout)" to check the connection. For mysql, if the response from the request sent for the connection validation does not come within the specified time, the connection is terminated and the "false" is returned.

```
https://gisub.com/mysql/mysql-connector-j/blob/release/5.1/src/com/mysql/jdbc/JDBC4Connection.java #L102
```

Therefore, the cause seems to have been concealed by the code below the HikariCP.

```
https://gisub.com/brettwooldridge/HikariCP/blob/HikariCP-3.2.0/src/main/java/com/zaxxer/hikari/poolBase.ja #L165
```

Because it is already terminated, it appears that the following secondary error occurs by the above code.

```
xxx - Failed to validation connection com.mysql.jdbc.JDBC4Connection @yyy (No opens followed after connection closed). Possibly concider using a storter maxLifetime value.
```

If possible, I think we should first check whether the connection is closed.

```
 }
 finally {
    if(!connection.isClosed()) {
        setNetworkTimeout(connection, networkTimeout);

        if (isIsolateInternalQueries && !isAutoCommit) {
            connection.rollback();
        }
    } else {
        LOGGER.warn("{} - Failed to validate connection {}. Possibly consider using a shorter maxLifetime value.", poolName, connection);
    }
 }
```

mdeinum:
We ran into something similar with Postgres and a mistakenly configured `connectionTestQuery`. The query is always executed even if the `Connection` is already closed. Guarding this method with a `connection.isClosed` would bypass this and prevent our logs from filling up. 

Preferably log an INFO message instead of WARN.

brettwooldridge:
@mdeinum I have intentionally steered clear of using the ``isClosed()`` call in the validation path.  The reason is a little difficult to explain, but I'll give it a shot...

The JDBC specification says nothing about how the ``isClosed()`` call should be implemented by a driver.  Some driver implementations might choose to fire TCP traffic over the wire to determine whether the Connection is open or closed -- because the Connection/socket can be closed unilaterally by the server.  Hopefully, driver developers choose more efficient state tracking than that, but again, it is only a hope.

[...]

PR #1387, opened 2019-06-04
Applies suggested fix in issue-1212 for PoolBase#isConnectionAlive

breskeby:
Hey

this PR applies the proposed fix from @tomdcc here: https://github.com/brettwooldridge/HikariCP/issues/1212#issuecomment-498180296

If you think this should have some better test coverage, would you mind giving me some guidance where this test should live?

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -152,8 +152,7 @@ abstract class PoolBase
                // Some drivers will throw here if the connection is already closed
                setNetworkTimeout(connection, validationTimeout);
             }
-            catch (SQLException sqlEx)
-            {
+            catch (SQLException sqlEx) {
                return false;
             }
 
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -148,13 +148,7 @@ abstract class PoolBase
    {
       try {
          try {
-            try {
-               // Some drivers will throw here if the connection is already closed
-               setNetworkTimeout(connection, validationTimeout);
-            }
-            catch (SQLException sqlEx) {
-               return false;
-            }
+            setNetworkTimeout(connection, validationTimeout);
 
             final int validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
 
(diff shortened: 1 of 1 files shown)


johnou (changes requested):
Please use the same format as the method above.

```java
            try {
               setNetworkTimeout(connection, validationTimeout);
            }
            catch (SQLException ignore) {
               return false;
            }
```

breskeby:
@johnou fixed the formatting.

breskeby:
@johnou Any news on this Pull request? We would really love to see this merged. Again, If you think this should have some better test coverage, would you mind giving me some guidance where this test should live?

johnou:
Merging is up to @brettwooldridge . But looking at this again, the second setNetworkTimeout is in a finally block so I think it's still going to be executed despite returning false early..

johnou:
SQLException doesn't necessarily mean the connection has been closed and the second setNetworkTimeout actually tries to set it to another value.. so I am not 100% on board with the proposed changes from issue-1212.

breskeby:
indeed. I've overseen the call in the finally block. I've moved the behaviour of just returning false into the finally block.

l0rinc:
Is there any progress on this? We could use this fix in our project.

jonnybbb:
@johnou is there anything missing to merge this PR?

l0rinc:
@brettwooldridge, @johnou, we would need this change to be merged, anything we can do to help with that?

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -165,8 +165,13 @@ boolean isConnectionAlive(final Connection connection)
...
          finally {
-            setNetworkTimeout(connection, networkTimeout);
-
+            try {
+               // Some drivers will throw here if the connection is already closed
+               setNetworkTimeout(connection, networkTimeout);
+            }
+            catch (SQLException sqlEx) {
+               return false;
+            }
```
lfbayer, line 174: The catch for the outer try block is already going to catch this exception and return false. Is the purpose to avoid setting the lastConnectionFailure?

I don't actually understand how this change fixes the issue from #1212 ? Can someone explain it to me?

And it's generally a code smell to return from within a finally block, so my instinct (and likely @brettwooldridge 's as well) is to just say no to any code that does so.

lfbayer (changes requested):

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
l0rinc: Valid concern!

I've addressed it in https://github.com/brettwooldridge/HikariCP/pull/1594

lfbayer:
I'm closing this pull request as it is not an appropriate fix for the problem and there has been no movement for a couple years.

I know the underlying concern isn't addressed yet. There is more discussion around this issue in #1594 that is relevant. There is likely a more "correct" solution to be found.

Issue #1394, opened 2019-06-24
Reduce the latency for PoolEntry getState()


he7g:

lfbayer:
Closing. See my comment on the pull request.

PR #1395, opened 2019-06-24
#1394 Reduce the latency for PoolEntry getState()

he7g:


lfbayer:
I'm closing this as there is no explanation for the change. Bypassing the AtomicIntegerFieldUpdater almost certainly would cause concurrency problems.

he7g:
> I'm closing this as there is no explanation for the change. Bypassing the AtomicIntegerFieldUpdater almost certainly would cause concurrency problems.

No, get and set volatile variable directly has no concurrency problem, you can see `AtomicInteger` 's `get()` and `set()` method, it operate value directly, and, you can see `getAndSet` method, it use `Unsafe` to guarantee atomicity.  Actually `AtomicIntegerFieldUpdaterImpl` also use `Unsafe` to implement `getAndSet`,  the principle is same, `AtomicIntegerFieldUpdaterImpl` just conserve memory. Since `AtomicInteger` can `get()` value directly, then `AtomicIntegerFieldUpdater` also can.

lfbayer:
I personally had never actually seen AtomicIntegerFieldUpdater before. I assumed that it must be there for a reason but after reading your comment and the javadocs it sounds likely that it might only be necessary for us in the compareAndSet call.

Did you encounter a noticeable performance problem that led you to identifying this?

he7g:
No, I just found this by reading HikariCP source code. Wish it helps better performance, even just a little.

lfbayer:
Have you run benchmarks before and after the change to see what impact it makes? It seems like it would be better to not make the change unless there is specific evidence showing a performance improvement.

he7g:
Please see and review my benchmarks [here](https://github.com/hebelala/benchmarks/blob/main/src/main/java/com/github/hebelala/benchmarks/AtomicUpdaterTest.java), running results: 

```
Benchmark                            Mode  Cnt           Score           Error  Units
AtomicUpdaterTest.testGetByDirect   thrpt    5  4020425626.332 ± 571737260.153  ops/s
AtomicUpdaterTest.testGetByUpdater  thrpt    5  1120008022.482 ±   4870812.315  ops/s
AtomicUpdaterTest.testSetByDirect   thrpt    5   108161571.240 ±  18097395.693  ops/s
AtomicUpdaterTest.testSetByUpdater  thrpt    5    50682812.876 ±    816756.725  ops/s
```

So, it's better to get and set value directly.

lfbayer:
It's not surprising that it is faster on its own, but I am actually more interested in what the impact on the existing benchmarks is.
https://github.com/brettwooldridge/HikariCP-benchmark

he7g:
Of course it has not influence on the HikariCP performance because its costs is too small and isn't the bottleneck. But it's not the reason to stop us doing things by the right way.

PR #1401, opened 2019-06-26
Fixing initializationFailTimeout.

gierlachg:
An attempt to fix `initializationFailTimeout` which does not work since 3.3.0 after [this change](https://github.com/brettwooldridge/HikariCP/blob/c509ec1a3f1e19769ee69323972f339cf098ff4b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L494).

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -572,16 +572,10 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
             return;
          }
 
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@gierlachg What *specific* behavior was broken by that change?

gierlachg:
Well, the functionality behind `initializationFailTimeout`. As I understand if you start your application and the DB is not there `Hikari` should try to establish an initial connection (before giving up) for up to `initializationFailTimeout` millis. In current version it just _throws_ after first failure.

runenielsen:
@brettwooldridge 

We are experiencing the same problem (that the pool fails on first exception, even though we have set initializationFailTimeout to a high number).

Any ETA on this fix?

Issue #1375, opened 2019-05-11
Change the thread pool size to minimumIdle on blocked initialization


YingLiu4203:
#### Environment
```
HikariCP version: 3.3.1
JDK version     : 1.8.0_111
Database        : PostgreSQL|MySQL|...
Driver version  : x.x.x
```
In the following code 

```Java
if (Boolean.getBoolean("com.zaxxer.hikari.blockUntilFilled") && config.getInitializationFailTimeout() > 1) {
         addConnectionExecutor.setCorePoolSize(Runtime.getRuntime().availableProcessors());
addConnectionExecutor.setMaximumPoolSize(Runtime.getRuntime().availableProcessors());

         final long startTime = currentTime();
         while (elapsedMillis(startTime) < config.getInitializationFailTimeout() && getTotalConnections() < config.getMinimumIdle()) {
            quietlySleep(MILLISECONDS.toMillis(100));
         }

         addConnectionExecutor.setCorePoolSize(1);
         addConnectionExecutor.setMaximumPoolSize(1);
}
```
The thread pool size of the `addConnectionExecutor` should be set as  the value of `minimumIdle`. There are two reasons:  first, `minimumIdle` is the number of connections we need, no matter it is smaller or bigger than the available processors. Second, creating connection is not CPU intensive, most works happen at the network and remote database server and the wait time is much bigger than the compute time. The number of threads should not depend on the number of available CPUs. Brian Goetz in his famous book "Java Concurrency in Practice" recommends the following formula:
 Number of threads = Number of Available Cores * CPU utilization percentage *  (1 + Wait time / compute time)

With the current code, if the `minimumIdle` is smaller than the available CPUs, we create some threads not used. If it is bigger than the available CPUs, it takes longer to create all the required connections because the initialization thread pool is too small.

brettwooldridge:
@YingLiu4203 I find no point of disagreement.  Feel free to submit a Pull Request.

PR #1404, opened 2019-07-02
Change the thread pool size to minimumIdle on blocked initialization

LangInteger:
Update addConnectionExecutor core pool size from available processors to minimum idle connections configured for performance improvement.

Link to issue [Change the thread pool size to minimumIdle on blocked initialization](https://github.com/brettwooldridge/HikariCP/issues/1375)


winflex:
Potential IllegalArgumentException from ThreadPoolExecutor.setMaximumPoolSize?

PR #1405, opened 2019-07-02
Change the thread pool size of `AddConnectionExecutor` to minimumIdle on `hikari` startup phase in blocked initialization

LangInteger:
Update addConnectionExecutor core pool size from available processors to minimum idle connections configured for performance improvement.

Link to issue [Change the thread pool size to minimumIdle on blocked initialization](https://github.com/brettwooldridge/HikariCP/issues/1375).

Improvement for [pull request earlier](https://github.com/brettwooldridge/HikariCP/pull/1404) with test added.


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -129,25 +127,29 @@ public HikariPool(final HikariConfig config)
 
       LinkedBlockingQueue<Runnable> addQueue = new LinkedBlockingQueue<>(config.getMaximumPoolSize());
       this.addConnectionQueue = unmodifiableCollection(addQueue);
-      this.addConnectionExecutor = createThreadPoolExecutor(addQueue, poolName + " connection adder", threadFactory, new ThreadPoolExecutor.DiscardPolicy());
-      this.closeConnectionExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), poolName + " connection closer", threadFactory, new ThreadPoolExecutor.CallerRunsPolicy());
+      this.addConnectionExecutor = createAddConnectionExecutor(threadFactory, addQueue);
+      this.closeConnectionExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), poolName + " connection " +
+            "closer",
+         threadFactory, new ThreadPoolExecutor.CallerRunsPolicy());
```
brettwooldridge: Please do not reformat line 131-133, it is unncessary.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -55,9 +55,7 @@
 import static com.zaxxer.hikari.util.ClockSource.plusMillis;
 import static com.zaxxer.hikari.util.ConcurrentBag.IConcurrentBagEntry.STATE_IN_USE;
 import static com.zaxxer.hikari.util.ConcurrentBag.IConcurrentBagEntry.STATE_NOT_IN_USE;
-import static com.zaxxer.hikari.util.UtilityElf.createThreadPoolExecutor;
-import static com.zaxxer.hikari.util.UtilityElf.quietlySleep;
-import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
+import static com.zaxxer.hikari.util.UtilityElf.*;
```
brettwooldridge: Do not fold imports using ``*`` operator.  Please adjust your IDE to leave imports as explicit.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -129,25 +127,29 @@ public HikariPool(final HikariConfig config)
...
       this.leakTaskFactory = new ProxyLeakTaskFactory(config.getLeakDetectionThreshold(), houseKeepingExecutorService);
 
-      this.houseKeeperTask = houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 100L, housekeepingPeriodMs, MILLISECONDS);
-
-      if (Boolean.getBoolean("com.zaxxer.hikari.blockUntilFilled") && config.getInitializationFailTimeout() > 1) {
-         addConnectionExecutor.setCorePoolSize(Runtime.getRuntime().availableProcessors());
-         addConnectionExecutor.setMaximumPoolSize(Runtime.getRuntime().availableProcessors());
+      this.houseKeeperTask = houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 100L,
+         housekeepingPeriodMs, MILLISECONDS);
+   }
```
brettwooldridge, line 143: Please do not reformat line 137-139, it is unncessary.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -129,25 +127,29 @@ public HikariPool(final HikariConfig config)
...
 
-         final long startTime = currentTime();
-         while (elapsedMillis(startTime) < config.getInitializationFailTimeout() && getTotalConnections() < config.getMinimumIdle()) {
-            quietlySleep(MILLISECONDS.toMillis(100));
-         }
+   private ThreadPoolExecutor createAddConnectionExecutor(ThreadFactory threadFactory,
+                                                          LinkedBlockingQueue<Runnable> addQueue)
```
brettwooldridge: Do not wrap line here.  We allow lines of up to 220 characters.

Parameters should be ``final``.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -129,25 +127,29 @@ public HikariPool(final HikariConfig config)
...
+   {
+      int poosSize = 1;
```
brettwooldridge: Typo.  ``poosSize`` should be ``poolSize``.

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -129,25 +127,29 @@ public HikariPool(final HikariConfig config)
...
-         addConnectionExecutor.setCorePoolSize(1);
-         addConnectionExecutor.setMaximumPoolSize(1);
+      if (Boolean.getBoolean("com.zaxxer.hikari.blockUntilFilled")
+         && config.getInitializationFailTimeout() > 1) {
+         poosSize = config.getMinimumIdle();
       }
+
+      return createThreadPoolExecutorWithFixedSize(addQueue, poolName + " connection adder",
+         threadFactory, new ThreadPoolExecutor.DiscardPolicy(), poosSize);
    }
```
brettwooldridge, line 170: This code substantially changes the behavior of HikariCP.

Currently, HikariCP starts up with a single-threaded "connection adder".  If "com.zaxxer.hikari.blockUntilFilled" is specified, it **_temporarily_** sets the "connection adder" thread pool size to ``availableProcessors()``.  But **_after_** the pool is filled, it returns the thread pool size back to ``1`` (single-threaded).

HikariCP is not designed for multiple "connection adder" threads _while the pool is in use_.  We allow multiple adder threads only here, during initialization, because we know that the pool will not be accessed during this time.

The proposed code does not return the pool size back to ``1`` after the pool is initialized.

However, the most serious issue with this proposed change is that it does not actually "block" the initialization at all when "_com.zaxxer.hikari.blockUntilFilled_" is requested!  The whole point of this flag is to block the initialization thread in the constructor until the pool reaches ``minimumIdle`` connections.

What is going on here?  It makes me suspect that you do not actually understand the code.

brettwooldridge (changes requested):

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
LangInteger: Sorry for the mistake in my previous code, your suspect makes me clearer of situation now.

I don't see the trigger of pool initialization (starting the `addConnectionExecutor` to work) in origin code of this place (actually it is scheduled by the `houseKeeperTask`) , so I treat the `while` loop as a mistake and remove it from the code., which is not a good behavior, sorry again for that. But origin code is apparently wrong if `blockUntilFilled` is true, because the pool executor is always created with a max pool size of 1, change it to available processor num (or `MinimumIdle`) will trigger a `IllegalArgumentException`.

After changing again, all my code do now is just as the pull request's title says.

As for the format and typo, I will treat them more carefully.

Any further suggestions will be highly appreciated.

src/test/java/com/zaxxer/hikari/db/AddConnectionExecutorTest.java
```diff
@@ -0,0 +1,96 @@
...
+         throw new RuntimeException(e);
+      }
+
+      assertEquals("Add connection executor's size equals 1",
+         1, addConnectionExecutor.getCorePoolSize());
+   }
+
+   private static int getRandom(int min, int max){
+      Random random = new Random();
+      return random.nextInt(max) % (max - min + 1) + min;
```
mohamedanees6: Can we use

`int randomInt = ThreadLocalRandom.current().nextInt(min, max+1);`

LangInteger: Accepted, this one seems much better.

LangInteger:
Any further comments? @brettwooldridge

icecreamhead (approved):

icecreamhead:
@brettwooldridge Please can we get this merged and released?

LangInteger:
Make some notes here to clarify the purpose of the pull request

- the first purpose of this pull request is to change the thread pool size of `AddConnectionExecutor` to minimumIdle on `hikari` startup phase in blocked initialization for reasons listed in #1375 
- but a bug was found when building tests, so the creation of `AddConnectionExecutor` was also modified in this pull request

@brettwooldridge

LangInteger:
Since there is no further feedback for a long period of time (the owner may have some potential consideration) and there are now some conflicts related to the main branch, I'll close this pull request. Thx the involved guys.

PR #1410, opened 2019-07-19
= change tips in LOGGER when close

zhaoxi2000:
be ready to do close


lfbayer:
The existing message is grammatically correct.

PR #1413, opened 2019-07-23
Flush ObjectStream before calling toByteArray on underlying ByteArrayOutputStream

emopers:
When an `ObjectStream` instance wraps an underlying `ByteArrayOutputStream` instance, it is recommended to flush or close the `ObjectStream` before invoking the underlying instances's `toByteArray()`.
It is good practice to call `flush`/`close` explicitly as mentioned, for example, [here](http://stackoverflow.com/questions/2984538/how-to-use-bytearrayoutputstream-and-dataoutputstream-simultaneously-java).
This pull request adds a call to the `flush` method before calls to the `toByteArray` methods.


lfbayer:
I'm closing this. Although this might be technically correct, it's just a unit test and is working as is.

Issue #1411, opened 2019-07-20
Connections closing after 5 minutes


wesoos:
#### Environment
```
HikariCP version: 2.7.8
JDK version     : 1.8.0_111
Database        : MySQL|...
Driver version  : 5.1.47
```
<sup>⚠️ Please verify that your issue still occurs on the latest version of HikariCP before reporting. </sup>

-----------------------------------------------------------------------------------------
*Have you searched the CLOSED issues already? How about checking [stackoverflow](https://stackoverflow.com/search?q=HikariCP)?*

Hello,  I am trying to deploy my app (API) to Azure App Service (Linux, Tomact 8.5).  I am using Spring Boot 2.1.6.

The app starts up fine and the connections are created and I can run queries.  However after 5 minutes the connections just close.  They just drop off if they have been idle for 5 minutes.  And Hikari does not create new ones and add back to the pool.

Any ideas?  Below is my application.properties:

`spring.datasource.jdbcUrl=jdbc:mysql://<ip>:3306/database?useSSL=false&characterEncoding=UTF-8&connectionCollation=UTF-8
spring.datasource.username=ssssss
spring.datasource.password=dddddd
spring.datasource.driverClassName=com.mysql.jdbc.Driver

spring.datasource.maximumPoolSize = 15

spring.dataSource.cachePrepStmts=true
spring.dataSource.prepStmtCacheSize=250
spring.dataSource.prepStmtCacheSqlLimit=2048
spring.dataSource.useServerPrepStmts=true
spring.dataSource.useLocalSessionState=true
spring.dataSource.useLocalTransactionState=true
spring.dataSource.rewriteBatchedStatements=true
spring.dataSource.cacheResultSetMetadata=true
spring.dataSource.cacheServerConfiguration=true
spring.dataSource.elideSetAutoCommits=true
spring.dataSource.maintainTimeStats=false`

wesoos:
The pool is not maintaining the minimumIdle value, which by default is supposed the same as the maximumPoolSize value...

wesoos:
So after 5 minutes all connections are closed.  Then when you try to hit the database, I get the following error stack.  However it does then open all the connections again.

[...]

PR #1414, opened 2019-07-23
Add a test case showing how failed connections result in pool failure

timothyjward:
This relates to the issue encountered in #1411 

If connections in the pool are "failed" by the remote database server then they get lazily evicted, however if *all* the connections in the pool are failed then the pool does not attempt to create a new connection to return, but instead causes a failure. This happens even if the connection timeout is easily long enough for a new connection to be created (e.g. 20,000 ms).

This PR includes a test case (failing) demonstrating the issue.

Signed-off-by: Tim Ward <timothyjward@apache.org>


johnou:
cc @brettwooldridge

PR #1430, opened 2019-08-13
use [testcontainers-java] for Postgres unit test

sullis:


sullis:
@brettwooldridge WDYT?

PR #1442, opened 2019-09-04
Make HikariConfig able to retrieve passwords dynamically from a password supplier

tucu00:
#1196 
----
Adds a passwordSupplier property with setter and getter
Refactors the password property to use a built-in passwordSupplier
If setting the password directly, the passwordSupplier property returns NULL
----

This pull request addresses the changes requested by Brett.

My particular use case is using Hashicorp Vault to manage MySQL account passwords... Performing a rotation of the MySQL account passwords in Vault would push the new password to MySQL, the existing connections (created with the old password) will be closed, then Hikari will try to obtain new connections and for that it will call the supplier that will fetch the new password from Vault.

Regarding the issue about the supplier making frequent calls to the system where the password is stored, in our case Vault, calls will be made only when needing to create a new connection, which would typically happen on startup of when an existing connection fails. Said that is up to the supplier implementation to try to be nice by caching if that is acceptable in the specific use case.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -39,6 +39,7 @@ import java.util.TreeSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.function.Supplier;
 
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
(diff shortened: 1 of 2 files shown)


ayushpateldev:
Could you please also add username supplier as well. As some implementations of ephemeral passwords provide a new user every time you request a new credential. 

Current implementation will work fine with 
https://learn.hashicorp.com/tutorials/vault/database-creds-rotation

However will not work with 
https://learn.hashicorp.com/tutorials/vault/database-secrets?in=vault/db-credentials (username also changes). 


Thanks

brotheroftux:
Any updates on this?

lfbayer:
@brotheroftux I had opened a pull request (#2011) as a potential solution to this. I was hoping someone would add their thoughts to the change before I moved forward.

PR #1458, opened 2019-10-01
Set max pool size before core pool size

icecreamhead:
As of JDK11 (tested on openjdk 11 & 12) an exception is thrown when attempting to set core pool size greater than max pool size

```
public void setCorePoolSize(int corePoolSize) {
        if (corePoolSize < 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();
        int delta = corePoolSize - this.corePoolSize;
        this.corePoolSize = corePoolSize;
        if (workerCountOf(ctl.get()) > corePoolSize)
            interruptIdleWorkers();
        else if (delta > 0) {
            // We don't really know how many new threads are "needed".
            // As a heuristic, prestart enough new workers (up to new
            // core size) to handle the current number of tasks in
            // queue, but stop if queue becomes empty while doing so.
            int k = Math.min(delta, workQueue.size());
            while (k-- > 0 && addWorker(null, true)) {
                if (workQueue.isEmpty())
                    break;
            }
        }
    }
```

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -138,8 +138,8 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
       this.houseKeeperTask = houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 100L, housekeepingPeriodMs, MILLISECONDS);
 
       if (Boolean.getBoolean("com.zaxxer.hikari.blockUntilFilled") && config.getInitializationFailTimeout() > 1) {
-         addConnectionExecutor.setCorePoolSize(Math.min(16, Runtime.getRuntime().availableProcessors()));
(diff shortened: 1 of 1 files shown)


quaff:
Will this conflict with https://github.com/brettwooldridge/HikariCP/commit/e8e9cdd00ab886b26340c05706485ddb092f5116 ?

icecreamhead:
That commit doesn't make sense to me. It appears to do nothing other than introduce the bug that my commit fixes.

johnou:
@icecreamhead please rebase.

icecreamhead:
@johnou done

johnou:
@brettwooldridge ping

icecreamhead:
@brettwooldridge any chance of a review on this? Is this repo even maintained any more?

johnou:
just noticed there is https://github.com/brettwooldridge/HikariCP/pull/1605 too which applies the fix in two places

brettwooldridge:
Merged #1605

PR #1469, opened 2019-10-14
Upgrade Micrometer to version 1.0.11

edysli:
Upgrade Micrometer from version 1.0.0 to 1.0.11. Version 1.0.11 is the last patch in the 1.0.x line.

Since Micrometer releases follow semantic versioning (see their [support policy](https://micrometer.io/docs/support)), 1.0.11 should not contain breaking changes.

An upgrade to the latest release (1.3.0) could be considered instead.


edysli:
Travis CI failed to install JDK 11... :thinking:

edysli:
no longer relevant

PR #1477, opened 2019-10-29
Add transaction level enum

kollstrom:
Would it be nice to add an override to `HikariConfig.setTransactionLevel()` that let's you use the `IsolationLevel` enum instead of having to type out the isolation level yourself?

Please let me know if this is viable at all. If it is I'm very interested in discussing how this can be solved in an even better way.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/IsolationLevel.java
+++ b/src/main/java/com/zaxxer/hikari/util/IsolationLevel.java
@@ -16,22 +16,28 @@
 
 package com.zaxxer.hikari.util;
 
-public enum IsolationLevel
-{
-   TRANSACTION_NONE(0),
-   TRANSACTION_READ_UNCOMMITTED(1),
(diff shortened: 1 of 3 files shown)


quaff:
Why not reuse enum.name() instead of introducing `levelName` ?

kollstrom:
> Why not reuse enum.name() instead of introducing `levelName` ?

Why, I because I love adding unnecessary code I guess🤦‍♂️
Thanks! I've changed it now, @quaff

src/main/java/com/zaxxer/hikari/util/IsolationLevel.java
```diff
@@ -16,8 +16,7 @@
 
 package com.zaxxer.hikari.util;
 
-public enum IsolationLevel
-{
+public enum IsolationLevel {
```
johnou: Nit, revert this.

johnou (approved):

Issue #1496, opened 2019-11-19
Driver properties with objects


joaopmerlin:
We have a custom driver here at the company. This driver expects some properties that are instances of classes. What happens is that hikari datasource is calling toString() in the properties.
What would be the impact of transforming this code
`driverProperties.setProperty(entry.getKey().toString(), entry.getValue().toString());`
in
`driverProperties.put(entry.getKey().toString(), entry.getValue());`
If this implementation has no impact, I can send a pull request.

#### Environment
```
HikariCP version: 3.4.1
JDK version     : 1.8.0_111
Database        : Custom
Driver version  : Custom
```
<sup>⚠️ Please verify that your issue still occurs on the latest version of HikariCP before reporting. </sup>

-----------------------------------------------------------------------------------------
*Have you searched the CLOSED issues already? How about checking [stackoverflow](https://stackoverflow.com/search?q=HikariCP)?*

PR #1557, opened 2020-03-18
chore: fix driver properties issue with only String values

mbasiuk-talend:
This PR fixes:
https://github.com/brettwooldridge/HikariCP/issues/1496

This issue makes a problem with Snowflake Driver when PrivateKey is passing as a String value instead of actual PrivateKey instance here:
https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/core/SFSessionProperty.java#L143

https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/core/SFSessionProperty.java#L35

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
@@ -46,7 +46,7 @@ public final class DriverDataSource implements DataSource
       this.driverProperties = new Properties();
 
       for (Entry<Object, Object> entry : properties.entrySet()) {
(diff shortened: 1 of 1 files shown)


A248:
You may also want to change PropertyElf for consistency:
https://github.com/brettwooldridge/HikariCP/blob/30c63c637229e11f050721372abc3f9c9d670b66/src/main/java/com/zaxxer/hikari/util/PropertyElf.java#L116

lfbayer:
I'm closing this pull request. #1895 covers the same issue and has more discussion.

PR #1581, opened 2020-05-11
Handle short in PropertyElf

freafrea:
When the dataSource class has a method setX(short Y) then PropertyElf did not handle this case. Added similarly to int/long/boolean handling. Added tests.


freafrea:
Hi, is there something that I should do, so you could merge this? :)

johnou:
@freafrea which datasource method are you invoking which uses short?

freafrea:
I have a jdbc driver in which setPort has signature void setPort(short port). It is an old one, but I still have to use it.

johnou (approved):

PR #1584, opened 2020-05-15
Simplified build of multi-release JAR

cowwoc:
I found a way to use maven-compiler-plugin to build multi-release JARs. The build file is much cleaner now and should be easier to maintain.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -325,7 +325,7 @@
                <executable>java</executable>
                <arguments>
                   <argument>-cp</argument>
-                  <argument>${project.build.outputDirectory}:${maven.compile.classpath}</argument>
+                  <argument>${project.build.outputDirectory}${path.separator}${maven.compile.classpath}</argument>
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@cowwoc Do I need to be concerned about this?
```
[INFO] --- maven-bundle-plugin:5.1.1:bundle (default-bundle) @ HikariCP ---
[WARNING] Bundle com.zaxxer:HikariCP:bundle:3.4.6-SNAPSHOT : Classes found in the wrong directory: {META-INF/versions/11/module-info.class=module-info}
```

cowwoc:
@brettwooldridge I don't think so. This sounds like a false warning on the part of the maven-bundle-plugin. You can find a workaround here: https://issues.apache.org/jira/browse/FELIX-5592?focusedCommentId=16281169&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16281169

Issue #1582, opened 2020-05-13
update module-info to mark optional dependencies as 'static'


prashantbhat:
After updating to the release version 3.4.5, I get the following error. (using Java 11)

> Error occurred during initialization of boot layer
java.lang.module.FindException: Module metrics.healthchecks not found, required by com.zaxxer.hikari

The `module-info` (release 3.4.5) defines following dependencies with `requires` directive:
```
requires hibernate.core;
requires javassist;
requires simpleclient;
requires metrics.core;
requires metrics.healthchecks;
requires micrometer.core;
```

For HikariCP, these dependencies are optional, as in the pom.xml.

Adding the `static` modifier makes them optional, as defined in [java.lang.module package summary](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/module/package-summary.html)
> 'requires' directives that have the 'static' modifier express an optional dependence at run time

So, it will be very useful if these dependencies are changed to be `static`. Thanks.

smironov:
Same issue on Java 14, using HikariCP 3.4.5

brettwooldridge:
@prashantbhat @smironov Can one of you send a pull request?

prashantbhat:
Opened PR #1587 with an update to latest Slf4J to use it's correct module name. Please review.

guyv:
@brettwooldridge any chance of getting this merged and released?

Frank-CG:
Is there any workaround for this error?

Borwe:
It's December, and still no one is close?

FanJups:
@prashantbhat  Thanks for the update but I still has the same error

I work on a project using JAVA 11, MAVEN and JAVAFX13.

module-info.java looks like : 

```
module com.mautourco.finance {
	requires javafx.controls;
	requires javafx.fxml;
	requires java.sql;
	requires javafx.base;
	requires javafx.graphics;
	requires com.zaxxer.hikari;
	
	opens com.mautourco.finance to javafx.fxml;
	opens com.mautourco.finance.controller to javafx.fxml;
	opens com.mautourco.finance.model to javafx.base;

	exports com.mautourco.finance;
}
```

I tried to add **requires com.zaxxer.hikari.metrics** but I got this : **com.zaxxer.hikari.metrics cannot be resolved to a module**

Please, any idea ?

FanJups:
Downgrading to 2.4.7 works for me 

```
		<dependency>
[...]

PR #1587, opened 2020-05-18
#1582: update module-info to mark optional dependencies as 'static'

prashantbhat:
Also updated the Slf4J version to [release 1.7.30](http://www.slf4j.org/news.html) as it has the `Automatic-Module-Name` defined i.e. `org.slf4j` for slf4j-api jar


Dragas:
As of 5.4, hibernate also has an automatic module name `org.hibernate.orm.core`

Kamefrede:
any ETA as to when this will get merged?

Issue #1403, opened 2019-07-01
Failed to validate connection org.mariadb.jdbc.MariaDbConnection (Connection.setNetworkTimeout cannot be called on a closed connection).


duplxey:
#### Environment
```
HikariCP version: 3.3.1
JDK version     : 1.8.0_111
Database        : 5.5.60-MariaDB
Driver version  : 2.4.2 (org.mariadb.jdbc.mariadb-java-client)
```
-----------------------------------------------------------------------------------------
#### Problem
HikariCP & database connection works well as long as it doesn't idle for more than ~10 minutes. If you try to execute a query after 10 minutes of idling the following warnings & errors will be thrown:
```
[JDA MainWS-ReadThread] WARN com.zaxxer.hikari.pool.PoolBase - HikariPool-1 - Failed to validate connection org.mariadb.jdbc.MariaDbConnection@2b6d3853 (Connection.setNetworkTimeout cannot be called on a closed connection). Possibly consider using a shorter maxLifetime value.
java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 30025ms.
	at com.zaxxer.hikari.pool.HikariPool.createTimeoutException(HikariPool.java:697)
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:196)
	at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:161)
	at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:100)
	at net.cubeverse.cvdiscordbot.database.Database.getConnection(Database.java:45)
	at net.cubeverse.cvdiscordbot.database.DiscordVerifyDAO.isVerified(DiscordVerifyDAO.java:81)
	at net.cubeverse.cvdiscordbot.command.discord.VerifyCommand.execute(VerifyCommand.java:25)
	at net.cubeverse.cvdiscordbot.core.command.manager.CommandManager.executeDiscordCommand(CommandManager.java:34)
	at net.cubeverse.cvdiscordbot.listeners.MessageListeners.onPrivateMessageReceived(MessageListeners.java:21)
	at net.dv8tion.jda.core.hooks.ListenerAdapter.onEvent(ListenerAdapter.java:397)
	at net.dv8tion.jda.core.hooks.InterfacedEventManager.handle(InterfacedEventManager.java:84)
[...]

Issue #1328, opened 2019-02-18
Connection.setNetworkTimeout cannot be called on a closed connection


sooyoung32:
#### Environment
```
HikariCP version: 3.2.0
JDK version     : 1.8.0_161
Database        : Aurora MySql 5..7
Driver version  : Maria JDBC ConnectorJ 2.2.6
```
-----------------------------------------------------------------------------------------

My app configurations are listed below.
spring boot version : 2.0.8
```
spring:
  datasource:
    hikari:
      jdbc-url: jdbc:mariadb:aurora:dbUrl:port/dbNanme?connectTimeout=10&validConnectionTimeout=10&socketTimeout=60000
      username:
      password: 
      validation-timeout: 1000 # must be less than the connectionTimeout
      max-lifetime: 580000
      maximum-pool-size: 15  # minimumIdle same as maximumPoolSize
      connection-timeout: 2000
````

And in HikariCP, there are warnings and errors in the pattern below.

```
2019-02-18 00:30:14.667 WARN 23197 --- [nio-5000-exec-8] com.zaxxer.hikari.pool.PoolBase : conn-pool - Failed to validate connection org.mariadb.jdbc.MariaDbConnection@721a4c34 (Connection.setNetworkTimeout cannot be called on a closed connection). Possibly consider using a shorter maxLifetime value.
2019-02-18 00:30:15.671 WARN 23197 --- [nio-5000-exec-8] com.zaxxer.hikari.pool.PoolBase : conn-pool - Failed to validate connection org.mariadb.jdbc.MariaDbConnection@6b173c77 (Connection.setNetworkTimeout cannot be called on a closed connection). Possibly consider using a shorter maxLifetime value.

2019-02-18 00:30:15.673 ERROR 23197 
org.springframework.transaction.CannotCreateTransactionException: Could not open JDBC Connection for transaction; nested exception is java.sql.SQLTransientConnectionException: conn-pool - Connection is not available, request timed out after 2006ms.
at org.springframework.jdbc.datasource.DataSourceTransactionManager.doBegin(DataSourceTransactionManager.java:305)
at org.springframework.transaction.support.AbstractPlatformTransactionManager.getTransaction(AbstractPlatformTransactionManager.java:378)
at org.springframework.transaction.interceptor.TransactionAspectSupport.createTransactionIfNecessary(TransactionAspectSupport.java:474)
[...]

PR #1594, opened 2020-05-29
Fix setNetworkTimeout throwing during isConnectionAlive

### LINKED ISSUE #1212: I think it is necessary to check the connection status in the post-processing logic of "HikariPool #isConnective Alive".

l0rinc:
Fixes https://github.com/brettwooldridge/HikariCP/issues/1212

---

Seeing that the Postgres JDBC driver [explicitly checks for closed connections](https://github.com/pgjdbc/pgjdbc/blob/master/pgjdbc/src/main/java/org/postgresql/jdbc/PgConnection.java#L1639) during `setNetworkTimeout`:
```
Caused by: org.postgresql.util.PSQLException: This connection has been closed.
    at org.postgresql.jdbc.PgConnection.*checkClosed*(PgConnection.java:857)
    at org.postgresql.jdbc.PgConnection.setNetworkTimeout(PgConnection.java:1639)
    at com.zaxxer.hikari.pool.PoolBase.setNetworkTimeout(PoolBase.java:552)
    at com.zaxxer.hikari.pool.PoolBase.isConnectionAlive(PoolBase.java:168)
    at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:185)
```

and that we're already aware of that when explicitly closing connections in https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L143
> connection.close(); // continue with the close even if setNetworkTimeout() throws

and

https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L531

----------

I've added a `try` method that delegates to previous one, turning the exception to a boolean
and using it in
* [quietlyCloseConnection](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L137)
* [isConnectionAlive](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L156-L173)

and continue throwing in case of:

* [resetConnectionState](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L234)
* [setupConnection](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L399-L426)

----------

I've separated the failing test to a dedicated commit so that the reviewer can make sure the issue was reproduced.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -133,15 +133,8 @@ abstract class PoolBase
          try {
             logger.debug("{} - Closing connection {}: {}", poolName, connection, closureReason);
 
-            try {
-               setNetworkTimeout(connection, SECONDS.toMillis(15));
-            }
-            catch (SQLException e) {
-               // ignore
-            }
-            finally {
-               connection.close(); // continue with the close even if setNetworkTimeout() throws
-            }
+            trySetNetworkTimeout(connection, SECONDS.toMillis(15));
+            connection.close();
          }
          catch (Exception e) {
             logger.debug("{} - Closing connection {} failed", poolName, connection, e);
@@ -153,7 +146,7 @@ abstract class PoolBase
    {
       try {
          try {
-            setNetworkTimeout(connection, validationTimeout);
+            trySetNetworkTimeout(connection, validationTimeout);
 
             final int validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
 
@@ -170,7 +163,7 @@ abstract class PoolBase
             }
          }
          finally {
-            setNetworkTimeout(connection, networkTimeout);
+            trySetNetworkTimeout(connection, networkTimeout);
 
             if (isIsolateInternalQueries && !isAutoCommit) {
                connection.rollback();
@@ -547,7 +540,7 @@ abstract class PoolBase
    }
 
    /**
-    * Set the network timeout, if <code>isUseNetworkTimeout</code> is <code>true</code> and the
+    * Set the network timeout, if <code>isNetworkTimeoutSupported</code> is <code>true</code> and the
     * driver supports it.
     *
     * @param connection the connection to set the network timeout on
@@ -558,6 +551,26 @@ abstract class PoolBase
    {
       if (isNetworkTimeoutSupported == TRUE) {
          connection.setNetworkTimeout(netTimeoutExecutor, (int) timeoutMs);
+      }
+   }
+
+   /**
+    * Set the network timeout, if <code>isNetworkTimeoutSupported</code> is <code>true</code> and the
+    * driver supports it.
+    *
+    * @param connection the connection to set the network timeout on
+    * @param timeoutMs  the number of milliseconds before timeout
+    * @return whether the timeout was successfully applied
+    */
+   private boolean trySetNetworkTimeout(final Connection connection, final long timeoutMs)
+   {
+      try {
+         setNetworkTimeout(connection, timeoutMs);
+         return true;
+      }
+      catch (SQLException e) {
+         // Connection was already closed
+         return false;
       }
    }
 
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -181,8 +181,8 @@ abstract class PoolBase
       }
       catch (Exception e) {
          lastConnectionFailure.set(e);
-         logger.warn("{} - Failed to validate connection {} ({}). Possibly consider using a shorter maxLifetime value.",
-                     poolName, connection, e.getMessage());
+         logger.warn("{} - Failed to validate connection {}. Possibly consider using a shorter maxLifetime value.",
+            poolName, connection, e);
          return false;
       }
    }
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -133,8 +133,11 @@ abstract class PoolBase
          try {
             logger.debug("{} - Closing connection {}: {}", poolName, connection, closureReason);
 
-            trySetNetworkTimeout(connection, SECONDS.toMillis(15));
-            connection.close();
+            try {
+               trySetNetworkTimeout(connection, SECONDS.toMillis(15));
+            } finally {
+               connection.close();
+            }
          }
          catch (Exception e) {
             logger.debug("{} - Closing connection {} failed", poolName, connection, e);
@@ -146,7 +149,10 @@ abstract class PoolBase
    {
       try {
          try {
-            trySetNetworkTimeout(connection, validationTimeout);
+            boolean timeoutSet = trySetNetworkTimeout(connection, validationTimeout);
+            if (!timeoutSet) {
+               return false;
+            }
 
             final int validationSeconds = (int) Math.max(1000L, validationTimeout) / 1000;
 
@@ -562,14 +568,15 @@ abstract class PoolBase
     * @param timeoutMs  the number of milliseconds before timeout
     * @return whether the timeout was successfully applied
     */
-   private boolean trySetNetworkTimeout(final Connection connection, final long timeoutMs)
+   private boolean trySetNetworkTimeout(final Connection connection, final long timeoutMs) throws SQLException
    {
       try {
          setNetworkTimeout(connection, timeoutMs);
          return true;
-      }
-      catch (SQLException e) {
-         // Connection was already closed
+      } catch (SQLException e) {
+         if (!connection.isClosed()) {
+            throw e;
+         }
          return false;
       }
    }
--- a/src/test/java/com/zaxxer/hikari/pool/TestConnectionTimeoutRetry.java
+++ b/src/test/java/com/zaxxer/hikari/pool/TestConnectionTimeoutRetry.java
@@ -223,45 +223,73 @@ public class TestConnectionTimeoutRetry
    @Test
    public void testConnectionIdleFill() throws Exception
    {
-      StubConnection.slowCreate = false;
+      try {
+         StubConnection.slowCreate = false;
 
-      HikariConfig config = newHikariConfig();
-      config.setMinimumIdle(5);
-      config.setMaximumPoolSize(10);
-      config.setConnectionTimeout(2000);
-      config.setValidationTimeout(2000);
-      config.setConnectionTestQuery("VALUES 2");
-      config.setDataSourceClassName("com.zaxxer.hikari.mocks.StubDataSource");
+         HikariConfig config = newHikariConfig();
+         config.setMinimumIdle(5);
+         config.setMaximumPoolSize(10);
+         config.setConnectionTimeout(2000);
+         config.setValidationTimeout(2000);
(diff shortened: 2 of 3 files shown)


src/test/java/com/zaxxer/hikari/pool/TestConnectionTimeoutRetry.java
```diff
@@ -223,46 +223,74 @@ public void run()
...
+         StubConnection.setNetworkTimeoutThrows = true; // needs to be after ds init, otherwise setting timeouts is disabled
 
-            sleep(1300);
-
-            assertSame("Total connections not as expected", 10, pool.getTotalConnections());
-            assertSame("Idle connections not as expected", 3, pool.getIdleConnections());
+         try (Connection connection1 = ds.getConnection()) {
+            assertNotNull(connection1);
+         } catch (Exception e) {
+            e.printStackTrace();
```
l0rinc: Failure proof:
Screen Shot 2020-05-29 at 2 10 29 PM
Screen Shot 2020-05-29 at 2 28 29 PM

lfbayer:
The only functional difference I see in this pull request is that connection.close() will no longer be called when exceptions other than SQLException are thrown. Is that the sole purpose of the change? As I said in the other pull request (#1387), I don't understand how this fixes the bug. 

With this code, connection.close() will still get called after the SQLException happens, and that would presumably also throw a SQLException again. I understood the original issue as being that an exception from connection.close() is masking a separate exception.

Any exception thrown from setNetworkTimeout that is an instanceof SQLException is already ignored, so it certainly isn't a question of trying to ignore those exceptions.

Am I missing something trivial?

l0rinc:
Hey @lfbayer, thanks for your quick response!

This PR won't fix the underlying cause of `getConnection` not being able to return a live connection, you're right about that. That needs individual investigations.

The concept here was rather to treat exceptions inside `isConnectionAlive` similarly to `quietlyCloseConnection`, since in those cases an already-closed-connection is not an exceptional state.

We also suspect some underlying cause may be hidden by these spurious logs, so it's rather meant as a log sanitization step.

Would you like me to change the impl of `isConnectionAlive` to return quickly in case `trySetNetworkTimeout` already returns `false`?

lfbayer:
I still am failing to see how this pull request changes anything. Any log message that would have been logged before will still be logged.  And any exception that was is ignored will still be ignored. Except for the fact that now connection.close() is no longer in a finally, thus causing the close call to be not called when an exception other than SQLException is thrown. 

Are you saying that the intent of this change is to make it so that connection.close() is no longer called for any exception that is not a SQLException? Because that is what I see this change as doing.

l0rinc:
If an exception happens in `isConnectionAlive`, even though the catch block will correctly return `false`, it will also [override the last connection failure](https://github.com/brettwooldridge/HikariCP/blob/009a7013cf75b69290abb9bd24b12b66b55a07bf/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L183), possibly masking the previous one, which could have been the real source of the error.

By making `isConnectionAlive` non-exceptional for closed connections, they won't swap out the underlying cause for why the connection was closed in the first place - at least this is our current assumption, what do you think?

> causing the close call to be not called when an exception other than SQLException is thrown.

Lemme' fix that, good point!

l0rinc:
Pushed, changed the test to validate that the extra logs are gone.

abhikiki:
When this change will be released?

lfbayer:
I'm still waiting on an explanation of why this change is needed and what problem it solves.

l0rinc:
> I'm still waiting on an explanation of why this change is needed and what problem it solves.

Currently exceptions in `isConnectionAlive` override the last connection failure, possibly masking the previous one, which could have been the real source of the error.

This PR whitelists closed connection related exceptions in `isConnectionAlive`, similarly to `quietlyCloseConnection`.

@lfbayer, does that explain the change?

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -170,7 +169,7 @@ boolean isConnectionAlive(final Connection connection)
             }
          }
          finally {
-            setNetworkTimeout(connection, networkTimeout);
+            trySetNetworkTimeout(connection, networkTimeout);
```
lfbayer, line 172: I don't know if it is a probable code path, but by no longer ignoring the SqlException now the `connection.rollback()` call might not be called when it would have been before.

It's probably better not to change this code if the change doesn't serve a specific purpose.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -134,13 +134,9 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
...
-               setNetworkTimeout(connection, SECONDS.toMillis(15));
-            }
-            catch (SQLException e) {
-               // ignore
-            }
-            finally {
-               connection.close(); // continue with the close even if setNetworkTimeout() throws
+               trySetNetworkTimeout(connection, SECONDS.toMillis(15));
+            } finally {
+               connection.close();
```
lfbayer, line 139: I believe it is intentional to ignore any SQLException here. If the connection is already closed and setNetworkTimeout throws the exception because of this, then the exception is ignored. I don't think this code need to be changed.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -153,7 +149,10 @@ boolean isConnectionAlive(final Connection connection)
    {
       try {
          try {
-            setNetworkTimeout(connection, validationTimeout);
+            boolean timeoutSet = trySetNetworkTimeout(connection, validationTimeout);
+            if (!timeoutSet) {
+               return false;
+            }
```
lfbayer, line 155: It's hard to describe, but something about this code just seems off to me.

The original code seems more correct, in that it would throw an exception if there is an error. This I believe is the real exception that the original issue is about. This is the exception that you want to be seeing.

The reason it is eaten is because the setNetworkTimeout call in the finally below would then throw another exception.

My gut tells me that this call should actually be one line up, just before the try rather than inside the try block. This way if the first setNetworkTimeout call throws, it won't try to call setNetworkTimeout a second time.

But then the rollback call at the bottom of the finally wouldn't get called. I don't know if it would need to be, but it is a change in behavior.

So actually it might just be that there should be one more try/finally added to this method so that the rollback call can be in a finally all by itself.

lfbayer (changes requested):

lfbayer:
Also, it would appear that the exception is being intentionally suppressed: https://github.com/brettwooldridge/HikariCP/issues/1328#issuecomment-465682190

So whatever the solution is has to consider all of the other issues in the past.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
acurvers: about the SQL exception that is ignored here, without concerning myself about the context of the discussion, I was thinking is the SQL exception not a log worthy event. Should it not be logged at least at trace level, or even a higher level be considered? 
Or perhaps I am mistaken?

aperger:
I got the the following error with  MariaDB (Client v2.6.2 + HirakiCP v3.4.5 + default connection configuration of Spring Boot v2.2.8):

```log
2020-10-05 16:05:00.003  WARN 11152 --- [pool-1-thread-2] com.zaxxer.hikari.pool.PoolBase          : poolDbKOMP - Failed to validate connection org.mariadb.jdbc.MariaDbConnection@38a19932 ((conn=1022572) Connection.setNetworkTimeout cannot be called on a closed connection). Possibly consider using a shorter maxLifetime value.
```

brettwooldridge:
I'd like to comment on the idea that ...
> Currently exceptions in isConnectionAlive override the last connection failure, possibly masking the previous one, which could have been the real source of the error.

It *is* the case that there is *only one* ``lastConnectionFailure``. Because the pool is accessed in a highly multi-threaded manner, the idea that the "previous" failure, or *even the current* one, is specifically related to the current thread requesting a connection is spurious. It is merely a "guide post" as to errors occurring.

Whatever the ``lastConnectionFailure``, it was definitely the cause of a "recent" failure, with no guarantee that it was the cause of the current thread's failed request.

If *that* is the issue you are trying to solve, I suggest an entirely different change that removes the "global" ``lastConnectionFailure``, and replaces it with a ThreadLocal variant that therefore guarantees that the failure reported is in fact the one that occurred on the requesting thread.

UPDATE: Note, that if the "current request thread" encounters an error in ``isConnectionAlive()`` that results in the necessity of a new connection to be added, and the Adder thread itself encounters exceptions establishing a new connection, there is no way to reliably associate *that* failure with the "current request thread". A ThreadLocal will not help in this case, because there can be several requests enqueued for the Adder thread, and which request thread gets which connection created is completely non-deterministic (at the whim of the OS scheduler).

UPDATE2: Ultimately, this may be a *fools errand* trying to improve the ``lastConnectionFailure`` semantics. The truth is this, the ultimate cause of any timeout is more simply stated as "the failure of a connection to be *returned to the pool*, or *added to the pool*, in a timely fashion."  When a failure in ``isConnectionAlive()`` occurs, the thread goes back to waiting for one of those two events (a returned connection or a new connection). If neither of those occurs "in time", then a timeout exception occurs. The "failure" in ``isConnectionAlive()`` is only the reason *that specific connection* was dead, but is only tangentially related to why the requesting thread could not obtain a connection in the end.

mauromol:
I perfectly agree with UPDATE2 from @brettwooldridge , that I fully understand only now that I have concluded my own analysis at https://github.com/brettwooldridge/HikariCP/issues/1403#issuecomment-967046208.

I think this is causing a lot of confusion on end users, if you search for "setNetworkTimeout" issues there are a lot of open issues for this.

lfbayer:
I believe this is fixed with commit a7f9a4b2.

PR #1599, opened 2020-06-10
Add OpenTracing support

### LINKED ISSUE #1499: OpenTracing integration
It would be nice for HikariCP to integrate [OpenTracing](https://opentracing.io/).
Tracking time spent with tasks like connection acquisition would be useful.

A [Tracer](https://opentracing.io/docs/overview/tracers/#tracer-interface) could be provided on Initialization and [Spans](https://opentracing.io/docs/overview/spans/) could be created for any operation which can take long time.

tobias-bahls:
This PR adds an `IMetricsTracker` implementation that logs connection requests/acquisitions/timeouts on the currently active OpenTracing span.

To make this possible, I needed to change on `IMetricsTracker`. Previously, the implementation assumed that every connection timeout should also result in a `recordConnectionAcquiredNanos` call. When logging these events as-is on OpenTracing spans, the logs on the span contain "requested, acquired, timeout", which is incorrect.

This PR moves the recording of "connection acquired" on timeouts to the individual `IMetricsTrackers`. This required a breaking change on `IMetricsTracker` to add the elapsed nanos to `recordConnectionTimeout`.

A non-breaking alternative would be to add `recordBorrowTimeoutStats` to `IMetricsTracker`, but I felt this is exactly the same as `recordConnectionTimeout`, so I went with the breaking change. If `IMetricsTracker` should remain stable, I can change the PR to include the non-breaking alternative.

Fixes https://github.com/brettwooldridge/HikariCP/issues/1499

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,7 @@
       <maven.release.version>2.5.3</maven.release.version>
       <metrics.version>3.2.4</metrics.version>
       <micrometer.version>1.3.2</micrometer.version>
+      <opentracing.version>0.33.0</opentracing.version>
       <simpleclient.version>0.0.26</simpleclient.version>
       <mockito.version>3.2.4</mockito.version>
       <pax.exam.version>4.13.1</pax.exam.version>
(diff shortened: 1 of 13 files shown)


quaff:
Have you tried https://github.com/opentracing-contrib/java-jdbc ?

tobias-bahls:
I'm mostly interested seeing the time between "connection requested" and "connection acquired". in my spans. I'm not sure if `java-jdbc` would provide this

quaff:
> I'm mostly interested seeing the time between "connection requested" and "connection acquired". in my spans. I'm not sure if `java-jdbc` would provide this

https://github.com/opentracing-contrib/java-jdbc/blob/master/src/main/java/io/opentracing/contrib/jdbc/TracingDataSource.java#L65

tobias-bahls:
I don't think this PR is the right place for discussing this. `java-jdbc` also doesn't work because I don't need the statement tracing. I'd also need a way to customize the operation names and tags to comply with our internal guidelines.

brettwooldridge:
@tobias-bahls I cannot promise to merge this. However, it would be essential for the change to be non-breaking of existing users. Additionally, the dependency on OpenTracing would need to be optional, not required.

PR #1605, opened 2020-06-22
HikariPool - fix major bug with usage of com.zaxxer.hikari.blockUntilFilled in JDK 11 and later

sw-pachanady:
Due to a change in JDK11 in ThreadPoolExecutor setMaximumPoolSize needs to be done before setCorePoolSize, the current implementation throws IllegalArgumentException and hence prefilling of connections by setting com.zaxxer.hikari.blockUntilFilled is broken.

HikariPool -line 133 - addConnectionExecutor initializes the ThreadPoolExecutor with 1 core and 1 max thread.

Then we get an IllegalArgumentException in line 141while trying to set setCorePoolSize due to check

NOTE: This bug and PR was submitted by some other developers in the past I think, but it's still not fixed.

 if (corePoolSize < 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();

 1 < some number of core pools make it throw IllegalArgumentException

Hence setting "com.zaxxer.hikari.blockUntilFilled "  to true and trying to prefill connections fails.
This feature was working fine in JDK 8. Kindly review and merge this pull request.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -138,16 +138,16 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
       this.houseKeeperTask = houseKeepingExecutorService.scheduleWithFixedDelay(new HouseKeeper(), 100L, housekeepingPeriodMs, MILLISECONDS);
 
       if (Boolean.getBoolean("com.zaxxer.hikari.blockUntilFilled") && config.getInitializationFailTimeout() > 1) {
-         addConnectionExecutor.setCorePoolSize(Math.min(16, Runtime.getRuntime().availableProcessors()));
          addConnectionExecutor.setMaximumPoolSize(Math.min(16, Runtime.getRuntime().availableProcessors()));
+         addConnectionExecutor.setCorePoolSize(Math.min(16, Runtime.getRuntime().availableProcessors()));
 
          final long startTime = currentTime();
          while (elapsedMillis(startTime) < config.getInitializationFailTimeout() && getTotalConnections() < config.getMinimumIdle()) {
             quietlySleep(MILLISECONDS.toMillis(100));
          }
 
-         addConnectionExecutor.setCorePoolSize(1);
          addConnectionExecutor.setMaximumPoolSize(1);
+         addConnectionExecutor.setCorePoolSize(1);
       }
    }
 


johnou (approved):

quaff:
https://github.com/openjdk/jdk/blob/98c9f8bdc7ef45c5ec7bb1f16c34234a7cb3890f/src/java.base/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1536

johnou:
@brettwooldridge this looks like a blocker for those wishing to adopt jdk 11.

pentlander:
I don't think this is correct, `addConnectionExecutor.setCorePoolSize(1);` should still occur before `addConnectionExecutor.setMaximumPoolSize(1);` otherwise you still run into issues. I'm still getting an exception when running `4.0.1` this time on https://github.com/sw-pachanady/HikariCP/blob/9c52fc556094f4d1bc3835bcb54a50cea3d48e97/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L149

johnou:
@pentlander please provide the stracktrace.

pentlander:
Well logically once line 149 is reached, the core size and max size are at least 16. Once you try to shrink the max size to 1, [this line gets run](https://github.com/openjdk/jdk15/blob/19b8204719c50d70fa435f9484a61365258b46c3/src/java.base/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1661) and fails because the core size is still 16. I'll attach a stacktrace tomorrow morning.

pentlander:
Using Hikari 4.0.1 with OpenJDK 15, with "com.zaxxer.hikari.blockUntilFilled" set to true and initialization fail timeout at 10_000:
```
Uncaught exception in thread main
java.lang.IllegalArgumentException: null
	at java.base/java.util.concurrent.ThreadPoolExecutor.setMaximumPoolSize(Unknown Source)
	at com.zaxxer.hikari.pool.HikariPool.<init>(HikariPool.java:149)
	at com.zaxxer.hikari.HikariDataSource.<init>(HikariDataSource.java:81)
        ...
```

johnou:
wow looks like we need some conditional checks when setting the value to avoid the iae.

```
    public void setCorePoolSize(int corePoolSize) {
        if (corePoolSize < 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();

    public void setMaximumPoolSize(int maximumPoolSize) {
        if (maximumPoolSize <= 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();
        this.maximumPoolSize = maximumPoolSize;
```

pentlander:
It should be fine if 149 and 150 are flipped back to their original order, i.e.
```
addConnectionExecutor.setCorePoolSize(1);
addConnectionExecutor.setMaximumPoolSize(1);
```

brettwooldridge:
Wait. Are we saying that in JDK 11 we need to call ``setMaximumPoolSize()`` before ``setCorePoolSize()``, but in JDK 15 we need to call ``setCorePoolSize()`` before ``setMaximumPoolSize()``?

pentlander:
No, I'm saying that the new code increasing the pool size is correct [here](https://github.com/brettwooldridge/HikariCP/pull/1605/files#diff-980e81eb213376eb8e36ab6b3e3e1ff2f04c963b9b79264130f5f5dbbf03be83R141-R142) because it increases the max size to 16 before increasing the core size, avoiding this condition:
```
public void setCorePoolSize(int corePoolSize) {
        if (corePoolSize < 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();
```

The new code is incorrect [here](https://github.com/brettwooldridge/HikariCP/pull/1605/files#diff-980e81eb213376eb8e36ab6b3e3e1ff2f04c963b9b79264130f5f5dbbf03be83R149-R150) because it is reducing the max pool size to 1 while the core pool size is still 16, thus triggering:
```
public void setMaximumPoolSize(int maximumPoolSize) {
        if (maximumPoolSize <= 0 || maximumPoolSize < corePoolSize)
            throw new IllegalArgumentException();
```

If the latter code is reverted to this it should work, since the the core size will always be less than the max size:
```
addConnectionExecutor.setCorePoolSize(1);
addConnectionExecutor.setMaximumPoolSize(1);
```

I don't know why the author of the pull request didn't run into this issue, the JDK11 checks are the same as the JDK15 ones: https://github.com/openjdk/jdk11/blob/37115c8ea4aff13a8148ee2b8832b20888a5d880/src/java.base/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1658

brettwooldridge:
Fixed and released as v4.0.2.

pentlander:
> Fixed and released as v4.0.2.

I validated it's working with JDK15. Thanks!

PR #1665, opened 2020-10-05
Use 'requires static' for optional dependencies

nictas:
Attempting to compile a project that depends on HikariCP results in the following exceptions:
```
java.lang.module.FindException: Module micrometer.core not found, required by com.zaxxer.hikari
java.lang.module.FindException: Module simpleclient not found, required by com.zaxxer.hikari
java.lang.module.FindException: Module javassist not found, required by com.zaxxer.hikari
java.lang.module.FindException: Module hibernate.core not found, required by com.zaxxer.hikari
```
Since all of these are defined as `optional` in the `pom.xml` of HikariCP, it also makes sense to mark them as such in the `module-info.java` file.
See https://nipafx.dev/java-modules-optional-dependencies

### DIFF
--- a/src/main/java11/module-info.java
+++ b/src/main/java11/module-info.java
@@ -1,15 +1,16 @@
 module com.zaxxer.hikari
 {
-   requires hibernate.core;
    requires java.sql;
    requires java.management;
    requires java.naming;
-   requires javassist;
-   requires simpleclient;
    requires slf4j.api;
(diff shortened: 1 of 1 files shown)


nictas:
Never mind, I see that there's an already open PR for the same thing: https://github.com/brettwooldridge/HikariCP/pull/1587

PR #1669, opened 2020-10-11
Fixed optional dependencies that were marked as required in module-in…

cowwoc:
…fo.java


cowwoc:
@brettwooldridge Quick fix for you Brett. Can you please deploy a SNAPSHOT for me to test?

cowwoc:
Tested locally, this seems to work. Please include this in the upcoming release. Thanks.

cowwoc:
Okay, now we're blocked by https://stackoverflow.com/q/64309353/14731

Applications that use Java Modules must use slf4j 2.0.0-alpha1 but Hikari uses version 1.x. The two versions are incompatible.

But Hikari cannot use version 2.0.0-alpha1 because it depends on log4j2's slf4j bridge, and the latter is not compatible with slf4j 2.0.0-alpha: https://issues.apache.org/jira/browse/LOG4J2-2942

For what it's worth, Logback (version 1.3.0-alpha5) is compatible, but I assume that you prefer sticking to log4j2.

cowwoc:
Nevermind. I think I found a solution... I'll post a more detailed follow up soon.

cowwoc:
So back in April 2017 Ceki posted slf4j 1.8.0-alpha0. This went through 8 subsequent releases eventually ending with 2.0.0-alpha1. Version 2.0.0 seems to be backwards compatible with version 1.8.0 and per https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/ there is in fact a separate bridge for versions 1.8.0 and up.

In the above commit, I've updated the pom.xml file to use separate bridges for Java 8 and Java 11. This seems to resolve all outstanding problems.

PR #1681, opened 2020-11-02
Refine JMX ObjectName

### LINKED ISSUE #1678: JMX key property "type" is overly specific and property "name" is missing
When enabling JMX each Pool (and PoolConfig) is registered as `com.zaxxer.hikari:type=Pool (HikariPool-1)`. This makes it hard to select a single metric of all Pools and split the metric on the name of pool (since the name is part of the type).

It would be easier when the object name were `com.zaxxer.hikari:type=Pool,name=HikariPool-1` and it would match the behavior of other pools in JMX (e.g. `java.lang.MemoryPool`).

This issues becomes readily apparent when trying to write a JMX Extension for Dynatrace Managed.

quaff:
fix https://github.com/brettwooldridge/HikariCP/issues/1678

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -279,8 +279,14 @@ abstract class PoolBase
       try {
          final MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
 
-         final ObjectName beanConfigName = new ObjectName("com.zaxxer.hikari:type=PoolConfig (" + poolName + ")");
-         final ObjectName beanPoolName = new ObjectName("com.zaxxer.hikari:type=Pool (" + poolName + ")");
+         ObjectName beanConfigName, beanPoolName;
+         if ("true".equals(System.getProperty("hikaricp.jmx.register2.0"))) {
+             beanConfigName = new ObjectName("com.zaxxer.hikari:type=PoolConfig,name=" + poolName);
+             beanPoolName = new ObjectName("com.zaxxer.hikari:type=Pool,name=" + poolName);
+         } else {
+            beanConfigName = new ObjectName("com.zaxxer.hikari:type=PoolConfig (" + poolName + ")");
+            beanPoolName = new ObjectName("com.zaxxer.hikari:type=Pool (" + poolName + ")");
+         }
          if (register) {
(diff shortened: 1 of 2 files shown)


brettwooldridge:
@bo0ts I like this change, however, in its current form it will break existing users. I suggest enabling the new behavior via a system property, ``hikaricp.jmx.register2.0`` (eg. ``-Dhikaricp.jmx.register2.0=true``).

quaff:
> @bo0ts I like this change, however, in its current form it will break existing users. I suggest enabling the new behavior via a system property, `hikaricp.jmx.register2.0` (eg. `-Dhikaricp.jmx.register2.0=true`).

It's too conservative, I don't think it will break any critical function.

brettwooldridge:
@quaff Wouldn't this change break how lookups are performed?
```java
MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
ObjectName poolName = new ObjectName("com.zaxxer.hikari:type=Pool (foo)");
HikariPoolMXBean poolProxy = JMX.newMXBeanProxy(mBeanServer, poolName, HikariPoolMXBean.class);
```
Any existing code performing programatic lookups seems like it would surely be broken.

quaff:
> @quaff Wouldn't this change break how lookups are performed?
> 
> ```java
> MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
> ObjectName poolName = new ObjectName("com.zaxxer.hikari:type=Pool (foo)");
> HikariPoolMXBean poolProxy = JMX.newMXBeanProxy(mBeanServer, poolName, HikariPoolMXBean.class);
> ```
> 
> Any existing code performing programatic lookups seems like it would surely be broken.

I mean rare projects will lookup it directly by code.

brettwooldridge:
@quaff HikariCP has millions of users, literally. "Rare" in this case could be thousands of users.  One of those "users" could easily be Spring Boot, the Play framework, or Oracle Cloud, all of which use HikariCP and in turn have millions of users of their own.

quaff:
> @quaff HikariCP has millions of users, literally. "Rare" in this case could be thousands of users. One of those "users" could easily be Spring Boot, the Play framework, or Oracle Cloud, all of which use HikariCP and in turn have millions of users of their own.

Updated according to your suggestion.

PR #1682, opened 2020-11-02
Supports maven "-DskipTests"

quaff:
before this commit tests cannot be skipped by "-DskipTests"
see https://maven.apache.org/surefire/maven-surefire-plugin/examples/skipping-tests.html


brettwooldridge:
Duplicate of #1249. Now merged.

PR #1727, opened 2021-01-25
FIX #1726 : Regression in v4.0.0 : Micrometer's gauges return no values for PoolStats

### LINKED ISSUE #1726: Regression in v4.0.0 : Micrometer's gauges return no values for PoolStats
Since v4.0.0, Micrometer's gauge metrics return no value.

For example, when using a Prometheus implementation, we will get `NaN` : 

```
hikaricp_connections_active{application="my-demo-webapp",pool="HikariPool-1",} NaN
hikaricp_connections_idle{application="my-demo-webapp",pool="HikariPool-1",} NaN
```

I checked, there is no regression with `PrometheusMetricsTracker`, only with `MicrometerMetricsTracker`.

This kind of problem is referenced by the Micrometer documentation as : 
- [Why is my Gauge reporting NaN or disappearing?](https://micrometer.io/docs/concepts#_why_is_my_gauge_reporting_nan_or_disappearing)

At first sight, it seems to be related to this commit, which should be (partially) reverted in my opinion : https://github.com/brettwooldridge/HikariCP/commit/6e0f4a8d441bb293d61755baefc6544ab6efeec4

ghusta:


quaff:
@ghusta Is it fixed by https://github.com/brettwooldridge/HikariCP/commit/4b8d71b8e8eceadabc603bd8b78d6c170d327bb5 ?

ghusta:
@quaff It's slightly different from what I commited but it fixed the problem. It's OK in release 4.0.1

quaff:
> @quaff It's slightly different from what I commited but it fixed the problem. It's OK in release 4.0.1

Because `this.poolStats` is equals to `poolStats`, then you should close this now.

ghusta:
> > @quaff It's slightly different from what I commited but it fixed the problem. It's OK in release 4.0.1
> 
> Because `this.poolStats` is equals to `poolStats`, then you should close this now.

Yes, it's the same reference.

Indeed, I can close it myself.

ghusta:
Fixed by https://github.com/brettwooldridge/HikariCP/commit/4b8d71b8e8eceadabc603bd8b78d6c170d327bb5

PR #1782, opened 2021-05-19
Fix Typo Readme

nguyensach:


lfbayer:
The original english is actually correct. Changing it to "it will" is actually incorrect.

lfbayer:
https://www.thoughtco.com/inversion-definition-1209968

PR #1825, opened 2021-08-10
Allow configuring explicit driver class loader

timja:
For more controlled classloading than was added in https://github.com/brettwooldridge/HikariCP/pull/881

I'm trying to use this to replace `commons-dbcp2` in Jenkins, and neither of the classloading options that are available here work (docs: https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/)

Similar issue:
https://github.com/FasterXML/jackson-dataformat-xml/issues/483#issuecomment-873723152

I've tested this in Jenkins and it works.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -79,6 +79,7 @@ public class HikariConfig implements HikariConfigMXBean
    private String dataSourceClassName;
    private String dataSourceJndiName;
    private String driverClassName;
+   private ClassLoader driverClassLoader;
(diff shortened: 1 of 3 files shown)


brettwooldridge:
I don't like this change. First, any change to class signatures requires a major revision increment. Second, the usual pattern for adapting to a classloading framework is to switch the context classloader and then switch back.
```java
var originalLoader = Thread.currentThread().getContextClassLoader();
try {
   Thread.currentThread().setContextClassLoader(myClassLoader);

   // setup
   return new HikariDataSource(config);
} finally {
   Thread.currentThread().setContextClassLoader(originalLoader);
}
```

PR #1842, opened 2021-09-07
Don't clear isCommitStateDirty flag in setReadOnly

dittos:
When autoCommit is false, if `setReadOnly` is called on a connection after executing some queries, closing the connection returns the dirty connection (with transaction open) to the pool.

```java
HikariDataSource ds = /* setup the pool */;
ds.setAutoCommit(false);
try (Connection conn = ds.getConnection()) {
  // ... execute some queries on `conn`
  conn.setReadOnly(false);
}
// connection returned to the pool with transaction open
```

In the above scenario, the following cleanup logic is not executed.

https://github.com/brettwooldridge/HikariCP/blob/ed2da5f1f4ef19f871fac12effc0b199706905dc/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L248-L251

Anyway I don't see why `isCommitStateDirty` should be set to false when `setReadOnly` is called. The line was added in 6ed3b185912bd835fb1a25da6ba80e0226fb7f2e without useful description.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -409,7 +409,6 @@ public abstract class ProxyConnection implements Connection
    {
       delegate.setReadOnly(readOnly);
       isReadOnly = readOnly;
-      isCommitStateDirty = false;
       dirtyBits |= DIRTY_BIT_READONLY;
    }
 
--- a/src/test/java/com/zaxxer/hikari/pool/ConnectionStateTest.java
+++ b/src/test/java/com/zaxxer/hikari/pool/ConnectionStateTest.java
@@ -164,6 +164,9 @@ public class ConnectionStateTest
 
             resultSet.updateRow();
             assertTrue(TestElf.getConnectionCommitDirtyState(connection));
+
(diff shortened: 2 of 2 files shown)


brettwooldridge:
Convince me this isn't user error. Why is ``setReadOnly()`` being called in the middle of an outstanding transaction?

dittos:
Of course it is user error. I think it would be best if exception is thrown on invalid `setReadOnly()` call.
But some JDBC drivers does not throw when `setReadOnly` is called during transaction if the flag is same as before.

I would expect transactions to be rolled back when closing the connection in any case.

brettwooldridge:
Well, I'm definitely not going to throw an exception, that's up the the driver. What concerns me about this fix is that it is a direct rollback of a a change submitted by another user many years ago (6ed3b185).

Granted this was a batch of fixes in a single commit called "small but important fixes" and he was just combing through the code looking for what he considered "correctness" fixes. The only concern I have is the potential change in behavior and whether it is likely to break existing code. I'm willing to put the change out there, but if breakage reports flood in, it will have to be reverted. You ok with that?

dittos:
I don't need the fix now because I found it in my previous job. (did apply custom build with the patch there, and it worked without problem though)

If you think applying the change is risky, feel free to close the PR! 😄

PR #1868, opened 2021-10-31
add mission groupId in pom.xml.

wanglunhui2012:


siepkes:
The group ID `org.apache.maven.plugins` is the default group ID in Maven. Therefor there is no need to specify it. See the Maven POM XSD: https://maven.apache.org/xsd/maven-4.0.0.xsd

wanglunhui2012:
> The group ID `org.apache.maven.plugins` is the default group ID in Maven. Therefor there is no need to specify it. See the Maven POM XSD: https://maven.apache.org/xsd/maven-4.0.0.xsd

But this will prompt error in my IDEA.

siepkes:
In that case you could file an issue with your IDE vendor. Because as you can see in the [Maven POM XSD](https://maven.apache.org/xsd/maven-4.0.0.xsd) `org.apache.maven.plugins` is the default group ID. Thats also why running `mvn package` works.

Issue #1889, opened 2021-12-16
Validation of closed connection causes logging of "Failed to validate connection" warning


pvinokurov:
In the current implementation of PoolBase#isConnectionDead(isConnectionAlive) setNetworkTimeout() is called firstly to check the connection validity:
[PoolBase#L151](https://github.com/brettwooldridge/HikariCP/blob/ed2da5f1f4ef19f871fac12effc0b199706905dc/src/main/java/com/zaxxer/hikari/pool/PoolBase.java#L151)

In many JDBC drivers, calling **Connection#setNetworkTimeout** throws SqlException in the case when a connection is closed.
For example, Postgres JDBC driver throws PSQLException("This connection has been closed.") by calling setNetworkTimeout()
[PgConnection#L1622](https://github.com/pgjdbc/pgjdbc/blob/9899c6fb4646ad032f83cec99be49180eab802b2/pgjdbc/src/main/java/org/postgresql/jdbc/PgConnection.java#L1622)
[PgConnection.java#L883](https://github.com/pgjdbc/pgjdbc/blob/9899c6fb4646ad032f83cec99be49180eab802b2/pgjdbc/src/main/java/org/postgresql/jdbc/PgConnection.java#L883)

The behavior of isConnectionDead() method looks odd because in the case of apriori closed connection it causes an exception and logging warning.
If a connection is not valid by the result of Connection.isValid method the warning is not logged.
The warning message could be confusing due to the indirect relation between maxLifeTime parameter and closed connections.

The proposal is to check that connection is closed before calling setNetworkTimeout thereby not causing the exception and logging the warning.

pvinokurov:
https://github.com/brettwooldridge/HikariCP/pull/1890

pvinokurov:
Duplicated https://github.com/brettwooldridge/HikariCP/pull/1890

ribalw:
Hi @pvinokurov, Is this issue still open and has not been sorted? We are facing this quite often in all of our services whereas all our metrics look good. We just have tons of warnings that we want to be removed.

PR #1890, opened 2021-12-17
Check connection has not been closed at first while checking the validity of connection

pvinokurov:
In the current implementation of PoolBase#isConnectionDead(isConnectionAlive) setNetworkTimeout() is called firstly to check the connection validity:
PoolBase#L151

In many JDBC drivers, calling Connection#setNetworkTimeout throws SqlException in the case when a connection is closed.
For example, Postgres JDBC driver throws PSQLException("This connection has been closed.") by calling setNetworkTimeout()
PgConnection#L1622
PgConnection.java#L883

The behavior of isConnectionDead() method looks odd because in the case of apriori closed connection it causes an exception and logging warning.
If a connection is not valid by the result of Connection.isValid method the warning is not logged.
The warning message could be confusing due to the indirect relation between maxLifeTime parameter and closed connections.

The proposal is to check that connection is closed before calling setNetworkTimeout thereby not causing the exception and logging the warning.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -147,6 +147,10 @@ abstract class PoolBase
    boolean isConnectionDead(final Connection connection)
    {
       try {
+         if(connection.isClosed()){
+            return false;
+         }
+
          try {
             setNetworkTimeout(connection, validationTimeout);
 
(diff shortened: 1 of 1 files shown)


pvinokurov:
Related to https://github.com/brettwooldridge/HikariCP/issues/1889

lfbayer:
Brett has explained the reason why not to solve this problem in this way in an earlier comment here: https://github.com/brettwooldridge/HikariCP/issues/1212#issuecomment-417404421

PR #1902, opened 2022-02-16
Remove the Java11 maven profile, which is no longer necessary now tha…

PascalSchumacher:
…t HikariCP requires Java 11.


brettwooldridge:
It was removed by another commit.

PR #1940, opened 2022-05-26
Expose whether pool is suspended in MXBean

ashok-square:


lfbayer:
@ashok-square sorry to take so long to respond to this pull request.

Can you add a unit test to cover this new method?

Issue #1950, opened 2022-07-13
Hikari support for TRANSACTION_LAST_COMMITED and TRANSACTION_CURSOR_STABILITY


liamsorsby:
Using the Informix jdbc plugin we have a use case where we need to use TRANSACTION_LAST_COMMITED however that doesn't look like it's support in the enum
https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/util/IsolationLevel.java#L19
Documentation of the allowed Isolation Levels here: https://www.ibm.com/docs/en/informix-servers/12.10?topic=database-informix-environment-variables-informix-jdbc-driver

The map to the following:
```
0 Equivalent to TRANSACTION_NONE
1 Dirty Read (equivalent to TRANSACTION_READ_UNCOMMITTED),
2 Committed Read (equivalent to TRANSACTION_READ_COMMITTED),
3 Cursor Stability (equivalent to TRANSACTION_READ_COMMITTED),
4 Repeatable Read (equivalent to TRANSACTION_REPEATABLE_READ)
5 Committed Read LAST COMMITTED (equivalent to TRANSACTION_LAST_COMMITTED)
8 Equivalent to TRANSACTION_SERIALIZABLE
```

Would there be any objection to implementing the missing ENUM's or was there a reason they weren't added?
```
3 Cursor Stability (equivalent to TRANSACTION_READ_COMMITTED),
5 Committed Read LAST COMMITTED (equivalent to TRANSACTION_LAST_COMMITTED)
```

PR #1952, opened 2022-07-14
[#1950] Adds support for TRANSACTION_CURSOR_STABILITY and TRANSACTION_LAST_COMMITTED

liamsorsby:
TRANSACTION_LAST_COMMITTED - This allows a readonly user to access a row and read the last commited data while Informix has an open lock against the data.
Unlike TRANSACTION_READ_UNCOMMITTED which would also accomplish the same but would end up allowing dirty reads. This would lead to us reading data that's potentially either rolledback or never actually commited.

TRANSACTION_CURSOR_STABILITY - Cursor stability will place a lock on a row as it's fetched which ensures that the data doesn't change while the program reads the data.
Documentation for cursor stability can be found here https://www.ibm.com/docs/en/informix-servers/12.10?topic=level-informix-cursor-stability-isolation

Documentation on Informix transaction isolation levels are here https://www.ibm.com/docs/en/informix-servers/12.10?topic=database-informix-environment-variables-informix-jdbc-driver


lfbayer:
@brettwooldridge This seems reasonable to me, but presumably there are other isolation levels as well. Why stop here, is there a way to support everything without HikariCP having to know?

liamsorsby:
As the Ifxconnection extends java.sql.connection and I suspect most of them would. It might be possible to just expect an instance of java.sql.connection to be passed in. That would likely result in a breaking change though.

https://javadoc.io/static/com.ibm.informix/jdbc/4.10.9/com/informix/jdbc/IfxConnection.html

PR #1959, opened 2022-07-26
Add GraalVM native-image configuration

mhalbritter:
This allows HikariCP to be used in a GraalVM native image without additional configuration.

The JSON configuration files have been generated by running 'run-tests-with-agent.sh', which executes the test suite with the native-image-agent attached and then generates the configuration metadata located in src/main/resources.

References:

* https://www.graalvm.org/22.1/reference-manual/native-image/Agent/
* https://www.graalvm.org/22.1/reference-manual/native-image/Reflection/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -250,6 +250,9 @@
          <resource>
             <directory>target/classes</directory>
          </resource>
+         <resource>
+            <directory>src/main/resources</directory>
+         </resource>
       </resources>
 
       <plugins>
@@ -653,6 +656,31 @@
                <scope>test</scope>
             </dependency>
          </dependencies>
+      </profile>
+
+      <profile>
+         <id>native</id>
+         <build>
+            <plugins>
+               <plugin>
+                  <groupId>org.graalvm.buildtools</groupId>
+                  <artifactId>native-maven-plugin</artifactId>
+                  <version>0.9.13</version>
+                  <extensions>true</extensions>
+                  <configuration>
+                     <agent>
+                        <options>
+                           <option>experimental-conditional-config-part</option>
+                        </options>
+                        <options name="test">
+                           <option>access-filter-file=${project.basedir}/native/access-filter.json</option>
+                           <option>caller-filter-file=${project.basedir}/native/caller-filter.json</option>
+                        </options>
+                     </agent>
+                  </configuration>
+               </plugin>
+            </plugins>
+         </build>
       </profile>
    </profiles>
 </project>
--- a/native/access-filter.json
+++ b/native/access-filter.json
@@ -1,0 +1,10 @@
+{
+  "rules": [
+    {
+      "includeClasses": "**"
+    },
+    {
+      "excludeClasses": "org.apache.logging.log4j.**"
+    }
+  ]
+}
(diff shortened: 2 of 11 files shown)


mhalbritter:
Not sure why the codecov is red, i didn't touch any code.

mhalbritter:
Hello! Any chance that this will be merged soon?

lfbayer:
Forgive my ignorance, but is it normal for the library providers to have to make changes to support GraalVM?

mhalbritter:
Thanks for asking! There are 3 possibilities for libraries:

1. ignore GraalVM and let users deal with a workaround
2. don't include GraalVM metadata and let users contribute to the [shared GraalVM reachability metadata](https://github.com/oracle/graalvm-reachability-metadata) repository. Once the metadata files are in there, all users can benefit from them
3. include the metadata files in their own source tree

It's up to you to decide how you want to handle this. GraalVM proponents would prefer option 3, but option 2. would work too. I've already contributed hints [here](https://github.com/oracle/graalvm-reachability-metadata/pull/16). I prefer the hints closer to the code which does the proxying / reflection etc. and that's why I have created the PR. But don't feel bad declining the PR if you don't want to have this metadata in your code.

lfbayer:
@mhalbritter, as a best practice generated code should generally not be committed. Is it standard practice for projects to commit all of these files directly? Is there not some maven plugin that can generate these automatically at build time?

It sounds like these files would need to be regenerated with any future code changes and it is unlikely that the maintainers would know what kind of changes would impact GraalVM metadata. I suspect this would mean someone will eventually and unwittingly cause the GraalVM metadata to be wrong.

I would prefer a solution similar to how the apache felix maven plugin generates the OSGi manifest. Felix is just a simple plugin in the pom.xml and everything is automatic.

If there is not a simpler way to support this, I would recommend your "option 2" until such a time as the GraalVM ecosystem can support a less invasive method.

mhalbritter:
These files have been originally created by running the test suite with the GraalVM agent attached. These files are only correct if the test suite has 100% coverage. The idea behind putting the GraalVM hints into the codebase is that the authors can update the hints files when changing the Java code (for example if more reflection is done). You can't easily auto-generate these GraalVM hints like you can with, for example, an OSGi manifest.

pom.xml
```diff
@@ -654,5 +657,30 @@
...
+                  <configuration>
+                     <agent>
+                        <options>
+                           <option>experimental-conditional-config-part</option>
+                        </options>
+                        <options name="test">
+                           <option>access-filter-file=${project.basedir}/native/access-filter.json</option>
+                           <option>caller-filter-file=${project.basedir}/native/caller-filter.json</option>
+                        </options>
+                     </agent>
```
linghengqian, line 679: - As a reminder, you have not configured `native/user-code-filter.json`

linghengqian: - Sorry, I noticed it in the shell script

brettwooldridge:
I would prefer the option above, "don't include GraalVM metadata and let users contribute to the [shared GraalVM reachability metadata](https://github.com/oracle/graalvm-reachability-metadata) repository. Once the metadata files are in there, all users can benefit from them." This looks pretty gross and adds to release-time headaches.

PR #1979, opened 2022-09-27
Don't log SSL truststore password

msteiger:
Currently, the keystore passwords are logged when the log level is set to debug for `HikariConfig`. The passwords should be masked though.

* The PostgreSQL driver uses `sslpassword` (https://jdbc.postgresql.org/documentation/use/) 
* MariaDB uses `trustStorePassword` (https://mariadb.com/kb/en/using-tls-ssl-with-mariadb-java-connector/) 

for client-side SSL truststores in JDBC URLs. These passwords should not be logged when DEBUG logging is enabled.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -1118,7 +1118,7 @@ public class HikariConfig implements HikariConfigMXBean
                value = "internal";
             }
             else if (prop.contains("jdbcUrl") && value instanceof String) {
-               value = ((String)value).replaceAll("([?&;]password=)[^&#;]*(.*)", "$1<masked>$2");
+               value = ((String)value).replaceAll("([?&;][^&#;=]*[pP]assword=)[^&#;]*", "$1<masked>");
             }
             else if (prop.contains("password")) {
                value = "<masked>";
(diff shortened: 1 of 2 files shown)


src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -1119,6 +1119,8 @@ else if (prop.matches("scheduledExecutorService|threadFactory") && value == null
             }
             else if (prop.contains("jdbcUrl") && value instanceof String) {
                value = ((String)value).replaceAll("([?&;]password=)[^&#;]*(.*)", "$1<masked>$2");
+               value = ((String)value).replaceAll("([?&;]sslpassword=)[^&#;]*(.*)", "$1<masked>$2");
+               value = ((String)value).replaceAll("([?&;]trustStorePassword=)[^&#;]*(.*)", "$1<masked>$2");
```
lfbayer: This change seems reasonable. The url properties are driver dependent so it is conceivable that there are other passwords as well. I wonder if it would be more future-proof to match any property that ends with "assword". What do you think?

Maybe a single replace all with something like this:
```java
value = ((String)value).replaceAll("([?&;][^&#;=]*assword=)[^&#;]*(.*)", "$1<masked>$2");
```

msteiger: Yeah, almost .. I've updated my PR and added some tests to make sure that also corner cases are covered.

PR #1992, opened 2022-10-18
Add uuid to identify connection timings

chillosu:
For our use of HikariCP we have determined that the average duration of time between connection creation attempt and connection established can be an indicator for determining health between the application layer and the database.

AFAIK, there is no way current way to correlate an attempt at establishing a connection and the existing logging "Added connection", this is an attempt to change that.


akashchouhan16 (commented):
@chillosu great addon for logs ✅

PR #2010, opened 2022-12-02
adding db url in the connection exception

peeyush18:
Hi @brettwooldridge

I am adding some extra data to the connection exception.

During getting a connection it is becomes hard to debug and troubleshoot. In case of multiple db servers(replicas) the change will help in isolating the server/s stalling while connecting


magdel:
The data is sensitive and it's unsafe to let it travel across application levels

brettwooldridge:
There is no scenario in which we will call ``poolEntry.connection.getMetaData()`` in the ``getConnection()`` path. Additionally, a JDBC URL is optional, not required.

EDIT: Not to mention that ``poolEntry.connection.getMetaData().getURL()`` only works if we actually acquired a ``Connection``, thus ``dbUrl`` is likely to uninitialized when the ``SQLException`` would be thrown.

PR #2011, opened 2022-12-03
Add connectionCreatorClassName config property (for dynamic passwords)

lfbayer:
add connectionCreatorClassName property to allow users to supply their own logic for creating connections

This is intended to be used for handling situation that require dynamic username/password combinations. For example for systems that require one-time-use passwords. The implementer could lookup or generate the password at the instant that the getConnection method would be called on the DataSource.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConnectionCreator.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConnectionCreator.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (C) 2022 Brett Wooldridge
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.zaxxer.hikari;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import javax.sql.DataSource;
+
+/**
+ * Users can implement this interface to override the default connection creation
+ * procedure.
+ * <p>
+ * Implementors typically determine how the username and password are chosen for
+ * the new connection. For example, the default implementation uses the username
+ * and password from {@link HikariConfig#getUsername()} and
+ * {@link HikariConfig#getPassword()} and passes them directly to
+ * {@link DataSource#getConnection(String, String)} to create the connection.
+ */
+public interface HikariConnectionCreator
+{
+   /**
+    *This method is called when a new connection needs to be added to the pool. One
+    * of {@link DataSource#getConnection(String, String)} or
+    * {@link DataSource#getConnection()} are expected to be called by implementors.
+    *
+    * @param dataSource The {@link DataSource} to use for getting a connection.
+    * @param config The current {@link HikariConfig}
+    * @return A newly created {@link Connection}
+    * @throws SQLException if a database access error occurs
+    */
+   Connection createConnection(DataSource dataSource, HikariConfig config) throws SQLException;
+}
--- a/README.md
+++ b/README.md
@@ -361,6 +361,14 @@ be used for various internally scheduled tasks.  If supplying HikariCP with a ``
 instance, it is recommended that ``setRemoveOnCancelPolicy(true)`` is used.
 *Default: none*
 
+&#128292;``connectionCreatorClassName``<br/>
+HikariCP will by default call the ``DataSource`` ``getConnection`` method directly to create new
+connections. This property allows you to supply your own instance of
+``com.zaxxer.hikari.HikariConnectionCreator`` that will be called any time a new connection is required.
+This is useful for dynamically choosing a username and/or password rather than relying on the
+``username`` and ``password`` properties on ``HikariConfig``.
+*Default: none*
+
 ----------------------------------------------------
 
 #### Missing Knobs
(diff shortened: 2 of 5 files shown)


lfbayer:
There have been various suggestions for how to provide users the ability to handle one-time-use-passwords or other variable password situations. 

My initial thought upon hearing the need for a password supplier was, why don't users just subclass or wrap their desired DataSource. Then they can do whatever they want. But it sounds like there are issues. For subclassing, it is possible that the DataSource class is final or fragile in someway that subclassing breaks how things work. And for wrapping the DataSource, there is a potential that frameworks expect the DataSource to be an instance of a specific class, so wrapping the DataSource would break those use-cases.

One common proposal has been to provide a simple supplier for the password. But there are cases that require both the username and password to be chosen. In that situation, it could be necessary that the retrieval of the username and password be atomic. For handling both username and password we would need either two suppliers or an interface that can return the username and password together as a single return value.

In my mind these options are unnecessarily complicated. So instead I decided it would just be easier to allow users to provide a factory class for creating the connection in a more generic way that isn't specifically for selecting passwords. The implementers can decide however they want to create the connection.

I would expect a typical implementation would just say something like:
```java
public Connection createConnection(DataSource dataSource, HikariConfig config) throws SQLException
{
   // get username and password from somewhere
   String username = ...
   String password = ...

   return dataSource.getConnection(username, password);
}
```

One potential issue I can see with my simple implementation is that the ConnectionCreator class is provided by classname, which means that there isn't a simple way for an IoC container to inject potentially necessary information into the creator instance itself. Maybe the connectionCreator should only be programmatically configured thus allowing the user to instantiate the instance themselves (eg: via an IoC framework). But this would preclude a simple properties based configuration from being able to use the functionality.

I am interested in hearing suggestions for improvements to this implementation.

xiaomaiyun:
When will this feature be supported?

brotheroftux:
>One potential issue I can see with my simple implementation is that the ConnectionCreator class is provided by classname

I'm a bit confused by this. I understand that maybe it's easier this way to configure Hikari via a `.properties` file, but as you've mentioned, this creates a bunch of issues when trying to make this work with an IoC/DI framework. As of right now, the only way to inject something into a `ConnectionCreator` class would be to ditch the DI pattern altogether and just use some global singleton.

Also, I have my concerns about connection soft eviction. The thing about dynamic credentials is we want to make sure that when there's a lease renewal, there's no potential downtime or issues caused by using the now outdated credentials.
I've seen [this](https://secrets-as-a-service.com/posts/hashicorp-vault/rotate-dynamic-relational-database-connection-in-spring-at-runtime/) article, where they do `softEvictConnections` every time the credentials are changed. I might be wrong, but, in my opinion, it might be Hikari's responsibility to ensure that connections in the pool are relevant enough to be used.

lfbayer:
I'm confused by what you are confused by. Are you agreeing or disagreeing with my concern regarding specifying the classname rather than specifying an instance?

As far as the staleness of the connection. Can you give me an example of a situation where the maxLifetime of a connection is not sufficient to define how long a connection can be good for? If we are talking about a one-time use password, then the password is only valid once anyway and the valid duration of the password is not related to how long the connection would then be good for.

I don't really understand the use-case of evicting existing connections because of the connection password changing. I feel like for the use-case in the article you linked that so long as every new connection uses the new credentials everything would be fine. Am I wrong? And if so does the implementation in this pull request prevent you from doing what you want to do. Or is there some other additional item that is necessary?

brotheroftux:
> I'm confused by what you are confused by. Are you agreeing or disagreeing with my concern regarding specifying the classname rather than specifying an instance?
> 
> As far as the staleness of the connection. Can you give me an example of a situation where the maxLifetime of a connection is not sufficient to define how long a connection can be good for? If we are talking about a one-time use password, then the password is only valid once anyway and the valid duration of the password is not related to how long the connection would then be good for.
> 
> I don't really understand the use-case of evicting existing connections because of the connection password changing. I feel like for the use-case in the article you linked that so long as every new connection uses the new credentials everything would be fine. Am I wrong? And if so does the implementation in this pull request prevent you from doing what you want to do. Or is there some other additional item that is necessary?

I'm sorry if I didn't make myself clear enough. Yes, I was agreeing with your concern — I think it would be hard to customize/inject dependencies to a `ConnectionCreator` in this specific implementation of the API.

Regarding connection lifetime/eviction — in Hashicorp Vault, the creds are not exactly for a single use, but rather for use during some lease duration, which can be extended up to a certain max lease. This potentially makes a system that utilizes a connection pool and such dynamic credentials prone to failure when the lease expires.

mrotteveel:
What does this add over using a custom `javax.sql.DataSource` on the `dataSource` or `dataSourceClassName` property?

src/main/java/com/zaxxer/hikari/HikariConnectionCreator.java
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+package com.zaxxer.hikari;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import javax.sql.DataSource;
+
+/**
```
awakelee, line 23: Tell users that the param `datasource` is the underlying DataSource instance(implemented by JDBC Driver), which created by HikariDataSource. 
Because HikariDataSource#getConnection(Strinig, String) is not implemented:
```java
   @Override
   public Connection getConnection(String username, String password) throws SQLException
   {
      throw new SQLFeatureNotSupportedException();
   }
```

lfbayer:
> What does this add over using a custom `javax.sql.DataSource` on the `dataSource` or `dataSourceClassName` property?

@mrotteveel I also fundamentally believe that all use cases that people have been asking about can be handled exactly as you explain. But the one limitation that seemed to prevent people from doing this was that in their systems that there was some reason why the DataSource instance had to be the specific underlying DataSource otherwise various libraries would not work.

I'm actually tempted to mark every issue and pull request related to this feature request as closed and just not do anything because doing exactly what you suggest seems to me that it would cover all the use-cases. And if there is an edge case where the exact instance matters, then I would think the feature request should be directed to that other library rather than the connection pool. HikariCP's fundamental philosophy is to not provide any more functionality than just a connection pool.

nigelwu329:
@lfbayer I'd like to use this feature in our projects. Do you have any expected time to release this feature?

Issue #1960, opened 2022-08-01
Excetion override does not work outside the predefined error codes / states


markwimpory:
Right now you cannot create an exception override when the sql state or error code falls outside the predefined values (that cannot be overridden). Suggestion in ProxyConnection:

         if (sqlState != null && sqlState.startsWith("08")
             || nse instanceof SQLTimeoutException
             || ERROR_STATES.contains(sqlState)
             || ERROR_CODES.contains(nse.getErrorCode())) {

            // broken connection
            evict = true;
            break;
         } else if (exceptionOverride != null && exceptionOverride.adjudicate(nse) == CONTINUE_EVICT) {
            evict = true;
            break;
         } else {
            nse = nse.getNextException();
         }

For example we are dealing with SQLState 25006 when database failover occurs and the read/write replicas are swapped.

davemorrissey:
I'm having the same issue. Setting a short maxLifetime eventually results in these connections being recycled, but with a performance trade off. Allowing addition of codes to the ERROR_STATES set may be an alternative solution.

patric-r:
+1

brettwooldridge:
fixed by c6d1403e

PR #2031, opened 2023-01-19
Update checkException to adjudicate various exception

mokamoto12:
I've addressed the issue https://github.com/brettwooldridge/HikariCP/issues/1960
Best regards.


brettwooldridge:
What does this do that ``adjudicate(final SQLException sqlException)`` doesn't handle? It doesn't look like it does anything.

brettwooldridge:
fixed by c6d1403e

Issue #2004, opened 2022-11-14
Allow for a configurable variance on maxLifetime timeouts


ssherwood:
Currently, HikariCP adds a 2.5% variance on the maxLifetime to reduce the impact of max eviction of the connection pool:

```java
// variance up to 2.5% of the maxlifetime
final var variance = maxLifetime > 10_000 ? ThreadLocalRandom.current().nextLong( maxLifetime / 40 ) : 0;
```

It would be advantageous to have this be configurable and not specifically proportional to the `maxLifetime` duration itself.

Avarga954:
I would absolutely love this functionality I just came here to create this issue myself. Ideally a configurable percentage.

Utilizing both RDS/Aurora DB in conjunction with HikariDB I am regularly getting issues with the mass extinction of connections and reopening which generally causes issues with load balancing on the DB side by overloading replicas.

Aurora DB reader endpoints utilize a[ round robin DNS based load balancing with a TTL ](https://medium.com/talabat-tech/solving-aws-aurora-rds-load-balancing-problem-with-pgpool-ii-d54897c95176) which can end up getting any new connection request routed within a small time frame to the same replica.

Utilizing RDS Proxy is supposed to mitigate this exact issue but I am still seeing similar problems due to the mass extinction of connections even when routing through RDS proxy.

This issue can be exacerbated when deploying your applications on Kubernetes when multiple pods can stand up simultaneously within a small window of time. 

Allowing a configurable jitter/variance would dramatically improve the stability of our applications by allowing us to tune them explicitly. The only knob we have to currently adjust is the "MaxLifeTime" setting which will increase the default variance but that is not ideal since if new DB replicas are added to the connection pool it can delay their utilization by HikariCP.

Other tickets that I have been following.
https://github.com/brettwooldridge/HikariCP/issues/1247
https://github.com/brettwooldridge/HikariCP/pull/480

@brettwooldridge  I would love to hear your thoughts on this, I have taken this issue up with AWS technical support but I think the only way of resolving this issue is going to be at the HikariCP level.

Avarga954:
Just to follow up here,

I went ahead and took the initiative to spin up a PR with this desired change.

https://github.com/brettwooldridge/HikariCP/pull/2035

ssherwood:
I'm going to close this requests as it is effectively resolved here: https://github.com/brettwooldridge/HikariCP/pull/2035

PR #2035, opened 2023-02-01
Configurable MaxLifeTime Variance

Avarga954:
Refactoring MaxLifeTime connection variance to utilize a configurable setting rather than a static 2.5%.
2.5% will remain the default however developers can choose to override this for a more aggressive percentage if desired.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -53,7 +53,7 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long SOFT_TIMEOUT_FLOOR = Long.getLong("com.zaxxer.hikari.timeoutMs.floor", 250L);
    private static final long IDLE_TIMEOUT = MINUTES.toMillis(10);
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
-   private static final double MAX_LIFETIME_VARIANCE = 2.5;
+   private static final double DEFAULT_MAX_LIFETIME_VARIANCE = 2.5;
    private static final long DEFAULT_KEEPALIVE_TIME = 0L;
    private static final int DEFAULT_POOL_SIZE = 10;
 
@@ -121,7 +121,7 @@ public class HikariConfig implements HikariConfigMXBean
       minIdle = -1;
       maxPoolSize = -1;
       maxLifetime = MAX_LIFETIME;
-      maxLifetimeVariance = MAX_LIFETIME_VARIANCE;
+      maxLifetimeVariance = DEFAULT_MAX_LIFETIME_VARIANCE;
       connectionTimeout = CONNECTION_TIMEOUT;
       validationTimeout = VALIDATION_TIMEOUT;
       idleTimeout = IDLE_TIMEOUT;
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -53,6 +53,7 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long SOFT_TIMEOUT_FLOOR = Long.getLong("com.zaxxer.hikari.timeoutMs.floor", 250L);
    private static final long IDLE_TIMEOUT = MINUTES.toMillis(10);
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
+   private static final double MAX_LIFETIME_VARIANCE = 2.5;
    private static final long DEFAULT_KEEPALIVE_TIME = 0L;
    private static final int DEFAULT_POOL_SIZE = 10;
 
@@ -66,6 +67,7 @@ public class HikariConfig implements HikariConfigMXBean
    private volatile long idleTimeout;
    private volatile long leakDetectionThreshold;
    private volatile long maxLifetime;
+   private volatile double maxLifetimeVariance;
    private volatile int maxPoolSize;
    private volatile int minIdle;
    private volatile String username;
@@ -119,6 +121,7 @@ public class HikariConfig implements HikariConfigMXBean
       minIdle = -1;
       maxPoolSize = -1;
       maxLifetime = MAX_LIFETIME;
+      maxLifetimeVariance = MAX_LIFETIME_VARIANCE;
       connectionTimeout = CONNECTION_TIMEOUT;
       validationTimeout = VALIDATION_TIMEOUT;
       idleTimeout = IDLE_TIMEOUT;
@@ -228,6 +231,17 @@ public class HikariConfig implements HikariConfigMXBean
    {
       this.leakDetectionThreshold = leakDetectionThresholdMs;
    }
+
+   /** {@inheritDoc} */
+   @Override
+   public double getMaxLifetimeVariance()
+   {
+      return maxLifetimeVariance;
+   }
+
+   /** {@inheritDoc} */
+   @Override
+   public void setMaxLifetimeVariance(double maxLifetimeVariance) { this.maxLifetimeVariance = maxLifetimeVariance; }
 
    /** {@inheritDoc} */
    @Override
--- a/src/main/java/com/zaxxer/hikari/HikariConfigMXBean.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfigMXBean.java
@@ -94,6 +94,24 @@ public interface HikariConfigMXBean
    void setLeakDetectionThreshold(long leakDetectionThresholdMs);
 
    /**
+    * This property controls the percentage of a connections maximum lifetime that will be used as variance/jitter.
+    * This percentage will be a maximum and random variances up to this percentage will be subtracted from the defined
+    * maxLifeTime value.
+    *
+    * @return the maximum percentage of a connections maxLifeTime to be utilized as a variance
+    */
+   double getMaxLifetimeVariance();
+
+   /**
+    * This property controls the percentage of a connections maximum lifetime that will be used as variance/jitter.
+    * This percentage will be a maximum and random variances up to this percentage will be subtracted from the defined
+    * maxLifeTime value.
+    *
+    * @param maxLifetimeVariance the maximum percentage of maxLifeTime that should be used for variance
+    */
+   void setMaxLifetimeVariance(double maxLifetimeVariance);
+
+   /**
     * This property controls the maximum lifetime of a connection in the pool. When a connection reaches this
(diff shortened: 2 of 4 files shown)


Avarga954:
Just wanted to leave a comment linking back to the open issue where I discussed the problem a little bit more at length.

https://github.com/brettwooldridge/HikariCP/issues/2004

Avarga954:
Here is an example of the default hikari variance that has a maximum of 2.5% variance
Screenshot 2023-02-10 at 1 20 17 PM

Here is an example with a much more aggresive maximum variance of 30%
Screenshot 2023-02-10 at 1 20 37 PM


The large spikes / mass closing of connections is much more smoothed out.

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -53,6 +53,7 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long SOFT_TIMEOUT_FLOOR = Long.getLong("com.zaxxer.hikari.timeoutMs.floor", 250L);
    private static final long IDLE_TIMEOUT = MINUTES.toMillis(10);
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
+   private static final double MAX_LIFETIME_VARIANCE = 2.5;
```
bruce-ricard: Maybe this constant should be called `DEFAULT_MAX_LIFETIME_VARIANCE` instead? As it appears to be the default value.

Avarga954: Great suggestion, went ahead and changed it.

bruce-ricard: I don't want to give you false expectations that your PR will be merged soon. I don't have push permissions. I'm just an open source engineer happening to look at this PR, because I want to see how reactive the maintainers are.

bruce-ricard: And thank you for your kind response to my review. I wish more developers appreciated code review suggestions :).

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -66,6 +67,7 @@ public class HikariConfig implements HikariConfigMXBean
    private volatile long idleTimeout;
    private volatile long leakDetectionThreshold;
    private volatile long maxLifetime;
+   private volatile double maxLifetimeVariance;
```
bruce-ricard, line 70: I personally **really** try to avoid floating point numbers as much as I can, as they cause a lot of trouble. ( [What Every Computer Scientist Should Know About Floating-Point Arithmetic )
](https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html)

I understand why you went for them: you need to be able to represent a value like `2.5%`, and it has a decimal point, which makes them not representable by integers.

I have a suggestion. How about you use [per mille](https://en.wikipedia.org/wiki/Per_mille) to represent the variance instead of percent. That way, you could change your type to `int`, which is a lot more friendly than `double`, and you'll be able to represent `2.5%` as `25‰`. If you think that we'll need more precision than that, you can change the type to make it more precise. Like "per million" for example. Though IMO "per mille" is enough.

Please let me know your thoughts.

wu89694756:
Great suggestion, went ahead and changed it.

sksamuel:
Would love to see this merged.

sean-:
Can someone please bless and merge this PR?  Thundering herds are problematic and can only be prevented by the clients staggering their connections.

lfbayer:
I think I would prefer a solution to this problem that doesn't require the user to configure anything. I don't see any reason why we wouldn't want to avoid these timing issues for everyone.

It seems like there has to be a way to implement this that would allow the connection close timings to be as near a flat graph as possible. 

If, for example, the lifetime was 10 minutes and the connection count is 10, to be flat we would want a connection closed once per minute (rather than 10 near the end of 10 minutes). If the system is up and running with a fixed size pool this would be a fairly straight forward. But even with a fixed size pool, you don't really want a connection closed after exactly one minute of running after startup. So there would be some heuristic to ease into the target flat graph. In the case of a non-fixed size pool it might be a little less straight forward.

In this scenario, even when closing 1 every minute, we wouldn't likely want a little randomness in there as if there were two applications both running HikariCP pointed at the same database, there is a chance that they would be on identical timings which would result in a similar spiky graph as far as the database is concerned. I haven't actually thought this out but my instinct tells me that would be a concern.

Any thoughts?

Avarga954:
> I think I would prefer a solution to this problem that doesn't require the user to configure anything. I don't see any reason why we wouldn't want to avoid these timing issues for everyone.

My thought process for this PR was to make this configuration option transparent by default as not to impact any existing deployments existing Hikari config settings on upgrade. However this could be achieved by simply changing the default value I provided to be something more aggressive than 2.5% which is what it has OOTB today.

> It seems like there has to be a way to implement this that would allow the connection close timings to be as near a flat graph as possible.
> 
> If, for example, the lifetime was 10 minutes and the connection count is 10, to be flat we would want a connection closed once per minute (rather than 10 near the end of 10 minutes). If the system is up and running with a fixed size pool this would be a fairly straight forward. But even with a fixed size pool, you don't really want a connection closed after exactly one minute of running after startup. So there would be some heuristic to ease into the target flat graph. In the case of a non-fixed size pool it might be a little less straight forward.
> 
> In this scenario, even when closing 1 every minute, we wouldn't likely want a little randomness in there as if there were two applications both running HikariCP pointed at the same database, there is a chance that they would be on identical timings which would result in a similar spiky graph as far as the database is concerned. I haven't actually thought this out but my instinct tells me that would be a concern.
> 
> Any thoughts?

My thoughts are, when utilizing this approach it may take a very long time for larger connection pools to expire connections through existing settings like "maxLifeTime". If you had a 100 connections instead of 10 connections it would take approximately 100 minutes to cycle all of your db connection pool which may not be desirable for all use cases. Hypothetically we could go with some sort config setting that allows simultaneous connections to refresh based on a percentage of total connections to increase this speed however I still think you should still expose these setting to the user to allow for flexibility across different implementations.

Irrespective of that, I feel that this approach would require a substantial but not impossible change to the Hikari connection reuse flow. Currently if a connection exceeds its configured maxLifeTime + (variance) and is no longer utilized its immediately removed from the pool and released/renewed, where this approach would be batching connections for the next cycle.

The above solution you proposed would indeed deliver an even more normalized cadence than a customizable variance however, but I think we both solutions could work with the correct hikari config

lfbayer:
My gut tells me that there isn't really any reason one user would want a different configuration from another user. And by exposing it as a configuration parameter rather than just fixing the algorithm or behavior to be correct™ means that many users who would want a more smoothed out wouldn't get it without having to investigate.

One thought experiment: Is it conceivable that if we changed the default variance calculation to use your specific desired value it would actual not cause a problem for anyone else? 

You mention a target variance of 30%. What would be the negative of just changing the current hardcoded variance to be 30% rather than 2.5%?

Also, what is the MaxLifeTime setting you are using currently using? Maybe it makes sense to just change the current code to increase the variance in proportion of the MaxLifeTime.

If your change was just an unofficially supported system property I would feel more comfortable merging it. But the addition of a configuration property means it becomes API and somewhat locks us in on this implementation.

ssherwood:
@lfbayer this PR seems stuck unfortunately.

I opened the original request to uncouple the 2.5% variance from the hard coded value.  I personally am fine with a parameter to get around the hard coded value, I am also fine with any implementation that gets us close to a better solution.  I just don't want to see progress stopped without a solid alternative plan.

FYI, this problem frequently shows up in environments where applications are deployed to k8s and all scaled up at the same time.  The tight variance on a short maxLifetime frequently leads to a thundering herd of connections and observable spikes in latency on performance graphs.

Avarga954:
My apologies for not following this PR closer. 

I agree with @ssherwood 's point. At this point I would just like to get some solution for this merged so that others utilizing k8s deployments can mitigate thundering herds. I'm happy to refactor this PR in order accomplish this. 

I have been using 30% via a forked branch for my production application successfully for about a year now. If there is consensus that changing the default system value from 2.5% to 30% is the best approach I am happy to refactor to just change this value. I understand @lfbayer 's concerns about configuration locking us into an api, personally I prefer the flexibility of configuration but I'm open to either solution as long as we can make progress here.

svendiedrichsen:
CockroachLabs is proposing 10% but assume to have it configurable: https://www.cockroachlabs.com/docs/stable/connection-pooling#avoid-spikes-in-new-connections

It seems to me that having a `maxLifeTimeJitter` configurable is a common solution to common connection pooling problem. Common enough to have names for the problem like 'thundering herd' and 'connection storm'. IMHO there should be a configurable jitter for the maximum connection lifetime. Preferrably using the same config type as the `maxLifeTime` itself.

bantify:
@svendiedrichsen version 4.0.9 is not available in maven.

https://repo1.maven.org/maven2/com/zaxxer/HikariCP/

I'm confused about the solution.

svendiedrichsen:
> @svendiedrichsen version 4.0.9 is not available in maven.
> 
> https://repo1.maven.org/maven2/com/zaxxer/HikariCP/
> 
> I'm confused about the solution.

Foulest fixed the problem on his own fork.

src/main/java/com/zaxxer/hikari/HikariConfig.java
magdel: Per mille will do fine, agree

magdel:
So what else should it take to have it implemented? The feature is needed

randrasek:
@brettwooldridge Can we merge this in?  We are running this in production and seeing severe latency spikes due to connection pooling mismanagement.  It sounds like several people are suffering from this as well.

Thanks! :) 

(Also, sorry for all the edits to this comment.  Getting confused that this releases under a different name and there have been no "Releases" in a while.)

brettwooldridge:
I prefer not to increase the configuration surface area. It is better for the pool to have correct default behavior. The committed change 28f81da4 increases the default variance from 2.5% to 25%. The variance factor _can_ be altered by system property ``com.zaxxer.hikari.lifeTimeVarianceFactor`` but altering this value does not elevate to the level of standard pool configuration and should not need to be changed by users.

randrasek:
Wonderful!  Thank you so much!  🦸‍♂️

brettwooldridge:
@randrasek @Avarga954 I plan to perform a release in the next day or two after reviewing further pull requests.

randrasek:
Wow, thank you so much for your responsiveness!  I appreciate all that you do to maintain this project.

Avarga954:
@brettwooldridge  Excellent news! Thanks so much

PR #2038, opened 2023-02-07
No reason to issue a warning in the log for any clock-related shifts.

moreaki:
For years, people's logs have been filled up with warnings about thread starvation and retrograde clock change detections. This change sets the logger level to info for the following reasons:

- There is really nothing the end user can do with this warning, the code solves the issue by itself.
- It creates many questions and searches for solutions on the Internet because people see a WARN, where clearly an INFO would be sufficient
- None of the code paths which follows (hot or cold) has any WARN statements, therefore the granularity of the logger is misleading

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -787,7 +787,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
             // Detect retrograde time, allowing +128ms as per NTP spec.
             if (plusMillis(now, 128) < plusMillis(previous, housekeepingPeriodMs)) {
-               logger.warn("{} - Retrograde clock change detected (housekeeper delta={}), soft-evicting connections from pool.",
+               logger.info("{} - Retrograde clock change detected (housekeeper delta={}), soft-evicting connections from pool.",
                            poolName, elapsedDisplayString(previous, now));
                previous = now;
                softEvictConnections();
@@ -795,7 +795,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
             }
             else if (now > plusMillis(previous, (3 * housekeepingPeriodMs) / 2)) {
                // No point evicting for forward clock motion, this merely accelerates connection retirement anyway
-               logger.warn("{} - Thread starvation or clock leap detected (housekeeper delta={}).", poolName, elapsedDisplayString(previous, now));
+               logger.info("{} - Thread starvation or clock leap detected (housekeeper delta={}).", poolName, elapsedDisplayString(previous, now));
             }
 
             previous = now;


thiagosperandio:
That happened to me too. 

I agree that it should be a WARN, because we always need to be aware of problems that may occur in a production environment. 

So, it's a warning about hibernation, either of the machine or our service thread.

Maybe can increasing some additional information, like "checks if the machine was suspended".

moreaki:
> That happened to me too.
> 
> I agree that it should be a WARN, because we always need to be aware of problems that may occur in a production environment.
> 
> So, it's a warning about hibernation, either of the machine or our service thread.
> 
> Maybe can increasing some additional information, like "checks if the machine was suspended".

Confused by your reply. With whom do you agree? Certainly not with me (which is fine, but your explanation makes little sense). I clearly made a patch to set the level to INFO from WARN, so we don't get pointless information. Did you read my explanation? There is nothing that you're being warned for which you can influence at all and all cold and warm code paths following that statement are on level INFO. 

**In short**: a WARN level statement in the logs at this point in the code, as argued in my original statement, makes absolutely NO sense.

Let me know if I misunderstood something from your reply. Also, I would be extremely happy to hear from you on how you intend to address the issue outside of the already existing code which amends the reported issue further down in the code path. Any insights that a human intervention can do to alleviate this wrongly reported issue as a WARN would be most welcome, and probably close dozens of stackoverflow questions with people equally wondering why their logs fill up with statements of a potential issue over which they don't seem to have an influence.

benedictdube:
> I clearly made a patch to set the level to INFO from WARN, so we don't get pointless information. Did you read my explanation? There is nothing that you're being warned for which you can influence at all and all cold and warm code paths following that statement are on level INFO.
> 
> **In short**: a WARN level statement in the logs at this point in the code, as argued in my original statement, makes absolutely NO sense.

I'm in the early stages of my career in software eng.  and would like to confirm,  is it not appropriate to log at WARN level when there is something in the program that might (or can) lead to an unwanted state?  
For example;  If I encounter this frequently,  it could guide my decisions in the scheduled jobs I have on my droplet/EC2 instance if I encounter this frequently? Saving me from waiting for a higher logging level which might cause unavailability for the service.  This,  I doubt I would pick up if it were logged at INFO level.

dishi-github (approved):

brettwooldridge:
The warning is warranted and the the logging level is correct. It is a serious condition that can dramatically affect performance and even cause application outages. Feel free to suppress the warning via logging configuration.

moreaki:
@brettwooldridge I disagree for the three reasons I stated in my original pull request. Stating that it's correct is a weak argument.

People are constantly looking for solutions to this, while there is actually no solution to it. And nothing ever happened because of what the warning indicates, the code "resolves" the issue by itself. No drama ever occurred, no application outages I've ever heard of were a direct consequence of what happened after this warning logged.

Do you have one scenario that dramatically affects performance or application outages from that specific code path where the warning occurs? What can the user really do with this warning, beside suppress it? 

Your code is running on probably millions of instances because it's included in so many tech stacks. Is it really the best suggestion to everyone to suppress the excessive, and in my humble view, unnecessary logging of a warning that nobody can do anything about?

deahtstroke:
I'm having a hard time understanding your argument here, recently I came across some issues with the clock leaping in MacOS and it was beneficial for me to know these logs were in the WARN level instead of the INFO level because the state of my app is in an undesirable spot. INFO logs do not convey this, so whilst you saying 'something is correct' is not an universally strong argument, which I agree, in this case it is. WARN logs and INFO logs are two different things and convey different information about the state of the application.

moreaki:
@dvillavicencio What did you do to fix it on your system, so the warning went away? Did it really affect your MacOS or the application? How was it fixed?

Read my arguments again, please. They are not about the semantics of INFO versus WARN. It's about the consistency of the code cold/hot path on the informational as well as on the functional level following such information. Presently, there is literally nothing documented on what you could ever do to remedy the situation because the code addresses the issue itself somehow.

brettwooldridge:
@dvillavicencio I suggest you read [this thread](https://serverfault.com/questions/1123895/what-are-examples-of-software-that-may-be-seriously-affected-by-a-time-jump) for some examples.

Other examples (helpfully provided by ChatGPT):
 * On December 31, 2016, [Cloudflare's DNS service](https://blog.cloudflare.com/how-and-why-the-leap-second-affected-cloudflare-dns/) experienced a disruption due to the leap second. The software encountered a negative value where it expected zero, leading to a system panic.
 * Many caching systems, like memcached or Redis, rely on timestamps for expiry. If the clock shifts backwards, items in the cache may suddenly become valid again when they should have expired.
 * In a ZooKeeper-based system, backward time shifts caused leader election thrashing, as nodes assumed leaders had failed due to stale timestamps.
 * Stock trading platforms have seen delays and misreporting due to incorrect clock synchronization, especially around Daylight Saving Time transitions.
 * Certain network outages have been attributed to retrograde clock shifts in cellular base stations.
 * Distributed systems often rely on time for coordination and event sequencing. A clock shift can result in Out-of-order logs or database records, leading to consistency issues.

brettwooldridge:
Basically, my view is that this is a WARNing, not logged at ERROR, not logged at FATAL. Take note and react accordingly. In real world production systems enterprises deploy systems like Splunk that scrape logs looking for warnings just like this one and alerting IT staff of potential issues. Removing this warning could literally affect monitoring systems already in place. You will find HikariCP deployed in some of the largest systems that you interact with on the internet on a daily basis. And keep in mind that HikariCP does not arbitrarily warn about time shifts, only those that are outside of the limits specified by the NTP specification.

deahtstroke:
@moreaki I see your point about the need for consistency in logging and how warnings like this could be misleading, especially when they aren't actionable. However, the challenge lies in the fact that issues like thread pool starvation or clock leaps often result from external factors beyond the direct control or scope of HikariCP.

In my case, for example, the issue was caused by sleep-mode on my Mac, which I resolved by using the `caffeinate` command.  Documenting every potential cause and solution isn't feasible, as it would require addressing system-specific configurations and scenarios that HikariCP can't predict. I do understand the desire for better alignment between log severity and actionability, I believe this is less about logging levels and more about encouraging users to understand their systems deeply and troubleshoot accordingly. A generic "functional consistency" in logging for these cases might oversimplify the problem or create unrealistic expectations for maintainers. Instead, the current log levels serve as a prompt for investigation, which is, in itself, valuable.

@brettwooldridge Insightful reads! My previous response was in support of you argument, and with some recent troubleshooting I came to realize how important this warning log is. I do agree in totality for the logging to stay at the `WARN` level, thanks for your sharing your view.

PR #2055, opened 2023-03-23
Add support for Virtual Threads

bdeneuter:
It seems that `synchronized` will still pinn carrier threads in JDK 21 for the moment. This is the draft JEP for JDK21:
https://openjdk.org/jeps/8303683

So I'm reoping the PR for using ReentrantLock.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -36,6 +36,7 @@ import java.util.TreeSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
@@ -55,6 +56,7 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
    private static final long DEFAULT_KEEPALIVE_TIME = 0L;
    private static final int DEFAULT_POOL_SIZE = 10;
+   private static final ReentrantLock HIKARI_CONFIG_LOCK = new ReentrantLock();
 
    private static boolean unitTest = false;
 
@@ -1166,10 +1168,13 @@ public class HikariConfig implements HikariConfigMXBean
       final var prefix = "HikariPool-";
       try {
          // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
-         synchronized (System.getProperties()) {
+         HIKARI_CONFIG_LOCK.lock();
+         try {
             final var next = String.valueOf(Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1);
             System.setProperty("com.zaxxer.hikari.pool_number", next);
             return prefix + next;
+         } finally {
+            HIKARI_CONFIG_LOCK.unlock();
          }
       } catch (AccessControlException e) {
          // The SecurityManager didn't allow us to read/write system properties
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -36,7 +36,6 @@ import java.util.TreeSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadLocalRandom;
-import java.util.concurrent.locks.ReentrantLock;
 
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
@@ -56,7 +55,6 @@ public class HikariConfig implements HikariConfigMXBean
    private static final long MAX_LIFETIME = MINUTES.toMillis(30);
    private static final long DEFAULT_KEEPALIVE_TIME = 0L;
    private static final int DEFAULT_POOL_SIZE = 10;
-   private static final ReentrantLock HIKARI_CONFIG_LOCK = new ReentrantLock();
 
    private static boolean unitTest = false;
 
@@ -673,7 +671,7 @@ public class HikariConfig implements HikariConfigMXBean
          metricRegistry = getObjectOrPerformJndiLookup(metricRegistry);
 
          if (!safeIsAssignableFrom(metricRegistry, "com.codahale.metrics.MetricRegistry")
-             && !(safeIsAssignableFrom(metricRegistry, "io.micrometer.core.instrument.MeterRegistry"))) {
+            && !(safeIsAssignableFrom(metricRegistry, "io.micrometer.core.instrument.MeterRegistry"))) {
             throw new IllegalArgumentException("Class must be instance of com.codahale.metrics.MetricRegistry or io.micrometer.core.instrument.MeterRegistry");
          }
       }
@@ -1168,13 +1166,10 @@ public class HikariConfig implements HikariConfigMXBean
       final var prefix = "HikariPool-";
       try {
          // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
-         HIKARI_CONFIG_LOCK.lock();
-         try {
+         synchronized (System.getProperties()) {
             final var next = String.valueOf(Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1);
             System.setProperty("com.zaxxer.hikari.pool_number", next);
             return prefix + next;
-         } finally {
-            HIKARI_CONFIG_LOCK.unlock();
          }
       } catch (AccessControlException e) {
          // The SecurityManager didn't allow us to read/write system properties
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -199,56 +199,59 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    {
       hikariPoolLock.lock();
       try {
-         poolState = POOL_SHUTDOWN;
-
-         if (addConnectionExecutor == null) { // pool never started
-            return;
-         }
-
-         logPoolState("Before shutdown ");
-
-         if (houseKeeperTask != null) {
-            houseKeeperTask.cancel(false);
-            houseKeeperTask = null;
-         }
-
-         softEvictConnections();
-
-         addConnectionExecutor.shutdown();
-         if (!addConnectionExecutor.awaitTermination(getLoginTimeout(), SECONDS)) {
-            logger.warn("Timed-out waiting for add connection executor to shutdown");
-         }
-
-         destroyHouseKeepingExecutorService();
-
-         connectionBag.close();
-
-         final var assassinExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), poolName + " connection assassinator",
-                                                                           config.getThreadFactory(), new ThreadPoolExecutor.CallerRunsPolicy());
          try {
-            final var start = currentTime();
-            do {
-               abortActiveConnections(assassinExecutor);
-               softEvictConnections();
-            } while (getTotalConnections() > 0 && elapsedMillis(start) < SECONDS.toMillis(10));
-         }
-         finally {
-            assassinExecutor.shutdown();
-            if (!assassinExecutor.awaitTermination(10L, SECONDS)) {
-               logger.warn("Timed-out waiting for connection assassin to shutdown");
+            poolState = POOL_SHUTDOWN;
+
+            if (addConnectionExecutor == null) { // pool never started
+               return;
+            }
+
+            logPoolState("Before shutdown ");
+
+            if (houseKeeperTask != null) {
+               houseKeeperTask.cancel(false);
+               houseKeeperTask = null;
+            }
+
+            softEvictConnections();
+
+            addConnectionExecutor.shutdown();
+            if (!addConnectionExecutor.awaitTermination(getLoginTimeout(), SECONDS)) {
+               logger.warn("Timed-out waiting for add connection executor to shutdown");
+            }
+
+            destroyHouseKeepingExecutorService();
+
+            connectionBag.close();
+
+            final var assassinExecutor = createThreadPoolExecutor(config.getMaximumPoolSize(), poolName + " connection assassinator",
+                                                                              config.getThreadFactory(), new ThreadPoolExecutor.CallerRunsPolicy());
+            try {
+               final var start = currentTime();
+               do {
+                  abortActiveConnections(assassinExecutor);
+                  softEvictConnections();
+               } while (getTotalConnections() > 0 && elapsedMillis(start) < SECONDS.toMillis(10));
+            }
+            finally {
+               assassinExecutor.shutdown();
+               if (!assassinExecutor.awaitTermination(10L, SECONDS)) {
+                  logger.warn("Timed-out waiting for connection assassin to shutdown");
+               }
+            }
+
+            shutdownNetworkTimeoutExecutor();
+            closeConnectionExecutor.shutdown();
+            if (!closeConnectionExecutor.awaitTermination(10L, SECONDS)) {
+               logger.warn("Timed-out waiting for close connection executor to shutdown");
             }
          }
-
-         shutdownNetworkTimeoutExecutor();
-         closeConnectionExecutor.shutdown();
-         if (!closeConnectionExecutor.awaitTermination(10L, SECONDS)) {
-            logger.warn("Timed-out waiting for close connection executor to shutdown");
+         finally {
+            logPoolState("After shutdown ");
+            handleMBeans(this, false);
+            metricsTracker.close();
          }
-      }
-      finally {
-         logPoolState("After shutdown ");
-         handleMBeans(this, false);
-         metricsTracker.close();
+      } finally {
          hikariPoolLock.unlock();
       }
    }
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -38,6 +38,7 @@ import java.sql.SQLTransientConnectionException;
 import java.util.Optional;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.zaxxer.hikari.util.ClockSource.*;
 import static com.zaxxer.hikari.util.ConcurrentBag.IConcurrentBagEntry.STATE_IN_USE;
@@ -80,6 +81,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
    private final ScheduledExecutorService houseKeepingExecutorService;
    private ScheduledFuture<?> houseKeeperTask;
+   private final ReentrantLock hikariPoolLock = new ReentrantLock();
 
    /**
     * Construct a HikariPool with the specified configuration.
@@ -193,8 +195,9 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
     *
     * @throws InterruptedException thrown if the thread is interrupted during shutdown
     */
-   public synchronized void shutdown() throws InterruptedException
+   public void shutdown() throws InterruptedException
    {
+      hikariPoolLock.lock();
       try {
          poolState = POOL_SHUTDOWN;
 
@@ -246,6 +249,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
          logPoolState("After shutdown ");
          handleMBeans(this, false);
          metricsTracker.close();
+         hikariPoolLock.unlock();
       }
    }
 
@@ -368,25 +372,35 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
    /** {@inheritDoc} */
    @Override
-   public synchronized void suspendPool()
+   public void suspendPool()
    {
-      if (suspendResumeLock == SuspendResumeLock.FAUX_LOCK) {
-         throw new IllegalStateException(poolName + " - is not suspendable");
-      }
-      else if (poolState != POOL_SUSPENDED) {
-         suspendResumeLock.suspend();
-         poolState = POOL_SUSPENDED;
+      hikariPoolLock.lock();
+      try {
+         if (suspendResumeLock == SuspendResumeLock.FAUX_LOCK) {
+            throw new IllegalStateException(poolName + " - is not suspendable");
+         }
+         else if (poolState != POOL_SUSPENDED) {
+            suspendResumeLock.suspend();
+            poolState = POOL_SUSPENDED;
+         }
+      } finally {
+        hikariPoolLock.unlock();
       }
    }
 
    /** {@inheritDoc} */
    @Override
-   public synchronized void resumePool()
+   public void resumePool()
    {
-      if (poolState == POOL_SUSPENDED) {
-         poolState = POOL_NORMAL;
-         fillPool(false);
-         suspendResumeLock.resume();
+      hikariPoolLock.lock();
+      try {
+         if (poolState == POOL_SUSPENDED) {
+            poolState = POOL_NORMAL;
+            fillPool(false);
+            suspendResumeLock.resume();
+         }
+      } finally {
+         hikariPoolLock.unlock();
       }
    }
 
@@ -496,18 +510,23 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    /**
     * Fill pool up from current idle connections (as they are perceived at the point of execution) to minimumIdle connections.
     */
-   private synchronized void fillPool(final boolean isAfterAdd)
+   private void fillPool(final boolean isAfterAdd)
    {
-      final var idle = getIdleConnections();
-      final var shouldAdd = getTotalConnections() < config.getMaximumPoolSize() && idle < config.getMinimumIdle();
+      hikariPoolLock.lock();
+      try {
+         final var idle = getIdleConnections();
+         final var shouldAdd = getTotalConnections() < config.getMaximumPoolSize() && idle < config.getMinimumIdle();
 
-      if (shouldAdd) {
-         final var countToAdd = config.getMinimumIdle() - idle;
-         for (int i = 0; i < countToAdd; i++)
-            addConnectionExecutor.submit(isAfterAdd ? postFillPoolEntryCreator : poolEntryCreator);
-      }
-      else if (isAfterAdd) {
-         logger.debug("{} - Fill pool skipped, pool has sufficient level or currently being filled.", poolName);
+         if (shouldAdd) {
+            final var countToAdd = config.getMinimumIdle() - idle;
+            for (int i = 0; i < countToAdd; i++)
+               addConnectionExecutor.submit(isAfterAdd ? postFillPoolEntryCreator : poolEntryCreator);
+         }
+         else if (isAfterAdd) {
+            logger.debug("{} - Fill pool skipped, pool has sufficient level or currently being filled.", poolName);
+         }
+      } finally {
+         hikariPoolLock.unlock();
       }
    }
 
@@ -704,6 +723,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    private final class PoolEntryCreator implements Callable<Boolean>
    {
       private final String loggingPrefix;
+      private final ReentrantLock poolEntryLock = new ReentrantLock();
 
       PoolEntryCreator()
       {
@@ -754,9 +774,14 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
        *
        * @return true if we should create a connection, false if the need has disappeared
        */
-      private synchronized boolean shouldContinueCreating() {
-         return poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
-            (getIdleConnections() < config.getMinimumIdle() || connectionBag.getWaitingThreadCount() > getIdleConnections());
+      private boolean shouldContinueCreating() {
+         poolEntryLock.lock();
+         try {
+            return poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
+               (getIdleConnections() < config.getMinimumIdle() || connectionBag.getWaitingThreadCount() > getIdleConnections());
+         } finally {
+            poolEntryLock.unlock();
+         }
       }
    }
 
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyStatement.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyStatement.java
@@ -20,6 +20,7 @@ import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * This is the proxy class for java.sql.Statement.
@@ -33,6 +34,7 @@ public abstract class ProxyStatement implements Statement
 
    private boolean isClosed;
    private ResultSet proxyResultSet;
+   protected final ReentrantLock proxyStatementLock = new ReentrantLock();
 
    ProxyStatement(ProxyConnection connection, Statement statement)
    {
@@ -61,12 +63,15 @@ public abstract class ProxyStatement implements Statement
    @Override
    public final void close() throws SQLException
    {
-      synchronized (this) {
+      proxyStatementLock.lock();
+      try {
          if (isClosed) {
             return;
          }
 
          isClosed = true;
+      } finally {
+         proxyStatementLock.unlock();
       }
 
       connection.untrackStatement(delegate);
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
       <commons.csv.version>1.5</commons.csv.version>
       <h2.version>2.1.212</h2.version>
       <junit.version>4.13.1</junit.version>
-      <testcontainers.version>1.15.1</testcontainers.version>
+      <testcontainers.version>1.17.6</testcontainers.version>
    </properties>
 
    <groupId>com.zaxxer</groupId>
--- a/src/main/java/com/zaxxer/hikari/HikariDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/HikariDataSource.java
@@ -29,6 +29,7 @@ import java.sql.Connection;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.zaxxer.hikari.pool.HikariPool.POOL_NORMAL;
 
@@ -45,6 +46,7 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
 
    private final HikariPool fastPathPool;
    private volatile HikariPool pool;
+   private final ReentrantLock hikariDataSourceLock = new ReentrantLock();
 
    /**
     * Default constructor.  Setters are used to configure the pool.  Using
@@ -103,7 +105,8 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
       // See http://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java
       HikariPool result = pool;
       if (result == null) {
-         synchronized (this) {
+         hikariDataSourceLock.lock();
+         try {
             result = pool;
             if (result == null) {
                validate();
@@ -122,6 +125,8 @@ public class HikariDataSource extends HikariConfig implements DataSource, Closea
                }
                LOGGER.info("{} - Start completed.", getPoolName());
             }
+         } finally {
+            hikariDataSourceLock.unlock();
          }
       }
 
--- a/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
+++ b/src/main/java/com/zaxxer/hikari/HikariJNDIFactory.java
@@ -23,6 +23,7 @@ import javax.naming.spi.ObjectFactory;
 import javax.sql.DataSource;
 import java.util.Hashtable;
 import java.util.Properties;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * A JNDI factory that produces HikariDataSource instances.
@@ -31,26 +32,32 @@ import java.util.Properties;
  */
 public class HikariJNDIFactory implements ObjectFactory
 {
+   private final ReentrantLock hikariJNDIFactoryLock = new ReentrantLock();
    @Override
-   synchronized public Object getObjectInstance(Object obj, Name name, Context nameCtx, Hashtable<?, ?> environment) throws Exception
+   public Object getObjectInstance(Object obj, Name name, Context nameCtx, Hashtable<?, ?> environment) throws Exception
    {
-      // We only know how to deal with <code>javax.naming.Reference</code> that specify a class name of "javax.sql.DataSource"
-      if (obj instanceof Reference && "javax.sql.DataSource".equals(((Reference) obj).getClassName())) {
-         var ref = (Reference) obj;
-         var hikariPropSet = PropertyElf.getPropertyNames(HikariConfig.class);
+      hikariJNDIFactoryLock.lock();
+      try {
+         // We only know how to deal with <code>javax.naming.Reference</code> that specify a class name of "javax.sql.DataSource"
+         if (obj instanceof Reference && "javax.sql.DataSource".equals(((Reference) obj).getClassName())) {
+            var ref = (Reference) obj;
+            var hikariPropSet = PropertyElf.getPropertyNames(HikariConfig.class);
 
-         var properties = new Properties();
-         var enumeration = ref.getAll();
-         while (enumeration.hasMoreElements()) {
-            var element = enumeration.nextElement();
-            var type = element.getType();
-            if (type.startsWith("dataSource.") || hikariPropSet.contains(type)) {
-               properties.setProperty(type, element.getContent().toString());
+            var properties = new Properties();
+            var enumeration = ref.getAll();
+            while (enumeration.hasMoreElements()) {
+               var element = enumeration.nextElement();
+               var type = element.getType();
+               if (type.startsWith("dataSource.") || hikariPropSet.contains(type)) {
+                  properties.setProperty(type, element.getContent().toString());
+               }
             }
+            return createDataSource(properties, nameCtx);
          }
-         return createDataSource(properties, nameCtx);
+         return null;
+      } finally {
+         hikariJNDIFactoryLock.unlock();
       }
-      return null;
    }
 
    private DataSource createDataSource(final Properties properties, final Context context) throws NamingException
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -26,6 +26,7 @@ import java.sql.*;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.Executor;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static com.zaxxer.hikari.SQLExceptionOverride.Override.DO_NOT_EVICT;
 
@@ -63,6 +64,7 @@ public abstract class ProxyConnection implements Connection
    private int transactionIsolation;
    private String dbcatalog;
    private String dbschema;
+   protected final ReentrantLock proxyConnectionLock = new ReentrantLock();
 
    // static initializer
    static {
@@ -184,9 +186,14 @@ public abstract class ProxyConnection implements Connection
       return sqle;
    }
 
-   final synchronized void untrackStatement(final Statement statement)
+   final void untrackStatement(final Statement statement)
    {
-      openStatements.remove(statement);
+      proxyConnectionLock.lock();
+      try {
+         openStatements.remove(statement);
+      } finally {
+         proxyConnectionLock.unlock();
+      }
    }
 
    final void markCommitStateDirty()
@@ -201,32 +208,42 @@ public abstract class ProxyConnection implements Connection
       leakTask.cancel();
    }
 
-   private synchronized <T extends Statement> T trackStatement(final T statement)
+   private <T extends Statement> T trackStatement(final T statement)
    {
-      openStatements.add(statement);
+      proxyConnectionLock.lock();
+      try {
+         openStatements.add(statement);
 
-      return statement;
+         return statement;
+      } finally {
+         proxyConnectionLock.unlock();
+      }
    }
 
    @SuppressWarnings("EmptyTryBlock")
-   private synchronized void closeStatements()
+   private void closeStatements()
    {
-      final var size = openStatements.size();
-      if (size > 0) {
-         for (int i = 0; i < size && delegate != ClosedConnection.CLOSED_CONNECTION; i++) {
-            try (Statement ignored = openStatements.get(i)) {
-               // automatic resource cleanup
+      proxyConnectionLock.lock();
+      try {
+         final var size = openStatements.size();
+         if (size > 0) {
+            for (int i = 0; i < size && delegate != ClosedConnection.CLOSED_CONNECTION; i++) {
+               try (Statement ignored = openStatements.get(i)) {
+                  // automatic resource cleanup
+               }
+               catch (SQLException e) {
+                  LOGGER.warn("{} - Connection {} marked as broken because of an exception closing open statements during Connection.close()",
+                     poolEntry.getPoolName(), delegate);
+                  leakTask.cancel();
+                  poolEntry.evict("(exception closing Statements during Connection.close())");
+                  delegate = ClosedConnection.CLOSED_CONNECTION;
+               }
             }
-            catch (SQLException e) {
-               LOGGER.warn("{} - Connection {} marked as broken because of an exception closing open statements during Connection.close()",
-                           poolEntry.getPoolName(), delegate);
-               leakTask.cancel();
-               poolEntry.evict("(exception closing Statements during Connection.close())");
-               delegate = ClosedConnection.CLOSED_CONNECTION;
-            }
-         }
 
-         openStatements.clear();
+            openStatements.clear();
+         }
+      } finally {
+         proxyConnectionLock.unlock();
       }
    }
 


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -246,6 +249,7 @@ public synchronized void shutdown() throws InterruptedException
          logPoolState("After shutdown ");
          handleMBeans(this, false);
          metricsTracker.close();
+         hikariPoolLock.unlock();
```
svendiedrichsen, line 255: Shouldn't this be in a finally block?

lfbayer:
Reading this page seems to imply that having the synchronized blocks is supported in Java 19.
https://neiljbrown.com/2022/07/23/preparing-for-java-virtual-threads/

> However, currently there is a limitation which means code executing in a synchronized block may in some cases hinder the throughput benefits of virtual threads.

This statement in that page implies that the synchronized blocks might not be a problem. Do you have any benchmarks or tests that can show that there is a real benefit to changing to using the locks? Or is there some evidence that the synchronized blocks are in fact not supported?

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
bdeneuter: Yes, you are right. If one of the statements would throw an exception, the lock won't be released.

bdeneuter:
@lfbayer Virtual threads use a small number of carrier threads (OS Threads) to execute. Normally this will be the same amount as the number of CPUs. If the virtual thread runs in a synchronized block, the carrier thread won't be available to execute other work when the virtual thread performs blocking IO. So it depends. When no IO is being performed, synchronized is fine to use. But when IO is being performed it will be better to use a ReentrantLock as the carrier thread can perform other work in the meantime. 

I was hoping that this limitation would be gone in JDK 21 but as the draft JEP is being prepared, it currently seems the limitation will still be present in JDK 21. It might still change of course in the coming months.

akashchouhan16 (approved):

salatmaster:
> I was hoping that this limitation would be gone in JDK 21 but as the draft JEP is being prepared, it currently seems the limitation will still be present in JDK 21. It might still change of course in the coming months.

There is already confirmed information that in Loom GA in JDK 21 this problem still persists. So I think we should accept this pull request.

https://openjdk.org/jeps/444
изображение

salatmaster (approved):

salatmaster (changes requested):
lock() call must be in try {} block

bdeneuter:
> lock() call must be in try {} block

I don't think it must be in the try block. This is documented in the ReentrantClass:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html

```
class X {
   private final ReentrantLock lock = new ReentrantLock();
   // ...

   public void m() {
     lock.lock();  // block until condition holds
     try {
       // ... method body
     } finally {
       lock.unlock();
     }
   }
 }
```

salatmaster:
> > lock() call must be in try {} block
> 
> I don't think it must be in the try block. This is documented in the ReentrantClass: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/locks/ReentrantLock.html
> 
> ```
> class X {
>    private final ReentrantLock lock = new ReentrantLock();
>    // ...
> 
>    public void m() {
>      lock.lock();  // block until condition holds
>      try {
>        // ... method body
>      } finally {
>        lock.unlock();
>      }
>    }
>  }
> ```

you're right

cj848:
When can this PR be merged and released? 
 I'm using it through Virtual Thread and I'm really looking forward to it.

enragedginger:
JDK 21 has released. Virtual threads are no longer an experimental feature. Can we get a new release with this change included?

XcrigX:
This is somewhat of a philosophical question related to this PR.  

In previous multi-threaded usages involving Hikari, there was likely a threadpool of platform threads which effectively throttled the number of requests to the connection pool. 
If applications simply switch from a fixed threadpool to virtual threads this throttling goes away. 
One can imagine then that there might be higher contention over connections in the connection pool, resulting in (likely) timeout exceptions waiting for a free connection. 

I'm trying to figure out where/how to handle this? 

Applications can implement their own counting semaphores around DB-bound operations to re-introduce throttling - but such operations could be sprinkled all over an application. It's probably just cleaner to stick with platform threads in a threadpool at that point. 

Applications could increase the connection pool timeout, but this may mask actual errors creating DB connections and seems iffy regardless. 

Should Hikari itself put counting Semaphores in getConnection() outside of the timeout checking? 

Nothing seems quite right. 

Thoughts?

bdeneuter:
Hi, my pull request was not trying to solve philosophical questions or determine what is the best way to use HikariCP with virtual threads. The goal is to make HikariCP virtual thread friendly and avoid pinning of the carrier thread when it is being called from a virtual thread. 
With virtual threads, a small set of carrier threads, which are OS threads, are being used. These few carrier threads are used by all virtual threads running in the Java process. So also virtual threads that don't call HikariCP (Tomcat, http client, ...) are using the same carrier threads. 
If HikariCP pins the carrier threads during IO then these carrier threads are not available for other virtuals threads and the service becomes unresponsive. This can happen when HikariCP opens a DB connection and has the virtual thread pinned for example.

Pinning the carrier thread is a limitation from virtual threads in JDK 21. The developers from Loom said that they will work on avoiding pinning the carrier thread when synchronize is used. So in future versions of the JDK it is a problem that will go away but I'm not sure if they will backport their fix to JDK 21.

enragedginger:
@XcrigX The Oracle[ docs on virtual threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-E695A4C5-D335-4FA4-B886-FEB88C73F23E) specifically address this use case. They recommend "unlearning this pattern" and using a counting semaphore.

(Look specifically for the "Use Semaphores to Limit Concurrency" section in the link above)

XcrigX:
@bdeneuter - understood - it's a welcome PR. I'm just trying to spark some discussion/thought around "what's next". 

@enragedginger - yes, counting semaphores are the likely answer, but from a client perspective this can be messy. If your data access operations are spread out in many places it's pretty ugly to go wrap each in a try/finally block, deal with InterruptedExceptions acquiring the semaphore, etc.  

getConnection() in Hikari is a potentially clean place to do this but may break existing contracts with the timeout. 
Actually, getConnection() is already using a semaphore in SuspendResumeLock which is acquired prior to the timeout clock starting, but it has 10000 permits. If this number could be adjusted to be based on the connection pool max - it may solve the problem I'm alluding to.

svendiedrichsen:
IMHO HikariCP already conceptually provides a (timeout-based) semaphore to guard access to pooled connections. This PR only prevents virtual threads to pin to platform threads while accessing the pool.

cj848:
I'm really curious, when will this PR be merged?

lfbayer:
As far as I understand, applications can continue to use virtual threads in their applications while using HikariCP, but it just means that there is a potential for the JVM to spawn non-virtual threads.

So that means that the purpose of this pull request presumably isn't specifically to be able to use HikariCP in an environment that uses virtual threads but rather, the intent is to get some performance improvement by having HikariCP itself use the virtual threads.

In my very first comment I asked for benchmarks that demonstrated the performance impact of these changes.

These changes are not trivial, and the risk to existing users is still there. I would want assurance that these changes are both correct and meaningful. And meaningful beyond just a cargo-cult thought process of "syncrhonized = bad". I don't entirely disagree with the change, but I also think, if semantically a lock and a synchronize block are the same, shouldn't it be on the JVM itself to add support for this? If they are not semantically the same, then clearly there are other risks of change from synchronize to locks.

This is the kind of change that I would want to truly understand myself before merging. To be perfectly honest, I don't have any immediate plan to put the work in to understand the change and its impact.

I think the "philosophical questions" are actually very much worth having. Part of me thinks that the model for a connection pool in the virtual thread use-case is likely a bit different from HikariCP's implementation. My gut tells me that simply changing the synchronize blocks to locks wouldn't actually solve as much as it might first seem.

To answer the question of "when will this be merged": I don't have any specific plan to do so. My gut tells me it would take two days of full time work to build the confidence that this change is safe. And I don't see myself spending those two days. If someone else can provide the evidence and convincing argument for why this specific change is safe, then I would be happy to merge it. And I mean this specific change, not just an argument for virtual-threads in general. I would also like to hear a good explanation of the risks and impact of the change, as well as how conceptually a database connection pool should work in virtual-thread environment.

I'm also really curious about how the current implementation actually behaves in a virtual-thread environment. Was the impetus for this PR in response to real experience using HikariCP with virtual threads? I think almost all of the current hot-path synchronize blocks aren't expected to ever really block anyway. For example, the getConnection() method actually uses zero synchronize blocks (it only uses a synchronize if the pool isn't already initialized). So if none of the hot-paths synchronize then is there any performance benefit to these changes?

enragedginger:
@lfbayer Thanks for the thorough explanation. It's helpful to understand where you're at on all of this.

> As far as I understand, applications can continue to use virtual threads in their applications while using HikariCP, but it just means that there is a potential for the JVM to spawn non-virtual threads.

I don't think this assumption is correct. Check out the section from the Oracle docs on virtual threads titled [Scheduling Virtual Threads and Pinned Virtual Threads](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-704A716D-0662-4BC7-8C7F-66EE74B1EDAD).

If a virtual thread hits a `synchronized` block of code, it remains pinned to its OS carrier thread and cannot be unpinned until that block of code finishes.

Deadlocks due to resource starvation due to pinned virtual threads is the situation we're all trying to avoid.

I've already encountered this situation with virtual threads when using pooling with the Apache commons HTTP library and it sucks. Switching to using `ReentrantLock` / `Semaphore` is simple if you're in control of the offending library, but is annoying to deal with if you're an application developer like me who just free loads on open source libraries others have created. I ultimately just rolled my own virtual thread safe object pooling code and switched over to the internal HttpClient provided by the JDK. It was annoying, but now I get the benefits of virtual threads for my IO and don't have deadlocks in my system.

FWIW, the section of the docs on [Avoid Lengthy and Frequent Pinning](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html#GUID-04C03FFC-066D-4857-85B9-E5A27A875AF9) states:

> Pinning may adversely affect the throughput of the server if the blocking operation is both long-lived and frequent. Guarding short-lived operations, such as in-memory operations, or infrequent ones with synchronized blocks or methods should have no adverse effect.

After reading this statement and re-reviewing this PR, I don't think any of the current usages of `synchronized` classify as `both long-lived and frequent` and therefore using synchronized `should have no adverse effect`. If `synchronized` was being used to wrap the execution of SQL statements, then we'd have a problem, but that doesn't appear to be the case here (someone please tell me if I'm wrong on this though).

> if semantically a lock and a synchronize block are the same, shouldn't it be on the JVM itself to add support for this?

This is also an interesting statement that I hadn't initially considered. That aforementioned section of the docs has this too:

> A _current_ limitation of the implementation of virtual threads is that performing a blocking operation while inside a synchronized block

I'd largely skipped over the word "current" there; it implies that this might change in the future. The docs advocate for switching to use explicit `ReentrantLock` / `Semaphore` instead of `synchronized`--implying it's a suitable drop-in replacement. However, if developers who wish to use virtual threads have to wait for all of their favorite IO libraries to move away from `synchronized` until they can safely use virtual threads, the adoption rate for virtual threads is going to be pretty low. Therefore, I'd like to think that some future release of the JVM will remove this `synchronized` limitation, but that's wishful thinking at this point in time.

In summary, I don't see how the changes from `synchronized` over to `ReentrantLock` in this PR would cause any kinds of troubles for existing users as the docs imply that `ReentrantLock` is a suitable drop-in replacement. The tests still pass, so that should be sufficient to guarantee we're not breaking existing users, no?

However, before advertising that "HikariCP is virtual thread compatible" it'd be good to add JDK 21 to the testing grid and add some virtual thread executor related test cases. IMO it doesn't make sense to do that until after this lands though.

Thoughts?

src/main/java/com/zaxxer/hikari/HikariConfig.java
```diff
@@ -1166,10 +1168,13 @@ private String generatePoolName()
       final var prefix = "HikariPool-";
       try {
          // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
-         synchronized (System.getProperties()) {
+         HIKARI_CONFIG_LOCK.lock();
```
lfbayer: If I understand correctly, this synchronized block is actually a bit of a hack to handle the case where there might be two separate HikariConfig classes in the JVM at the same time. In this case the locks wouldn't be the same and you could get a race condition between these two as now there would be two "HIKARI_CONFIG_LOG" instances. (Even though it is static, the same class loaded into two different class loaders essentially ends up being two different classes.)

lfbayer:
@enragedginger thank you for your detailed reply.

I also wouldn't go as far as saying HikariCP is specifically "virtual thread compatible" at this point. Although I don't feel confident saying it _isn't_ either. It seems like there is potentially other issues than just the use of synchronize that might mean we don't play well. Since I don't have ANY experience using virtual threads yet it's hard for me to feel comfortable with any judgement on whether this change is sufficient. 

My main concern with using the ReentrantLock is that even if it doesn't change anything functionally, it might have a negative impact on performance somehow. 

For context: when I write code these days I always prefer java.util.concurrent over synchronized. So I have nothing against ReentrantLock and no specific preference for synchronized. What I don't like is change without full comprehension of the impact of the change.

I would absolutely expect Reentrant locks to not perform worse than synchronize, but I still want to see some benchmark numbers to verify this before merging. Just because junit passes doesn't mean there is no negative impact.

I also feel like the issue of pinning a virtual thread still wouldn't be enough to cause a deadlock. It would make me feel more comfortable if I could see a concrete case where this actually is a problem.

Also, I'm still not satisfied with the answer related to locks and synchronize blocks being semantically the same. If the JVM doesn't support something with synchronize that it does with Reentrant locks, then that implies there is something about synchronize that is different. So what are the reasons that synchronize support isn't available. Knowing these underlying reasons would be helpful for us to analyze whether those same reasons would have any impact on us.

For me to be satisfied I need the following questions to be resolved:
1. What is a specific example of code that causes a virtual thread to be pinned? A JUnit case that uses virtual-threads and can force this scenario would be appreciated, because then we could be guaranteed no future change could cause a regression without us noticing.
2. If the code paths that can cause pinning are only in the initial DataSource setup code, then is this really an issue that needs to be solved?
3. For a user to get support for virtual-threads with this pull request, are they required to provide their own ThreadFactory to the HikariConfig? If so, is there a more idiomatic way a library is expected to be implemented so that the library doesn't create non-virtual-threads when run within a vritual-thread environment? How is the library user supposed to know they need to do this. Or is HikariCP's creation of its own threads not a problem?
4. I would like to see the existing benchmarks run pre- and post- PR and see a comparison of the performance to show that it is at least unchanged if not better.

lfbayer (changes requested):

src/main/java/com/zaxxer/hikari/HikariConfig.java
lfbayer: There is a ticket (#2104) that describes a bug with the existing implementation. I would like a solution that solves that bug that also satisfies the virtual thread requirements. I had a terrible idea for a solution that would be something like this:

```java
   private String generatePoolName()
   {
      final var prefix = "HikariPool-";

      String failureReason;
      // Pool number is global to the VM to avoid overlapping pool numbers in classloader scoped environments
      try {
         // create a unique ID to identify ourselves for the locking semantics below
         String lockId = UUID.randomUUID().toString();

         // get the start time so that we can check how long we have been running and implement a timeout.
         long startTime = System.currentTimeMillis();

         // This code is essentially reimplementing lock semantics using system properties in order to get an atomically
         // incrementing pool number. We can't use a global lock since that won't work across classloaders.
         do {
            String lock = System.getProperty("com.zaxxer.hikari.pool_number.lock");
            // if the lock is null, then another thread is not definitely running.
            if (lock == null) {
               // set the lock to our ID to indicate that we want to start work
               System.setProperty("com.zaxxer.hikari.pool_number.lock", lockId);

               // get the lock again to check that no other thread has
               lock = System.getProperty("com.zaxxer.hikari.pool_number.lock");
               if (lockId.equals(lock)) {
                  try {
                     final var next = String.valueOf(Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1);
                     System.setProperty("com.zaxxer.hikari.pool_number", next);
                     return prefix + next;
                  }
                  finally {
                     System.clearProperty("com.zaxxer.hikari.pool_number.lock");
                  }
               }
            }

            // If we are here it means there is another thread trying to set the pool number. Let's tell the
            // JVM to let other threads run first since we are waiting and we don't want to spin too fast.
            Thread.yield();
         } while (System.currentTimeMillis() - startTime < 5000L);

         failureReason = "timed out waiting for lock on pool numbering";
      } catch (AccessControlException e) {
         failureReason = "security manager prevented access to system properties";
      }

      // The SecurityManager didn't allow us to read/write system properties
      // so just generate a random pool number instead
      final var random = ThreadLocalRandom.current();
      final var buf = new StringBuilder(prefix);

      for (var i = 0; i < 4; i++) {
         buf.append(ID_CHARACTERS[random.nextInt(62)]);
      }

      LOGGER.info("assigned random pool name '{}' ({})", buf, failureReason);

      return buf.toString();
   }
   ```

Comment (bdeneuter) [reply src/main/java/com/zaxxer/hikari/HikariConfig.java]: @lfbayer you are right and I reverted this file back to use the synchronized block

Comment (lfbayer) [reply src/main/java/com/zaxxer/hikari/HikariConfig.java]: But this code is also bad, as I believe I already see a race condition.

**Comment by bdeneuter** [comment]
> Also, I'm still not satisfied with the answer related to locks and synchronize blocks being semantically the same. If the JVM doesn't support something with synchronize that it does with Reentrant locks, then that implies there is something about synchronize that is different

The pinning of the OS carrier thread occurs when calling native code from Java. It seems that Java is calling native code in the implementation for Java monitors and this seems to be the root cause that synchronized pins the carrier thread. That is why in the JEP it is specified that both native calls and synchronized pins the carrier thread. 

I found this BUG report for OpenJDK where they mention that they are reimplementing Java monitors:
https://bugs.openjdk.org/browse/JDK-8305069

So I'm fine to wait for this new implementation and to not make code changes to avoid pinning. Not sure if they would backport this to JDK 21 though. I would expect not.

**Comment by lfbayer** [comment]
@bdeneuter Your answer satisfies me as to how synchronized and ReentrantLock are different.

I think looking forward, even if support for synchronized in virtual-threads is added, HikariCP likely needs to be changed somehow. Do you have any thoughts on my questions above? Specifically 1 and 3.

My comments up until now might sounds like I do not want to fix this problem. But the opposite is true, I just want to be absolutely sure the answer is the correct one.

**Comment by bdeneuter** [comment]
Each time the code enters a synchronized block the underlying carrier thread gets pinned. If this is a bad thing depends on what happens in the synchronized block. If the virtual thread yields and gets parked, the carrier thread stays pinned and won't pick up other virtual threads that want to continue. If all carrier threads are pinned, the system can block and stops working. If the virtual thread doesn't yield and just leaves the synchronized block after it did its work everything is fine.

This is a nice example that is not complex and where the pinning  can cause a deadlock:
https://www.reddit.com/r/java/comments/1512xuo/virtual_threads_interesting_deadlock/

This is their question on the loom-dev mailing list:
https://mail.openjdk.org/pipermail/loom-dev/2023-July/005989.html

What happens in this example:

-  Virtual threads entering a synchronized block and each virtual thread will pin its carrier thread while in the synchronized block
- It appears that System.out.println uses a ReentrantLock underneath
- When a virtual thread calls the System.out.println before the synchronized block, this virtual thread can get the lock on the internal ReentrantLock. This virtual yields because it has the lock and can perform its IO.
- The other virtual threads that entered the synchronized block are all waiting for getting the internal lock.
- But when the virtual thread that has the internal lock wants to continue after it performed its IO, it can not continue as all carrier threads are pinned
- And the system is in a deadlock

To me it starts to look like it is safer to wait until they solved the pinning of the threads.

To answer your other question. I don't think it would break the users application if HikariCP creates OS threads. Of course it would be better if the different libraries avoid creating their own OS threads or allow the user  to configure the library to use virtual threads if they want them. 

> My comments up until now might sounds like I do not want to fix this problem. But the opposite is true, I just want to be absolutely sure the answer is the correct one.

I understand and don't worry. Everybody is searching on how to use this technology with its current limitations. I prefer to not rush it in if there are doubts.

**Comment by XcrigX** [comment]
I can add some color to my own experience trying to use virtual threads with Hikari. 

I have a batch processing app that takes a set of data in one format, and converts to a flattened SQL model. 
The app uses a regular thread pool i.e. (Executors.newFixedThreadPool(8)). 

It receives a an array of records, and for each it creates a task and submits to the ExecutorService to schedule and run on a thread, then blocks for them all to complete. 
Each task will do some conversion on the source data, and eventually insert or update a record in the DB. 

I switched to using virtual threads to see how it would work, pretty much by simply replacing the fixed threadpool with: 
Executors.newVirtualThreadPerTaskExecutor();

So suppose I receive 10000 records in a batch. Before at most 8 (threadpool size) could ever be executing concurrently. 
Hikari's default connection pool size is 8, so there would always be a free connection. A connection timeout would mean an actual problem connecting to the DB, not just an timeout waiting for a free connection from the pool. 

When switching to virtual threads there is no limit on the number of threads that can be running concurrently. Since the DB IO is the main bottleneck, many more than 8 virtual threads will spin up. Eventually they all start to stack up waiting for a connection from Hikari. And eventually, one of the tasks will wait > 30 seconds and timeout. 

Again, my suggestion to alleviate this problem would be to have a counting semaphore in getConnection() that re-introduces some throttling based on the connection pool size before the timeout timer starts. This may change the existing contract with clients however, so there may be more to it - some additional explicit configuration or something so clients can "opt-in" to this contract change.

**Comment by svendiedrichsen** [comment]
@XcrigX If HikariCP would allow to disable the connection timeout you essentially would end up with a counting semaphore.

**Comment by XcrigX** [comment]
@svendiedrichsen   I suppose one could set the connection timeout to a really high value to approximate it. My thought was that I'd still want to know about problems actually getting a connection from the DB vs just waiting for a connection to become available from the pool.

**Comment by svendiedrichsen** [comment]
@XcrigX But this is more a question on how HikariCP differentiates those two cases. i.e. By using different exceptions.

**Comment by dpiliouras-ut** [comment]
I'm certainly not familiar with the hikariCP internals, but it sounds like there are two topics here:

1. Allowing/enabling **upstream code** to use virtual-threads (w/o pinning). Think of the canonical web-server use-case, where your server is configured to use virtual-threads, and the actual handlers that will run on these threads, talk to the DB. You really don't want these handlers to pin the carrier thread. If I'm understanding correctly, this PR addresses, or is attempting to address, this part. A good/similar example would be the native `HttpClient`, which was tweaked to not pin (in java-21) in a similar fashion. 
2. Allowing/configuring hikariCP **itself** to use virtual-threads (i.e. internally). This may be a nice enhancement, but is somewhat unrelated to this PR.  Yes, a semaphore will be needed if this 'enhancement' goes forward (because virtual executors are unbounded, and you're not supposed to pool virtual-threads), but as far as I understand, there is nothing wrong with hikariCP spawning native threads (e.g. via a pooled-executor) internally. If these threads are doing a lot of blocking IO, then perhaps this increases the value proposition of the aforementioned enhancement, but still, the two should not be conflated.  In fact, the example with  the native `HttpClient` applies here too! It also provides support for a custom executor (or thread-factory), for the threads **it** may spawn.

Finally, i think the deadlock behaviour (with `println`), can only occur under somewhat unique circumstances.  As Ron (the tech-lead for project loom) states [here](https://mail.openjdk.org/pipermail/loom-dev/2023-July/005990.html):

> The problem occurs when some shared j.u.c lock is acquired both when some virtual thread is pinned (i.e. inside a synchronized block/method) as well as unpinned

Hope my understanding is correct, and that this helps someone 👍 . Thanks for the great work - eagerly awaiting on a resolution...

**Comment by jimpil** [comment]
Hello,

I was just watching the java 22 launch live stream, and apparently, `synchronized` still pins the carrier thread :( .  Is there any news/consensus around replacing with `ReentrantLock`, or is this PR considered 'stale'?  thanks in advance...

**Comment by sureshg** [comment]
@jimpil Sorry, this isn't answering your question.  I'd like to let you know that the short-term fix for pinning is available on the Loom-ea build (https://mail.openjdk.org/pipermail/loom-dev/2024-February/006433.html). Hopefully, this will be ready for production use on JDK 23.

**Comment by sinsuren** [comment]
@sureshg given that java 21 is GA, are we taking a stand that users of HikariCP lib should not use virtual thread  and have to wait for java 25 (optimistic that issue of pinning gets fixed in java 25).

File: src/main/java/com/zaxxer/hikari/pool/ProxyStatement.java [hunk 3]
```diff
@@ -61,12 +63,15 @@ public final String toString()
    @Override
    public final void close() throws SQLException
    {
-      synchronized (this) {
+      proxyStatementLock.lock();
```
sinsuren, line 66: This is certainly not needed.

sinsuren:
Did a minor benchmarking with spring boot 3.2 + mysql-connector-j(8.3.0) + HikariCP (5.1.0) and results are already mentioned in this blog post by another user. https://medium.com/naukri-engineering/virtual-thread-performance-gain-for-microservices-760a08f0b8f3

Stack trace while threads are getting pinned.

```
Thread[#60,ForkJoinPool-1-worker-10,5,CarrierThreads]
    java.base/java.lang.VirtualThread$VThreadContinuation.onPinned(VirtualThread.java:185)
    java.base/jdk.internal.vm.Continuation.onPinned0(Continuation.java:393)
    java.base/java.lang.VirtualThread.parkNanos(VirtualThread.java:631)
    java.base/java.lang.System$2.parkVirtualThread(System.java:2648)
    java.base/jdk.internal.misc.VirtualThreads.park(VirtualThreads.java:67)
    java.base/java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:408)
    java.base/sun.nio.ch.Poller.pollIndirect(Poller.java:137)
    java.base/sun.nio.ch.Poller.poll(Poller.java:102)
    java.base/sun.nio.ch.Poller.poll(Poller.java:87)
    java.base/sun.nio.ch.NioSocketImpl.park(NioSocketImpl.java:175)
    java.base/sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:280)
    java.base/sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:304)
    java.base/sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:346)
    java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:796)
    java.base/java.net.Socket$SocketInputStream.read(Socket.java:1099)
    com.mysql.cj.protocol.ReadAheadInputStream.fill(ReadAheadInputStream.java:107)
    com.mysql.cj.protocol.ReadAheadInputStream.readFromUnderlyingStreamIfNecessary(ReadAheadInputStream.java:150)
    com.mysql.cj.protocol.ReadAheadInputStream.read(ReadAheadInputStream.java:180) <== monitors:1
    java.base/java.io.FilterInputStream.read(FilterInputStream.java:119)
    com.mysql.cj.protocol.FullReadInputStream.readFully(FullReadInputStream.java:64)
    com.mysql.cj.protocol.a.SimplePacketReader.readHeaderLocal(SimplePacketReader.java:81)
    com.mysql.cj.protocol.a.SimplePacketReader.readHeader(SimplePacketReader.java:63)
    com.mysql.cj.protocol.a.SimplePacketReader.readHeader(SimplePacketReader.java:45)
    com.mysql.cj.protocol.a.TimeTrackingPacketReader.readHeader(TimeTrackingPacketReader.java:52)
    com.mysql.cj.protocol.a.TimeTrackingPacketReader.readHeader(TimeTrackingPacketReader.java:41)
    com.mysql.cj.protocol.a.MultiPacketReader.readHeader(MultiPacketReader.java:54)
    com.mysql.cj.protocol.a.MultiPacketReader.readHeader(MultiPacketReader.java:44)
    com.mysql.cj.protocol.a.NativeProtocol.readMessage(NativeProtocol.java:576)
    com.mysql.cj.protocol.a.NativeProtocol.checkErrorMessage(NativeProtocol.java:762)
    com.mysql.cj.protocol.a.NativeProtocol.sendCommand(NativeProtocol.java:701)
    com.mysql.cj.protocol.a.NativeProtocol.sendCommand(NativeProtocol.java:157)
    com.mysql.cj.NativeSession.ping(NativeSession.java:735)
    com.mysql.cj.jdbc.ConnectionImpl.pingInternal(ConnectionImpl.java:1468)
    com.mysql.cj.jdbc.ConnectionImpl.isValid(ConnectionImpl.java:2496) <== monitors:1
    com.zaxxer.hikari.pool.PoolBase.isConnectionDead(PoolBase.java:156)
    com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:170)
    com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:146)
    com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:100)
    org.hibernate.engine.jdbc.connections.internal.DatasourceConnectionProviderImpl.getConnection(DatasourceConnectionProviderImpl.java:122)
    org.hibernate.internal.NonContextualJdbcConnectionAccess.obtainConnection(NonContextualJdbcConnectionAccess.java:46)
    org.hibernate.resource.jdbc.internal.LogicalConnectionManagedImpl.acquireConnectionIfNeeded(LogicalConnectionManagedImpl.java:113)
    org.hibernate.resource.jdbc.internal.LogicalConnectionManagedImpl.getPhysicalConnection(LogicalConnectionManagedImpl.java:143)
    org.hibernate.engine.jdbc.internal.StatementPreparerImpl.connection(StatementPreparerImpl.java:54)
    org.hibernate.engine.jdbc.internal.StatementPreparerImpl$5.doPrepare(StatementPreparerImpl.java:153)
    org.hibernate.engine.jdbc.internal.StatementPreparerImpl$StatementPreparationTemplate.prepareStatement(StatementPreparerImpl.java:183)
    org.hibernate.engine.jdbc.internal.StatementPreparerImpl.prepareQueryStatement(StatementPreparerImpl.java:155)
    org.hibernate.sql.exec.spi.JdbcSelectExecutor.lambda$list$0(JdbcSelectExecutor.java:85)
    org.hibernate.sql.results.jdbc.internal.DeferredResultSetAccess.executeQuery(DeferredResultSetAccess.java:231)
    org.hibernate.sql.results.jdbc.internal.DeferredResultSetAccess.getResultSet(DeferredResultSetAccess.java:167)
    org.hibernate.sql.results.jdbc.internal.AbstractResultSetAccess.getMetaData(AbstractResultSetAccess.java:36)
    org.hibernate.sql.results.jdbc.internal.AbstractResultSetAccess.getColumnCount(AbstractResultSetAccess.java:52)
    org.hibernate.query.results.ResultSetMappingImpl.resolve(ResultSetMappingImpl.java:193)
    org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.resolveJdbcValuesSource(JdbcSelectExecutorStandardImpl.java:325)
    org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.doExecuteQuery(JdbcSelectExecutorStandardImpl.java:115)
    org.hibernate.sql.exec.internal.JdbcSelectExecutorStandardImpl.executeQuery(JdbcSelectExecutorStandardImpl.java:83)
    org.hibernate.sql.exec.spi.JdbcSelectExecutor.list(JdbcSelectExecutor.java:76)
    org.hibernate.sql.exec.spi.JdbcSelectExecutor.list(JdbcSelectExecutor.java:65)
    org.hibernate.query.sql.internal.NativeSelectQueryPlanImpl.performList(NativeSelectQueryPlanImpl.java:138)
    org.hibernate.query.sql.internal.NativeQueryImpl.doList(NativeQueryImpl.java:621)
    org.hibernate.query.spi.AbstractSelectionQuery.list(AbstractSelectionQuery.java:427)
    org.hibernate.query.spi.AbstractSelectionQuery.getSingleResult(AbstractSelectionQuery.java:564)
    org.springframework.data.jpa.repository.query.JpaQueryExecution$SingleEntityExecution.doExecute(JpaQueryExecution.java:223)
    org.springframework.data.jpa.repository.query.JpaQueryExecution.execute(JpaQueryExecution.java:92)
    org.springframework.data.jpa.repository.query.AbstractJpaQuery.doExecute(AbstractJpaQuery.java:149)
    org.springframework.data.jpa.repository.query.AbstractJpaQuery.execute(AbstractJpaQuery.java:137)
    org.springframework.data.repository.core.support.RepositoryMethodInvoker.doInvoke(RepositoryMethodInvoker.java:170)
    org.springframework.data.repository.core.support.RepositoryMethodInvoker.invoke(RepositoryMethodInvoker.java:158)
    org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.doInvoke(QueryExecutorMethodInterceptor.java:164)
    org.springframework.data.repository.core.support.QueryExecutorMethodInterceptor.invoke(QueryExecutorMethodInterceptor.java:143)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.data.projection.DefaultMethodInvokingMethodInterceptor.invoke(DefaultMethodInvokingMethodInterceptor.java:70)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.transaction.interceptor.TransactionInterceptor$1.proceedWithInvocation(TransactionInterceptor.java:123)
    org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:392)
    org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:119)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:137)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.invoke(CrudMethodMetadataPostProcessor.java:135)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
    org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184)
    org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:223)
    jdk.proxy2/jdk.proxy2.$Proxy107.executeSleep(Unknown Source)
    com.github.sinsuren.demo.controller.TestController.executeSleep(TestController.java:18)
    java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
    java.base/java.lang.reflect.Method.invoke(Method.java:580)
    org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:255)
    org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:188)
    org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:118)
    org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:926)
    org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:831)
    org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:87)
    org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1089)
    org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
    org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
    org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903)
    jakarta.servlet.http.HttpServlet.service(HttpServlet.java:564)
    org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
    jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)
    org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:206)
    org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
    org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
    org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:175)
    org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
    org.springframework.web.filter.RequestContextFilter.doFilterInternal(RequestContextFilter.java:100)
    org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
    org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:175)
    org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
    org.springframework.web.filter.FormContentFilter.doFilterInternal(FormContentFilter.java:93)
    org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
    org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:175)
    org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
    org.springframework.web.filter.CharacterEncodingFilter.doFilterInternal(CharacterEncodingFilter.java:201)
    org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
    org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:175)
    org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
    org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:167)
    org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:90)
    org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:482)
    org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:115)
    org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:93)
    org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:74)
    org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:344)
    org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:391)
    org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:63)
    org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:896)
    org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1736)
    org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:52)
    java.base/java.lang.VirtualThread.run(VirtualThread.java:311)
```
Below are the troublesome section, which are coming from mysql-connector-j.

```    
	com.mysql.cj.protocol.ReadAheadInputStream.read(ReadAheadInputStream.java:180) <== monitors:1

	com.mysql.cj.jdbc.ConnectionImpl.isValid(ConnectionImpl.java:2496) <== monitors:1
```

Fix for which is already getting published  here https://bugs.mysql.com/bug.php?id=110512

Based on finding, I will second @lfbayer's opinion to wait till connector issue is fixed, as above changes doesn't fall under any of the critical section or long wait time section.

godenji:
> Fix for which is already getting published here https://bugs.mysql.com/bug.php?id=110512
> 
> Based on finding, I will second @lfbayer's opinion to wait till connector issue is fixed, as above changes doesn't fall under any of the critical section or long wait time section.

@sinsuren  Connector/J 9.0 released with above referenced PR merged.

Am I correct in stating that Hikari is now the "bottleneck" given that drivers for Postgres, MySQL, etc. all seem to have de-`synchronize`d their respective code bases?

I understand there's still some question of how Hikari can seamlessly transition to the new world of virtual threads, but it would be fantastic if there was some movement in that direction -- all the pieces are now in place it seems...

sinsuren:
Thanks for notifying about the latest release of mysql-connector. I reran the service again with virtual thread pinning detection and no virtual thread pinning was detected.

With mysql-connector 9.x.x, there won't be any scalability issues for application having virtual thread enabled and connecting to  persistence layer using 9.x.x driver.

So should HikariCP should convert all synchronised to ReEntrant lock ?
That will be good to have(not must) , given that jdk haven't removed all synchronised instances from code base as well. The places where synchronized is being used in HikariCP, are places which doesn't come under major I/o operations path and majorly having computations only.


eg:

```java

public syncronized int largest(int [] inputArray) {
	int max = inputArray[0];
	
	for(int i = 1; i < inputArray.length; i++) {
		max = Math.max(max, inputArray[i]);
	} 

   return max;
}


public int largest(int [] inputArray) {

    reEntrantlock.lock();
    try {
		int max = inputArray[0];
		
		for(int i = 1; i < inputArray.length; i++) {
			max = Math.max(max, inputArray[i]);
		} 
	
	    return max;
   } finally {
	  reEntrantlock.unlock();
   }
}

```

Given that action executed in above function is computation which will be done with the help of platform thread only, there might be slight performance difference. 

> The results were clear and fairly unsurprising: synchronized() is substantially faster. In the single-threaded test, synchronized() was about 7.5x faster on average than Lock.lock(). In the two-threaded test, synchronized() was still the clear winner, about 2x faster on average.
from: https://medium.com/liveramp-engineering/java-performance-synchronized-vs-lock-301130e62f47

bdeneuter:
If no I/O occurs within the synchronized block then it is ok to leave the synchronized block as the carrier thread (platform thread) won't be released anyway. There is only a problem when I/O occurs within a synchronized block as there are only limited carrier threads and those carrier threads can not perform other work while being pinned.
I also tested our code for pinning and didn't detect any pinning anymore
fyi, I used `io.quarkus.junit5:junit5-virtual-threads` for testing pinning.
https://quarkus.io/guides/virtual-threads#testing-virtual-thread-applications

cfredri4:
> > The results were clear and fairly unsurprising: synchronized() is substantially faster. In the single-threaded test, synchronized() was about 7.5x faster on average than Lock.lock(). In the two-threaded test, synchronized() was still the clear winner, about 2x faster on average.
> > from: https://medium.com/liveramp-engineering/java-performance-synchronized-vs-lock-301130e62f47

This was written 2011, a **lot** of things has changed since then.

justEli:
When I moved from OpenJDK to GraalVM, HikariCP stopped working completely within virtual threads, without any exceptions. Spent hours trying to find out it was caused by this. I changed the executors to `Executors.newCachedThreadPool()`, and it started working again. Maybe someone finds this info useful.

juliojgd:
> I'm certainly not familiar with the hikariCP internals, but it sounds like there are two topics here:
> 
>     1. Allowing/enabling **upstream code** to use virtual-threads (w/o pinning). [...]
>     2. Allowing/configuring hikariCP **itself** to use virtual-threads (i.e. internally).  [...]
> Hope my understanding is correct, and that this helps someone 👍 . Thanks for the great work - eagerly awaiting on a resolution...

I agree with @dpiliouras-ut that this is a two faced issue. But this PR addresses at least the first point "Allowing/enabling **upstream code** to use virtual-threads (w/o pinning)." 

Given that MySQL connector for Java has fixed this pinning (and someone told that PR should be retained until MySQL fixes that), I think this PR should be merged as it contains the removal of  "synchronized" blocks and methods that does have I/O operations inside so they are prone to thread pinning when used with virtual threads. In addition [PostgreSQL java client has already made changes to be virtual threads friendly](https://jdbc.postgresql.org/changelogs/2023-03-17-42.6.0-release/).

So I think  this PR could improve HikariCP behavior (thread pinning) when using with Virtual Threads. 

**Is there any reason left as of today for not merging this PR?**

Thanks in advance for clarification

AllanConsolaro (approved):

bdeneuter:
fyi, the pinning problem will probably be solved in JDK 24. Then it will be fine to use the synchronized keyword:
https://openjdk.org/jeps/491

juliojgd:
@bdeneuter That JEP status is "Candidate". I think then it's not a "probably" but a "maybe". IMHO it probably will come in the future but I wouldn't bet for it in JDK24.

Using ReentranLocks will be OK in JDK24+ as it already is in JDK 21, 22 and 23.

So I think, any improvement must be done now (merged), because of the "maybe" nature of the candidate JEP you mentioned and the real current need for Virtual Threads.

Glad to see activity again in this PR. Any chance of merging it after your approval @AllanConsolaro ?

Thanks in advance

sdeleuze:
From a Spring perspective, I see a fair amount of added value making Hikari VT friendly as of Java 21, so +1 for that PR.

brettwooldridge:
This pull request is nonsense as far as I can tell. The existing code has zero performance impact on virtual threads and zero percent chance of "blocking" them. The only blocking that could occur would be down the stack from HikariCP in the database driver itself.

Changing this...
``ProxyStatement.java``
```java
public final void close() throws SQLException
{
   synchronized (this) {
      if (isClosed) {
         return;
      }

      isClosed = true;
   }
}
```
to this...
```java
public final void close() throws SQLException
{
   proxyStatementLock.lock();
   try {
      if (isClosed) {
         return;
      }

      isClosed = true;
   } finally {
      proxyStatementLock.unlock()
   }
}
```
...is meaningless.

First of all, this is a safety guard. Multiple threads calling the same ``Statement`` is not supported by JDBC. This just guarantees that even if a user does that the state is consistent. Second, this has zero chance of *blocking* a thread because the inherent code inside does not perform any blocking operation. Could a thread, virtual or not, be blocked for microseconds? Potentially, if two threads were violating the JDBC specification by accessing the same ``Statement`` instance at the same time. I simply don't care. This is not a real usecase.

XcrigX:
Is it the case that in thread pinning only occurs when multiple threads try to access a synchronized block concurrently, or can a single thread accessing a synchronized block lead to pinning? If it's the latter then it doesn't matter that the case can't actually occur, no?

cfredri4:
> Is it the case that in thread pinning only occurs when multiple threads try to access a synchronized block concurrently, or can a single thread accessing a synchronized block lead to pinning? If it's the latter then it doesn't matter that the case can't actually occur, no?

Thread pinning happens if a virtual thread does blocking (or some native) operations inside a synchronized block or method.
It's not a problem to have synchronized blocks/methods, as long as you don't do anything blocking in them.

The example @brettwooldridge shows above is indeed a meaningless change (no chance of blocking).
But if there's any chance of blocking down the stack from HikariCP or in the database driver itself, then it's not meaningless to use a real lock instead.

brettwooldridge:
Virtual thread pinning only occurs when a virtual thread persists execution inside of a ``synchronized`` block. All cited examples follow this pattern:
```java
synchronized(lock) {
   someLongRunningWork();
}
```
and the proposed solution is:
```java
try {
   lock.lock();
   someLongRunningWork();
} finally {
   lock.unlock();
}
```
However, there is still a cost to executing ``try...finally`` blocks, and _entering_ a reentrant lock via ``lock.lock()`` involves a virtual method invocation which is much more expensive than the native ``synchronized`` construct.

All of which makes applying that pattern blindly counterproductive to code like:
```java
synchronized(lock) {
   // bunch of non-blocking operations
}
```
This code is not susceptible to pinning and introducing ``try...finally`` and non-native (library-level not language level) locks only worsens performance for both virtual and non-virtual threads.

XcrigX:
Got it. Appreciate the explanations.

andreak:
Sorry for being late to the party here, but does this mean that HikariCP isn't affected by VT's "pinning issue" and is safe to with VTs?
Thanks.

cfredri4:
> fyi, the pinning problem will probably be solved in JDK 24. Then it will be fine to use the synchronized keyword: https://openjdk.org/jeps/491

> @bdeneuter That JEP status is "Candidate". I think then it's not a "probably" but a "maybe". IMHO it probably will come in the future but I wouldn't bet for it in JDK24.

That JEP has now been promoted to Proposed to Target.

cfredri4:
@brettwooldridge

> ...introducing `try...finally` and non-native (library-level not language level) locks only worsens performance for both virtual and non-virtual threads.

Have any tests been run to confirm this with a relatively recent version of Java? 
[This post](https://www.mo4tech.com/comparison-of-synchronized-and-reentrantlock-performance-in-java.html) claims that ReentrantLock performs _better_ than synchronized.

woj-tek:
> Have any tests been run to confirm this with a relatively recent version of Java?
> [This post](https://www.mo4tech.com/comparison-of-synchronized-and-reentrantlock-performance-in-java.html) claims that ReentrantLock performs _better_ than synchronized.

Hmm...

https://old.reddit.com/r/java/comments/1gmeeny/comparison_of_synchronized_and_reentrantlock/ seems to indicated that indeed `ReentrantLock` performs better than `synchronized`

donbeave:
Based on JEP 491 (https://openjdk.org/jeps/491):

```
We [previously recommended](https://openjdk.org/jeps/444#Pinning) solving frequent and long-lived pinning problems by migrating code from using synchronized to using ReentrantLock. Once the synchronized keyword no longer pins virtual threads, such migration will no longer be necessary. You need not revert code that has been migrated to use ReentrantLock back to using synchronized.
```

This JEP  proposed to target JDK 24. Maybe it's not worth fixing these. And wait for the release of OpenJDK 24.

juliojgd:
@donbeave But in JDK 24 (if it it's indeed included) it will be a Preview feature, not usable in Production. 

I think it's important to fix this now as the suggested JEP may  take years to be included as a G.A. feature in the JDK.

andreak:
@juliojgd Where do you see JEP-491 is a "preview feature"?

XcrigX:
I still see two potential issues here. 

**Issue 1:** potential for thread pinning - 
This is either : 
	A) currently NOT an issue with Hikari - from what brettwoolridge explained, pinning should not occur with the way Hikari uses synchronized blocks
	or
	B) will be fixed in a future JRE versions regardless

Even if pinning can occur, it doesn't stop the use of virtual threads in apps that use Hikari, it just might make them less performant than is possible? 

**Issue 2:** If you do use virtual threads in your app, there's not a "good" way to throttle Hikari to avoid overrunning the connection pool. As I understand it, the only thing possible today is to effectively disable the connection pool timeout and allow VTs to stack up waiting indefinitely for a connection. Maybe this isn't an issue? It doesn't quite feel right though.

cfredri4:
**Issue 3**: there are several examples above of ReentrantLock being higher performance than synchronized, meaning a change would be beneficial regardless of any thread pinning issues.

brettwooldridge:
@cfredri4 So far evidence using JMH benchmarks (the gold standard) is that ReentrantLock is substantially slower, by orders of magnitude.

(lower _ns/op_ is better)
Screenshot 2024-11-10 at 3 27 22
[source](https://stackoverflow.com/questions/77968278/why-acquiring-a-locked-lock-is-slower-when-using-jdk21-compared-to-jdk11)

In fact, uncontended locks, as all of the proposed changes were, are essentially transparent. This is because object locks have been continually optimized in the JVM for the past 25 years. Java employs biased locks that escalate to lightweight locks that escalate to heavyweight locks under different levels of contention.

Screenshot 2024-11-10 at 3 51 23

In short, it's complicated. If we had code sections subject to long running code that needed to execute while holding a contended lock, we would certainly employ ``ReentrantLock``. _We don't have those as they were intentionally avoided._

cfredri4:
@brettwooldridge not saying you are wrong, but the test you link to definitely is. 😉

> As a side notice, your syncLock() benchmark does not actually measure performance of synchronized, since locking on a local non-escaped object is a no-op, and JIT compiler happily eliminates unnecessary locking altogether.

slovdahl:
> In fact, uncontended locks, as all of the proposed changes were, are essentially transparent. This is because object locks have been continually optimized in the JVM for the past 25 years. Java employs biased locks that escalate to lightweight locks that escalate to heavyweight locks under different levels of contention.

Also not fully disputing everything you say here, but it's worth pointing out that "biased locking" has been disabled by default since JDK 15 (https://openjdk.org/jeps/374). So, uncontended `synchronized` is not necessarily as cheap as it used to be.

woj-tek:
> So far evidence using JMH benchmarks (the gold standard) is that ReentrantLock is substantially slower, by orders of magnitude.
> 
> (lower _ns/op_ is better)

Hmm... In the thread on reddit (https://old.reddit.com/r/java/comments/1gmeeny/comparison_of_synchronized_and_reentrantlock/lw28x2c/) Slanec got slightly different results:

Captura de pantalla 2024-11-10 a las 11 08 55

And in side comment Ron Pressler (@pron ?) gives this explanation:

> I would say that they have somewhat different performance characteristics and neither one dominates the other in all situations (and if one happens to work better than the other in one version of the JDK, it may work worse in another). In new code, I would recommend using ReentrantLock (or even the more sophisticated StampedLock where appropriate) if only because that implementation is under heavier maintenance and improvement work [1], but there's no need to migrate old code from synchronized to ReentrantLock unless there's a clear issue (such as pinning virtual threads when synchronized guards an IO operation in JDKs prior to 24).
> 
> Just note that one important difference between a native monitor (`synchronized`) and ReentrantLock is that the former does some spinning before blocking while the latter doesn't, as spinning is sometimes advantageous and sometimes harmful, depending on the amount of spinning. If spinning is desired, you can do it manually with a loop of `tryLock`.
> 
> [1]: For this reason we may someday change `synchronized` so that it uses ReentrantLock or something very similar to it in its implementation.

slovdahl:
I took some time to re-run the benchmark in https://stackoverflow.com/questions/77968278/why-acquiring-a-locked-lock-is-slower-when-using-jdk21-compared-to-jdk11. On my AMD Ryzen 7 8845HS I got the following results:

JDK 11:
```
openjdk version "11.0.25" 2024-10-15
OpenJDK Runtime Environment Temurin-11.0.25+9 (build 11.0.25+9)
OpenJDK 64-Bit Server VM Temurin-11.0.25+9 (build 11.0.25+9, mixed mode)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  7.103 ± 0.082  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.312 ± 0.016  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.415 ± 0.002  ns/op
LockNoLockBenchmark.syncLock              avgt   10  0.420 ± 0.002  ns/op
```

JDK 17:
```
openjdk version "17.0.13" 2024-10-15
OpenJDK Runtime Environment Temurin-17.0.13+11 (build 17.0.13+11)
OpenJDK 64-Bit Server VM Temurin-17.0.13+11 (build 17.0.13+11, mixed mode, sharing)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  8.263 ± 0.099  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.391 ± 0.022  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.410 ± 0.007  ns/op
LockNoLockBenchmark.syncLock              avgt   10  0.419 ± 0.005  ns/op
```

JDK 21:
```
openjdk version "21.0.5" 2024-10-15 LTS
OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.5+11 (build 21.0.5+11-LTS, mixed mode, sharing)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  8.277 ± 0.044  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.305 ± 0.030  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.354 ± 0.004  ns/op
LockNoLockBenchmark.syncLock              avgt   10  0.360 ± 0.002  ns/op
```

After that, I made one small change to the benchmark that avoids optimizing away `synchronized` in the `syncLock` benchmark:

```diff
diff --git src/main/java/org/example/LockNoLockBenchmark.java src/main/java/org/example/LockNoLockBenchmark.java
index 1ec4807..be25279 100644
--- src/main/java/org/example/LockNoLockBenchmark.java
+++ src/main/java/org/example/LockNoLockBenchmark.java
@@ -54,6 +54,7 @@ public class LockNoLockBenchmark {
     int counter;
 
     ReentrantLock lock = new ReentrantLock();
+    Object lockObject = new Object();
 
     @Benchmark
     public void noLock() {
@@ -62,7 +63,7 @@ public class LockNoLockBenchmark {
 
     @Benchmark
     public void syncLock() {
-        synchronized (new Object()) {
+        synchronized (lockObject) {
             ++counter;
         }
     }
```

Re-running the benchmarks gave me this:

JDK 11:
```
openjdk version "11.0.25" 2024-10-15
OpenJDK Runtime Environment Temurin-11.0.25+9 (build 11.0.25+9)
OpenJDK 64-Bit Server VM Temurin-11.0.25+9 (build 11.0.25+9, mixed mode)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  7.108 ± 0.074  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.323 ± 0.026  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.414 ± 0.006  ns/op
LockNoLockBenchmark.syncLock              avgt   10  1.101 ± 0.014  ns/op
```

JDK 17:
```
openjdk version "17.0.13" 2024-10-15
OpenJDK Runtime Environment Temurin-17.0.13+11 (build 17.0.13+11)
OpenJDK 64-Bit Server VM Temurin-17.0.13+11 (build 17.0.13+11, mixed mode, sharing)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  8.355 ± 0.019  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.373 ± 0.010  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.410 ± 0.004  ns/op
LockNoLockBenchmark.syncLock              avgt   10  3.920 ± 0.019  ns/op
```

JDK 21:
```
openjdk version "21.0.5" 2024-10-15 LTS
OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.5+11 (build 21.0.5+11-LTS, mixed mode, sharing)

Benchmark                                 Mode  Cnt  Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  8.251 ± 0.049  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  3.349 ± 0.014  ns/op
LockNoLockBenchmark.noLock                avgt   10  0.362 ± 0.011  ns/op
LockNoLockBenchmark.syncLock              avgt   10  4.149 ± 0.022  ns/op
```

Let me summarize the relevant parts:

```
LockNoLockBenchmark.lockUnlock            avgt   10  3.323 ± 0.026  ns/op # JDK 11
LockNoLockBenchmark.lockUnlock            avgt   10  3.373 ± 0.010  ns/op # JDK 17
LockNoLockBenchmark.lockUnlock            avgt   10  3.349 ± 0.014  ns/op # JDK 21
LockNoLockBenchmark.syncLock              avgt   10  1.101 ± 0.014  ns/op # JDK 11 - with biased locking
LockNoLockBenchmark.syncLock              avgt   10  3.920 ± 0.019  ns/op # JDK 17 - no biased locking
LockNoLockBenchmark.syncLock              avgt   10  4.149 ± 0.022  ns/op # JDK 21 - no biased locking
```

juliojgd:
> @juliojgd Where do you see JEP-491 is a "preview feature"?

The status of that JEP is  "Targeted" to Release 24, it  means it will be delivered in JDK 24, but it does not mean that it will be a Final feature (so usable without preview flags). The more likely scenario is that the first instance of a feature will be released as "Preview". No assurances but I think the usual path for a feature to get "Final" is to be released as "Preview" in one (or several) prior  releases.

So we can't rely on that JEP being final in JDK 24.

slovdahl:
> The status of that JEP is "Targeted" to Release 24, it means it will be delivered in JDK 24, but it does not mean that it will be a Final feature (so usable without preview flags). The more likely scenario is that the first instance of a feature will be released as "Preview". No assurances but I think the usual path for a feature to get "Final" is to be released as "Preview" in one (or several) prior releases.
> 
> So we can't rely on that JEP being final in JDK 24.

If a feature is a "Preview feature" it is _always_ mentioned in the JEP itself. See e.g. https://openjdk.org/jeps/495 that is called "JEP 495: Simple Source Files and Instance Main Methods (Fourth Preview)". The summary also ends with "This is a [preview language feature](https://openjdk.org/jeps/12).". So, I'm pretty sure https://openjdk.org/jeps/491 is fully finalized in JDK 24.

cfredri4:
> Let me summarize the relevant parts:
> 
> ```
> LockNoLockBenchmark.lockUnlock            avgt   10  3.323 ± 0.026  ns/op # JDK 11
> LockNoLockBenchmark.lockUnlock            avgt   10  3.373 ± 0.010  ns/op # JDK 17
> LockNoLockBenchmark.lockUnlock            avgt   10  3.349 ± 0.014  ns/op # JDK 21
> LockNoLockBenchmark.syncLock              avgt   10  1.101 ± 0.014  ns/op # JDK 11 - with biased locking
> LockNoLockBenchmark.syncLock              avgt   10  3.920 ± 0.019  ns/op # JDK 17 - no biased locking
> LockNoLockBenchmark.syncLock              avgt   10  4.149 ± 0.022  ns/op # JDK 21 - no biased locking
> ```

So tl;dr; _in this specific test_ by @slovdahl, for JDK >=17 then ReentrantLock is faster than synchronized.

brettwooldridge:
@slovdahl I get different results:
```
openjdk version "21.0.2" 2024-01-16 LTS
OpenJDK Runtime Environment Zulu21.32+17-CA (build 21.0.2+13-LTS)
OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)

Benchmark                                 Mode  Cnt   Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  41.577 ± 0.120  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  12.931 ± 0.050  ns/op
LockNoLockBenchmark.noLock                avgt   10   0.714 ± 0.012  ns/op
LockNoLockBenchmark.syncLock              avgt   10   9.634 ± 0.071  ns/op
```

This only change I made was to make the lock objects ``private final ...``

> Escape Analysis and Optimization: The JVM relies on the lock object being stable, especially in scenarios where it might be shared across threads. If the lock object is not final, the JVM must assume that it could be reassigned at any point. This could prevent certain optimizations, as the JVM would need to check that the lock reference hasn't changed each time it enters the synchronized block, adding some overhead.

NOTE: Java 21 disables biased locking by default, so the above is with no biased locking.

UPDATE: This is likely to be architecture dependent. For example get this on one of our Epyc servers:

```
openjdk version "21.0.2" 2024-01-16 LTS
OpenJDK Runtime Environment Zulu21.32+17-CA (build 21.0.2+13-LTS)
OpenJDK 64-Bit Server VM Zulu21.32+17-CA (build 21.0.2+13-LTS, mixed mode, sharing)

Benchmark                                 Mode  Cnt   Score   Error  Units
LockNoLockBenchmark.lockLockUnlockUnlock  avgt   10  26.200 ± 0.057  ns/op
LockNoLockBenchmark.lockUnlock            avgt   10  10.104 ± 0.011  ns/op
LockNoLockBenchmark.noLock                avgt   10   2.110 ± 0.002  ns/op
LockNoLockBenchmark.syncLock              avgt   10  10.007 ± 0.021  ns/op
```
The first run (at the top) is on an AWS Graviton server.

Obviously, the JVM version (11, 17, 21, etc) also makes a big difference due to the difference in the default of biased or non-biased locking.

pron:
1. JEP 491 is not Preview.
2. When looking at microbenchmarks, it is very important to understand what they're measuring. They're not measuring the general performance of the locking construct, but only the performance of the locking construct *in the context of the microbenchmark*. For one, different locking constructs can behave differently depending on the level of contention, and so a "better" one is the one that's better for the level of contention that's relevant for the application. More specifically, `synchronized` has a different spinning behaviour than that of ReentrantLock, but the appropriate amount of spinning for the relevant contention behaviour can be set up for a ReentrantLock. It doesn't make sense to compare `synchronized` to RL when the RL is not used with a similar spinning strategy.

brettwooldridge:
@pron This was kinda my point before we got off in the weeds of benchmarks. None of the methods in question for this proposed change are even valid to be called simultaneously. For example, ``ResultSet.close()`` on a single instance of a ``ResultSet``. In the real world there should never be contention. It would be an application error and violation of the JDBC specification to simultaneously call one of these methods on the same instance from multiple threads.

And to your point, microbenchmarks are not real world. For example, in real world use every Connection, Statement, ResultSet etc. would contain an additional object (``ReentrantLock``) that must be constructed for each instance. So, first this is additional construction overhead compared to synchronizing on ``this``. Second, each ReentrantLock instance must be garbage collected when a ResultSet, Statement, or Connection is closed and disposed of. Obviously, Connections are long lived in the case of a pool, but Statements are ResultSets are constructed by the thousands per second in a large application server. The number of garbage objects to collect is 2x when introducing a standalone lock object inside of each instance.

On top of everything, none of the methods this ReentrantLock is proposed for are capable of virtual thread pinning, so this doesn't even rise to the level of an acedemic discussion.

slovdahl:
I didn't make it clear in my previous posts, but my point was not to suggest that `synchronized` should be replaced with `ReentrantLock`, and even less so that one in general is superior to the other. No need for me to repeat what @brettwooldridge and @pron said above :point_up: I just could not resist the urge to make the the benchmark from the SO post a tiny little bit less bad. :see_no_evil:

cfredri4:
JEP-491 is now status Integrated. So it's 100% coming in JDK 24.

PR #1661, opened 2020-09-23
Make generated proxy class final

quaff:
The final modifier is overridden by public modifier before this commit

PR #2068, opened 2023-05-10
Removes final modifier from generated Proxy classes.

cccs-ian:
**Description**

This PR removes the final modifier from the generated proxy classes, allowing clients code to extend them. This PR relates to some changes made here: https://github.com/brettwooldridge/HikariCP/pull/1661 where a small bug was fixed associated with the codegen.

**Reasoning**

My team has extended the HikariProxyResultSet in older versions of HikariCP, to allow us to package up stuff like warnings into the ResultSet implementation. I would be open to the suggestion that only the HikariProxyResultSet be made non-final, and be willing to change the PR to match if is decided that this is the best course of action.

Thanks for your consideration.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/JavassistProxyFactory.java
+++ b/src/main/java/com/zaxxer/hikari/util/JavassistProxyFactory.java
@@ -108,7 +108,7 @@ public final class JavassistProxyFactory
 
       var superCt = classPool.getCtClass(superClassName);
       var targetCt = classPool.makeClass(newClassName, superCt);
-      targetCt.setModifiers(Modifier.setPublic(Modifier.FINAL));
+      targetCt.setModifiers(Modifier.PUBLIC);
 
       System.out.println("Generating " + newClassName);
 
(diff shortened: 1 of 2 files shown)


cccs-ian:
Hoping to start  conversation about this PR :)

brettwooldridge:
This seems super dangerous to me. Allowing users to extend the generated classes opens up the possibility of altering the commitments that they make to the pool

I understand the "buyer beware" philosophy and just saying the user is responsible if they break things, but ultimately this can generate crazy issues opened up against the pool for some odd behavior when we have no idea that the user is doing something wrong in their code. I've seen this kind of thing lead to dozens of hours of investigation only to find that there was no issue with the underlying code.

This is one of the things that lead to, for example, the pool configuration being sealed once the pool starts. Before that, users were modifying pool limits and timeouts while the pool itself was running causing utter havoc. Only after hours and hours investigation and long discussion threads did it come out what was going on.

It's going to take a lot to convince me that this PR is a sane thing to do.

I would suggest wrapping/extending the driver classes and providing the wrapper as the datasource to HikariCP...

cccs-ian:
Thanks for your advice and consideration! 

I will investigate working with the driver class as suggested.

PR #2095, opened 2023-07-29
HikariConfig comment typo

m1a2st:
fix comment typo in HikariConfig


StravitoMike (approved):

lfbayer:
I've merged this with commit 2883f846544f88ef505d6237b0cf67131a7323c1. The "may be" is actually correct as it is used in the sense of "allowed to be" rather than "might happen to be".

PR #2115, opened 2023-09-29
Replaced the synchronized use of `System.getProperties` with AtomicInteger in pool name generation

### LINKED ISSUE #2104: Synchronizing System.getProperties in generatePoolName / Number creates a deadlock-able window.
Since Hikari is a library, it should not rely on synchronizing on any objects not within it's own scope, Interactions between code either more shallow or deeper in the stack can create opportunities for a deadlock.  It certainly shouldn't synchronize on on the System.getProperties() Properties object.

The code appears to facilitate sharing of the pool name across classloaders.  This could be obviated by using the hashcode of a specific bootloader loaded class.

https://github.com/brettwooldridge/HikariCP/blob/2e07be7b325b05f7225611c60f421e253e1107a5/src/main/java/com/zaxxer/hikari/HikariConfig.java#L1169

echo724:
Fix: #2104 

The objective of using synchronized was to avoid overlapping in the generation of HikariPool numbers, likely in sequential operations following a call to System.getProperties. However, synchronizing these operations could potentially lead to deadlocks. Therefore, I considered an alternative that does not rely on synchronized access to the System properties. Using AtomicInteger for Pool number generation eliminates the risk of overlap without the deadlock risk associated with synchronized blocks.

1. **Reason for using AtomicInteger:** ﻿
AtomicInteger serves as a JVM level shared value to handle concurrency safely while generating unique numbers within a single JVM. This ensures that each ﻿HikariConfig instance is assigned a unique pool name, even in situations where multiple threads attempt to generate pool names simultaneously.

2. **Reason for using AtomicInteger over UUID:** 
While UUID does provide uniqueness, it comes at a higher generation cost and complexity. Conversely, using ﻿AtomicInteger allows for relatively lightweight and swift generation of unique numbers in asynchronous and concurrent environments. Furthermore, using ﻿AtomicInteger alleviates the need to handle synchronization during pool name generation, thus reducing code complexity while addressing concurrent access issues.

The generation code is as follows:

```java
// Added priavte static field POOL_NUMBER to track pool number across the instances
private static final AtomicInteger POOL_NUMBER = new AtomicInteger(0);
...
private String generatePoolName() {
    final var prefix = "HikariPool-";
    // The pool number will be incremented atomically
    final var next = String.valueOf(poolNumber.incrementAndGet());
    return prefix + next;
}
...
```

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -29,13 +29,12 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.lang.reflect.Modifier;
-import java.security.AccessControlException;
 import java.sql.Connection;
 import java.util.Properties;
 import java.util.TreeSet;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.zaxxer.hikari.util.UtilityElf.getNullIfEmpty;
 import static com.zaxxer.hikari.util.UtilityElf.safeIsAssignableFrom;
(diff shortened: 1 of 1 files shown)


lfbayer (changes requested):
This fix is insufficient as it doesn't handle the case where there are multiple classloaders loading their own hikaricp classes. In that case you would have multiple instances of POOL_NUMBER (one per classloader).

brettwooldridge:
@lfbayer is correct, and that is why we use a system property, as unconventional as that is.

PR #2079, opened 2023-05-31
Added Support For beginRequest and endRequest

Deeptanshu-Chowdhuri:
#2071 Added support of beginRequest and endRequest from JDBC 4.3 onwards.

- [beginRequest](https://docs.oracle.com/javase/9/docs/api/java/sql/Connection.html#beginRequest--) hints to the driver that a request, an independent unit of work, is beginning on this connection.
- [endRequest](https://docs.oracle.com/javase/9/docs/api/java/sql/Connection.html#endRequest--) hints to the driver that a request has completed.

Both these calls function to let the driver know when a connection is checked out of the pool and back into the pool.These boundaries can help driver for load balancing and other stuff. These functions are driver specific and may not do anything.

PR #2126, opened 2023-10-24
Added Support For beginRequest and endRequest (new version)

fmeheust:
Continued work from pull request https://github.com/brettwooldridge/HikariCP/pull/2079. Addressed review comments https://github.com/brettwooldridge/HikariCP/pull/2079#discussion_r1367934838 and added tests.


jeandelavarene (approved):

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
```diff
@@ -56,6 +56,7 @@ final class PoolEntry implements IConcurrentBagEntry
 
    private final boolean isReadOnly;
    private final boolean isAutoCommit;
+   private boolean isJDBC43OrLater;
```
akashchouhan16: The choice of variable is inconsistent with the codebase

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
```diff
@@ -70,6 +71,17 @@ final class PoolEntry implements IConcurrentBagEntry
       this.isAutoCommit = isAutoCommit;
       this.lastAccessed = currentTime();
       this.openStatements = new FastList<>(Statement.class, 16);
+      boolean isJDBC43OrLater = false;
```
akashchouhan16: any specific reason to use `isJDBC43OrLater` inside the static initializer block? Could this be moved?

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
```diff
@@ -69,6 +71,17 @@ final class PoolEntry implements IConcurrentBagEntry
...
+      try {
+         DatabaseMetaData dm = connection.getMetaData();
+         isJDBC43OrLater = ((dm != null) && (
```
lfbayer: If the connection instance has the methods "beginRequest" and "endRequest", wouldn't that be sufficient to know we can call them, rather than doing a version check?

Maybe using reflection on the connection instance?

I'm wondering if adding the call to getMetaData has any performance implications.

It also seems like it would better to set this pool-wide, rather than have the value calculated for every new connection in the pool. I can't imagine that you would have a pool that would have different JDBC versions for each connection.

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
lfbayer: I imagine there is a desired to make the member variable `final` by having only a single line actually set the value. But this code still has a non-final member.

lfbayer (changes requested):
It feels like moving this code to HikariPool getConnection an recycle might be a little better.

I also wonder if we need to make sure there is a way to explicitly disable having the pool call the begin and end methods.

src/main/java/com/zaxxer/hikari/pool/PoolEntry.java
fmeheust: Thank you for the feedback. I chose to do it for each connection because I thought that one could change the version of the driver that is in the application classpath and the pool would start using a new driver, but if that is not possible I propose the following approach:

I will move the check to PoolBase.checkDriverSupport(Connection) and store the value(s) in a private member. This/these value(s) will be added as a parameter on the constructor of the PoolEntry class, and passed in the PoolBase.newEntry() method.

Concerning using relection instead of database metadata, I could argue that if Hikari starts using new JDBC functionality in the future, the code could end up with several boolean variables (one for each method present in from specific version of the JDBC standard) instead of to int values jdbcMajorVersion and jdbcMinorVersion. I am new to Hikari, so please let me know which choice you prefer and I will comply.

fmeheust:
Given that this is a Java 11+ version of HikariCP and that request boundaries (Connection.beginRequest() and Connection.endRequest()) are present since Java 9, I have removed the check for JDBC version.
I have added a property that allows to explicitly enable request boundaries. Calls to Connection.beginRequest() and Connection.endRequest() have been moved to HikariPool.getConnection and HikariPool.recycle respectively.

fmeheust:
@lfbayer Can you please review my latest changes to this pull request ?

jeandelavarene:
@lfbayer could you please review this request? Thank you!

krismohan:
Hi @lfbayer any update on the merge? Thanks!

brettwooldridge:
Manually resolved conflicts. Merged.

PR #2139, opened 2023-11-08
Changed the log level of `final void logPoolState(String... prefix)` from `debug` to `info`.

1chz:
The housekeeper's record of the current status of total, active, idle and waiting connections in the DBCP is highly useful when operating a product targeting real users. However, to view this log, specific log level settings need to be configured, and the relevant information is being shared. Our company also conducts this setup separately to reference the related logs. 

What do you think about changing this log level from debug to info?

Please refer to this Stack Overflow question and answer for more details.

https://stackoverflow.com/questions/60757590/logging-hikaricp-spring-boot
```properties
logging.level.com.zaxxer.hikari.HikariConfig=DEBUG 
logging.level.com.zaxxer.hikari=TRACE
```

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -395,17 +395,15 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    // ***********************************************************************
 
    /**
-    * Log the current pool state at debug level.
+    * Log the current pool state.
     *
     * @param prefix an optional prefix to prepend the log message
     */
    void logPoolState(String... prefix)
    {
-      if (logger.isDebugEnabled()) {
-         logger.debug("{} - {}stats (total={}, active={}, idle={}, waiting={})",
-                      poolName, (prefix.length > 0 ? prefix[0] : ""),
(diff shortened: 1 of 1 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -395,17 +395,15 @@ public synchronized void resumePool()
...
     * @param prefix an optional prefix to prepend the log message
     */
    void logPoolState(String... prefix)
    {
-      if (logger.isDebugEnabled()) {
-         logger.debug("{} - {}stats (total={}, active={}, idle={}, waiting={})",
-                      poolName, (prefix.length > 0 ? prefix[0] : ""),
-                      getTotalConnections(), getActiveConnections(), getIdleConnections(), getThreadsAwaitingConnection());
-      }
+      logger.info("{} - {}stats (total={}, active={}, idle={}, waiting={})",
```
gary258796, line 404: I think if really needs to be change to info, it's better to still have infoEnabled check since those info are all getting by one by one function, not simple parameter.

lfbayer:
If you are really interested in monitoring pool state it would be better to use one of the metrics trackers (ie: dropwizard, micrometer, or prometheus).

I think that a large percentage of users would complain if this were suddenly logged at info. The only time you really need this information in the logs specifically is if there is a problem with your application and the way it is configured.

I will not be merging this change.

PR #2146, opened 2023-11-25
ProxyConnection should close when ProxyStatement close

soosue:
If a connection is not closed, the PoolEntry remains in the IN_USE state, leading to a connection leak.


quaff:
`Connection` 1:n `Statement`, I don't understand why `Connection` should be closed when `Statement` close.

lfbayer:
@quaff I agree. This pull request is definitely not correct.

The connection absolutely should not be closed when a statement is closed as that would, along with potentially thousands of other problems, prevent users from using multiple statements for a single connection.

@soosue, what you want to do is close your connection in your application code explicitly after you have finished with it, you should not expect a statement close to close a connection for you. A try-with-resources block with the connection is the most idiomatic way to achieve this.

I'm marking this as closed.

PR #2155, opened 2023-12-06
Duplicate connection in try with resources clause removed (Fixes #2152)

### LINKED ISSUE #2152: PoolBase.quietlyCloseConnection() calls close twice
I'm using a Proxy wrapper around a Hikari pool, counting connections created and destroyed.

I noticed that my destroy count seemed to be off, and on further inspection it turned out that it was doubled.

Looks like the problem is in the try with resources clause, where ```connection``` is specified twice, resulting in it being closed twice:

```
void quietlyCloseConnection(final Connection connection, final String closureReason)
{
   if (connection != null) {
      try {
         logger.debug("{} - Closing connection {}: {}", poolName, connection, closureReason);

         // continue with the close even if setNetworkTimeout() throws
         try (connection; connection) { // <<<<<<<<
            if (!connection.isClosed())
               setNetworkTimeout(connection, SECONDS.toMillis(15));
         } catch (SQLException e) {
            // ignore
         }
      }
      catch (Exception e) {
         logger.debug("{} - Closing connection {} failed", poolName, connection, e);
      }
   }
}
```
This doesn't really cause any problems, since ```ProxyConnection.close()``` doesn't do anything if the connection is already closed, but I figured I'd point this out anyway. I've added a similar guard around my count method.

I'm ready to create a pull request if required.

bjorndarri:
Fixes #2152

Duplicate connection in try with resources clause caused close() being called twice on each connection.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
+++ b/src/main/java/com/zaxxer/hikari/pool/PoolBase.java
@@ -133,7 +133,7 @@ abstract class PoolBase
             logger.debug("{} - Closing connection {}: {}", poolName, connection, closureReason);
 
             // continue with the close even if setNetworkTimeout() throws
-            try (connection; connection) {
+            try (connection) {
                if (!connection.isClosed())
                   setNetworkTimeout(connection, SECONDS.toMillis(15));
                } catch (SQLException e) {


lfbayer:
Can you add a unit test for this change?

I actually don't know what the duplicate connection is in there. It looks oddly intentional, but removing it seems like a good idea.

quaff:
It's an obvious typo fix, It's safe enough if it doesn't fail other tests.

PR #2166, opened 2024-01-30
Small (even micro) improvements

anavrotski:
Regex removed where possible and some other polishing


brettwooldridge:
I cherry picked the majority of these but left a few.

PR #2189, opened 2024-03-26
add support to get and set db credentials in an atomic operation

benapple:
This pull request attempts to address points raised in https://github.com/brettwooldridge/HikariCP/pull/1442#issuecomment-1287697706, https://github.com/brettwooldridge/HikariCP/pull/2011#issuecomment-1336197290, and https://github.com/brettwooldridge/HikariCP/pull/1196#issuecomment-436720123 about rotating both username and password atomically.

Currently, you can call the setUsername and setPassword methods on the MBean or subclass the HikariDataSource to dynamically fetch credentials. Either way this is done presents a (tiny) window where the credentials used to connect may be in flux. In the case of updating via the MBean, a new connection may be created in between the call to setUsername and setPassword. And when subclassing the data source to dynamically provide username and password, the credentials may have changed between PoolBase's call to getUsername and getPassword.

To close these windows, I have introduced a new Credentials pojo that is essentially an immutable pair of username and password and replaced the HikariConfig's username and password fields with an AtomicReference to a Credential. It should be noted that even with these changes, you are still able to individually get and set the username and password, however if you need things to be atomic you should make use of the new API in HikariConfig, getCredentials and setCredentials. PoolBase now makes use of the getCredentials to atomically get the pair. Additionally, there is an extra method on the HikariConfigMXBean to atomically set the pair.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -18,6 +18,7 @@ package com.zaxxer.hikari;
 
 import com.codahale.metrics.health.HealthCheckRegistry;
 import com.zaxxer.hikari.metrics.MetricsTrackerFactory;
+import com.zaxxer.hikari.util.Credentials;
 import com.zaxxer.hikari.util.PropertyElf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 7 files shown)


IRus:
@benapple, is `AtomicReference` necessary here? Since it’s not using CAS, you could consider using var handles or just @Volatile instead.

PR #2223, opened 2024-07-30
Update micrometer

kziel1:
Needed to streamline premetheus versions in the company I work for. Migration guide that could be needed for some use cases: https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide

### DIFF
--- a/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
+++ b/src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
@@ -16,20 +16,22 @@
 
 package com.zaxxer.hikari.metrics.prometheus;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
 import com.zaxxer.hikari.metrics.PoolStats;
+
 import io.prometheus.metrics.model.registry.MultiCollector;
 import io.prometheus.metrics.model.snapshots.GaugeSnapshot;
 import io.prometheus.metrics.model.snapshots.Labels;
 import io.prometheus.metrics.model.snapshots.MetricMetadata;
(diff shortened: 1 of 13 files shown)


kziel1:
Changes suggested in the issue https://github.com/brettwooldridge/HikariCP/issues/2225
Generating a sample value hasn't been translated, hence tests not compiling.

src/main/java/com/zaxxer/hikari/metrics/prometheus/HikariCPCollector.java
```diff
@@ -63,14 +61,13 @@ void remove(String name)
...
+   private GaugeSnapshot createGauge(String metric, String help,
+                                                            Function<PoolStats, Integer> metricValueFunction)
    {
-      var metricFamily = new GaugeMetricFamily(metric, help, LABEL_NAMES);
-      poolStatsMap.forEach((k, v) -> metricFamily.addMetric(
-         Collections.singletonList(k),
-         metricValueFunction.apply(v)
+      Collection<GaugeSnapshot.GaugeDataPointSnapshot> gaugeDataPointSnapshots = new ArrayList<>();
+      poolStatsMap.forEach((k, v) -> gaugeDataPointSnapshots.add(
+         new GaugeSnapshot.GaugeDataPointSnapshot(metricValueFunction.apply(v), Labels.of(k), null)
```
henriAbel: `Labels.of(k)` expects at least 2 arguments.
I think correct would be `Labels.of("pool", k)`

kziel1: makes sense, I pushed the change suggestion

brettwooldridge:
Would this change break existing 1.5 micrometer users? It appears like it would.

kziel1:
Yes, it will break Micrometer because of breaking changes in Prometheus (and simpleclient).
https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide

PR #2229, opened 2024-08-13
 healthcheck registry can be set even if pool is sealed

rahulbhawsar:
Hi All, 

Request you to Please review this PR. 

As of today `HealthCheckRegistry` this property can be set only via programmatic configuration or IoC container.
Since, for the user who are using the `hibernate-hikaricp` and getting the data source from `HikariCPConnectionProvider` by unwrap this option is not available, please find example below - 

 ```
HikariCPConnectionProvider hikariCPConnectionProvider = sessionFactory.getSessionFactoryOptions()
                .getServiceRegistry()
                .getService(HikariCPConnectionProvider.class);
        HikariDataSource dataSource = hikariCPConnectionProvider.unwrap(HikariDataSource.class);

```

Returned `datasource` is `sealed`. 

I believe setting healthcheck should be allowed once! 

Please suggest if you think there exists another approach for such scenarios. 

thanks,

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -698,7 +698,6 @@ public class HikariConfig implements HikariConfigMXBean
     */
    public void setHealthCheckRegistry(Object healthCheckRegistry)
    {
-      checkIfSealed();
 
       if (healthCheckRegistry != null) {
          healthCheckRegistry = getObjectOrPerformJndiLookup(healthCheckRegistry);

### CODE REVIEW

**Comment by IRus** [comment]
@rahulbhawsar why this can't be done in `HikariCPConnectionProvider`?

**Comment by rahulbhawsar** [comment]
Hi @IRus  , 
I have thought of this place as it will be consistent with method `setMetricRegistry` as we one can set MetricRegistry once, after it has created.
Please suggest more on how you want to do this in `HikariCPConnectionProvider` class, if this is more logical, I will work on this. 
thanks,

**Comment by IRus** [comment]
@rahulbhawsar I'm not an expert in hibernate, have to use it many years ago, but looking at docs it seems you need to provide own implementation of `ConnectionProvider` which would use `HikariDataSource`, and create `SessionFactory` Bean using this custom `ConnectionProvider`, i.e:

```
import org.hibernate.SessionFactory
import org.hibernate.cfg.Configuration
import org.hibernate.service.ServiceRegistry
import org.hibernate.boot.registry.StandardServiceRegistryBuilder
import org.hibernate.service.jdbc.connections.spi.ConnectionProvider

object HibernateUtil {

    fun getSessionFactory(): SessionFactory {
            val configuration = Configuration().configure()
            // Add any configuration properties or mappings here

            val registryBuilder = StandardServiceRegistryBuilder()
            registryBuilder.applySettings(configuration.properties)

            // Create a custom ConnectionProvider instance
            val connectionProvider: ConnectionProvider = CustomConnectionProvider()

            // Add it to the registry
            registryBuilder.addService(ConnectionProvider::class.java, connectionProvider)

            val serviceRegistry: ServiceRegistry = registryBuilder.build()

            return configuration.buildSessionFactory(serviceRegistry)
    }
}
```

brettwooldridge:
This is allowed because ``HikariConfigMXBean`` allows the registry to be set after the pool has constructed. However, the MXBean enforces that the registry cannot be set multiple times. Unfortunately, there is no way to seal this part of the configuration and still have the MXBean still function.

PR #2238, opened 2024-09-16
Handle SQLTimeoutException without eviction

### LINKED ISSUE #1388: SQLTimeoutException closes connection
#### Environment
```
HikariCP version: 3.3.1
JDK version     : 1.8.0_201
Database        : SQL Server
Driver version  : 7.2.2.jre8
```
If executing a java.sql.Statement results in a SQLTimeoutException then  Hikari loggs a warning and force closes the connection.

```
WARN [pool-4-thread-30] - HikariPool-28 - Connection ConnectionID:10 ClientConnectionId: f3925681-0baf-4cc2-9a16-39ae212b92e8 marked as broken because of SQLSTATE(HY008), ErrorCode(0) [com.zaxxer.hikari.pool.ProxyConnection]
java.sql.SQLTimeoutException: The query has timed out.
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.executeStatement(SQLServerStatement.java:226) ~[mssql-jdbc-7.2.2.jre8.jar:?]
	at com.microsoft.sqlserver.jdbc.SQLServerStatement.execute(SQLServerStatement.java:744) ~[mssql-jdbc-7.2.2.jre8.jar:?]
	at com.zaxxer.hikari.pool.ProxyStatement.execute(ProxyStatement.java:95) ~[HikariCP-3.3.1.jar:?]
	at com.zaxxer.hikari.pool.HikariProxyStatement.execute(HikariProxyStatement.java) ~[HikariCP-3.3.1.jar:?]
	...
```

The java.sql.SQLTimeoutException extends java.sql.SQLTransientException and that clearly states in the documentation that  

> The subclass of SQLException is thrown in situations where a previously failed operation might be able to succeed when the operation is retried without any intervention by application-level functionality

So the JDBC specification says that the connection is still alive and well after a timeout so Hikari should **not** log a warning or close the connection!
It is up to the application to catch any SQLTimeoutException and either retry or throw exception and close the connection.

This bug was introduced in fix of https://github.com/brettwooldridge/HikariCP/issues/1308.

Right now it is impossible to use Hikari with code like this that works just fine when using a raw SQL Server JDBC connection:

```
try (Connection con = getConnectionFromPool(); Statement stmt = con.createStatement()) {
	stmt.setQueryTimeout(1);

	try (ResultSet rs = stmt.executeQuery(sql)) {
		// Handle result
	} catch (SQLTimeoutException e) {
		// Not done within a second? Use alternate query
		try (ResultSet rs = stmt.executeQuery(alternateSQL)) {
			// Handle result
		}
	}
} catch (SQLException e) {
	// Handle in some way
}

```

imjuju:
Fix : #1388 

* Rollback commit cda2605accd258b569bfd9dbf16265faaf1fa18c according to JDBC specification
* ~~Add a rollback when evicting a connection~~

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -160,7 +160,6 @@ public abstract class ProxyConnection implements Connection
             break;
          }
          else if (sqlState != null && sqlState.startsWith("08")
-             || nse instanceof SQLTimeoutException
              || ERROR_STATES.contains(sqlState)
              || ERROR_CODES.contains(nse.getErrorCode())) {
 


svendiedrichsen:
@brettwooldridge This PR seems good to me. Especially when it brings JDBC conformity. What do you think?

svendiedrichsen:
Also dropping db connections in case of SQLTimeoutExeptions seems IMHO wrong. Which this PR corrects.
If those two points make sense to you please consider this PR @brettwooldridge .

brettwooldridge:
I haven't looked at it in a while, but I didn't close it, so it would be considered. However, it appears abandoned, so unless it is rebased and the conflicts resolved, it is unlikely to be merged.

imjuju:
Hello, @brettwooldridge I've just completed rebase and resolved the conflicts. Could you please consider this PR?

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
```diff
@@ -135,10 +135,14 @@ void quietlyCloseConnection(final Connection connection, final String closureRea
 
             // continue with the close even if setNetworkTimeout() throws
             try (connection) {
-               if (!connection.isClosed())
+               if (!connection.isClosed()) {
+                  if (!connection.getAutoCommit()) {
```
svendiedrichsen: Should there be some check for a dirty state of the connection to avoid rollbacks on all connections closed?

imjuju: When auto commit is true, calling rollback results in an error, so I had to add additional handling.

svendiedrichsen: Which is ok IMO, but to avoid rollbacks on every connection closed, can we check for connections to have uncommitted changes before rollback?

imjuju: As far as I know, there is no way to check if there are uncommitted changes in a connection. If I’m mistaken, please let me know. Also, as mentioned in the JDBC close method documentation, I think it would be better to explicitly call rollback before closing the connection.

> It is strongly recommended that an application explicitly commits or rolls back an active transaction prior to calling the close method.

svendiedrichsen: ok

svendiedrichsen: The ProxyConnection knows about the dirtyCommitState and could decide not to call its delegate rollback Method if not dirty.

imjuju: I'm concerned that this method might be called before a ProxyConnection is created through the ProxyFactory.

svendiedrichsen: I don't know if and why the factory would do this and if it is ever a problem to suppress rollbacks if the connection is not marked dirty.

imjuju: I agree with the idea of suppressing rollbacks by checking isCommitStateDirty. However, my concern is that the connection might not be an instance of ProxyConnection. Do you think it would be better to use instanceof to cast and check the value of isCommitStateDirty?

svendiedrichsen: Yes, this could be a way. In case the dynamically created instance fails the instanceOf check it might be better to use isAssignableFrom instead.

imjuju: ok, I'll make the revisions as you suggested. However, I have some other schedule, so I might not be able to work on it right away. I plan to complete it within a day or two.

imjuju: Hello @svendiedrichsen. As per your advice, I tested the code below within my application.
When quietlyCloseConnection method is called, the Connection object is not always a ProxyConnection,  which makes the first if statement unnecessary. Sadly, it seems I’ll need to keep the current code as it is.
```java
private boolean shouldRollback(final Connection connection) throws SQLException
{
   if (ProxyConnection.class.isAssignableFrom(connection.getClass())) {
      ProxyConnection proxyConnection = (ProxyConnection) connection;
      return proxyConnection.getCommitStateDirtyState() && !proxyConnection.getAutoCommitState();
   }

   return !connection.getAutoCommit();
}
```

svendiedrichsen: Ok.

svendiedrichsen: Could you point to the part of the JDBC spec where it says to rollback connections when returning it to the pool? I'm just curious.

imjuju: The quote in https://github.com/brettwooldridge/HikariCP/pull/2238#discussion_r1831436725 is from the JDBC specifications. 
I based my changes on the suggestion in https://github.com/brettwooldridge/HikariCP/issues/1388#issuecomment-1346172517, so it might be helpful to take a look at it.

venu2-git:
Could you please tell us when these commits will be merged and moved to maven with what verion ?  We are facing partial commits issue during connection closed and transaction times out (its partial committing now).

Ideally we would like to have explicit commands for commit and rollback in our application without hikari library doing anything of that sort.

src/main/java/com/zaxxer/hikari/pool/PoolBase.java
svendiedrichsen: The second comment  talks about the case of evicting a connection from the pool. In which case a rollback is expected. The method quietlyCloseConnection is called in several more cases, isn't it?

imjuju: After an SQLTimeoutException occurs, quietlyCloseConnection closes the connection, but without a rollback, resulting in a partial commit. That’s why I modified quietlyCloseConnection. Also, I don’t think adding a rollback in quietlyCloseConnection would cause issues, even if it’s called in other cases.

If you have concerns, I can skip applying commit 0da3ec366a4d0ad10e4ba1fd174e3dfca1bd7137 and only apply commit 3d17a991a06373a80ee1be9aa787d902a4518c9b instead.

svendiedrichsen: I think it is @brettwooldridge who is to decide about the changes ultimately. So for me its good as it is. Thank you for your patience and your work.

brettwooldridge:
@venu2-git At this point I have no intention of merging this change. Rollback is already implemented on "dirty" connections (see [here](https://github.com/brettwooldridge/HikariCP/blob/1695eb5012918cc2e28b083264fad37b244aa731/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L250)). This change does nothing useful as far as I can tell, unless someone can prove otherwise. This pull request appears to be submitted by someone who doesn't understand the actual flow of the code.

imjuju:
Hello, @brettwooldridge. I have a question about the following code https://github.com/brettwooldridge/HikariCP/blob/1695eb5012918cc2e28b083264fad37b244aa731/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L162-L183

This 181 line is executed, and the connection is closed without performing a rollback. Is this behavior fine?

When an SQLTimeoutException occurs, the 181 line is executed. As a result, I am experiencing issues with partial commits in an Oracle environment. I think that even if rollback during eviction is not applied, the commit 3d17a991a06373a80ee1be9aa787d902a4518c9b should still be applied to prevent partial commits in Oracle.

brettwooldridge:
Here is the issue in a nutshell. Oracle, unique among databases, performs a _commit_ when a "dirty" connection is closed whereas nearly every other database performs a _rollback_. See [here](https://stackoverflow.com/questions/218350/does-java-connection-close-rollback/218495).

According to Oracle's documentation:
> If auto-commit mode is disabled and you close the connection without explicitly committing or rolling back your last changes, then an implicit <code>COMMIT</code> operation is executed.

The JDBC specification is neutral on this behavior (see [here](https://in.relation.to/2005/10/20/pop-quiz-does-connectionclose-result-in-commit-or-rollback)).

> When auto-commit is disabled, each transaction must be explicitly committed by calling the Connection method commit or explicitly rolled back by calling the Connection method rollback, respectively.

Regardless of what HikariCP itself can do, it would be of no use in the case of a terminated JVM. If the JVM is terminated, the default Oracle behavior of committing the connection will take effect and there is nothing that HikariCP could do about it.

Update: It appears that the WebLogic (Oracle application server) [implements the opposite behavior](https://docs.oracle.com/en/middleware/standalone/weblogic-server/14.1.1.0/jdbca/transactions.html#GUID-6A7A2ECB-B86F-423A-8C69-C4AD19E6FB77):
> There are some drivers (Oracle 10.x and 11.x driver) that do not commit the local transaction. If the application does not complete (commit or rollback) the local transaction before closing the connection, a connection is returned to the pool with outstanding work and that work may never be completed or it may be committed or rolled back by the next reservation of that connection. To prevent that situation from happening, a WebLogic data source calls commit on the connection when returning it to the pool, if running with the Oracle 10.x or 11.x driver.

But Oracle notes:
> Note: It is not a good programming practice to leave abandoned transactions. It is recommended that applications explicitly commit or rollback local transactions.

Short answer is that there is no consensus on "correct" behavior, either by the driver nor by a pool

imjuju:
@brettwooldridge I understand what you are saying. However, as many people mentioned in https://github.com/brettwooldridge/HikariCP/issues/1388 , evicting connections due to SQLTimeoutException should be removed. It is a separate issue from adding rollback to evict. I have updated the PR with this content, so I would appreciate it if you could consider it.
Otherwise, it would be helpful to provide a configuration to remove this row. (especially for Oracle users)
https://github.com/brettwooldridge/HikariCP/blob/1695eb5012918cc2e28b083264fad37b244aa731/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java#L163

brettwooldridge:
See also ec6891a0 for updated ``SQLExceptionOverride`` handling behavior, specifically the introduction of the ``Override.MUST_EVICT`` enum return value.

venu2-git:
Guys, looks like a commit made it through the merge, when the artifact will be made available through maven repository, I'm a novice regarding the pipeline from git to maven, which version of the artifact will make it to maven?

brettwooldridge:
@venu2-git I published a new version last night. The latest version is v6.2.0 and should be available in the maven central repository now.

venu2-git:
@brettwooldridge Looks like the 6.2.0 version has a dependency on apache compress.
Getting the build error in my local project
Could not find commons-compress-1.27.2-SNAPSHOT.jar (org.apache.commons:commons-compress:1.27.2-SNAPSHOT:20240912.171524-3).

Looks like 1.27.2 hasnt make it yet to the maven.  Only upto 1.27.1 that too last Aug,24. Im on gradle, which looks for these libraries when i move to hikari 6.2.0

Is it possible to rectify or there is a remedy for it ?

Thanks for all the hardwork guys.

brettwooldridge:
@venu2-git Oh man, good catch! I upgraded the dependency, which was supposed to be "test" scope, and left off the ``<scope>`` clause. Releasing 6.2.1 now.

UPDATE: Published. It should appear in the maven central repository within 15-30 minutes.

PR #2240, opened 2024-09-24
Use volatile field instead of AtomicReference

quaff:
it's not necessary to use `AtomicReference` here since the result of atomic method `updateAndGet()` is unused, and not necessary to use `AtomicReferenceFieldUpdater` here since only `set()` is using, volatile field access is more appropriate.


brettwooldridge:
This is incorrect. _volatile_ does not provide the same cross-thread memory consistency guarantees as the barriers provided by ``AtomicReference`` and ``AtomicReferenceFieldUpdater``. While _volatile_ may work in the current code, it is fragile towards future changes in non-obvious ways due to its weaker guarantees. [Java Memory Model](https://www.jcp.org/en/jsr/detail?id=133)

PR #2246, opened 2024-09-27
Add support for copying state of final Collection and Map

quaff:
And mark Properties field as final


brettwooldridge:
What problem is this solving?

quaff:
> What problem is this solving?

We can make them `final` for immutability.

```java
private final Properties dataSourceProperties = new Properties();
```

brettwooldridge:
This doesn't protect anything real. It protects the *reference* but the properties themselves are still mutable.

quaff:
> This doesn't protect anything real. It protects the _reference_ but the properties themselves are still mutable.

Yes, just for improving coding style, feel free to close it if you think it's not worthy.

Issue #2191, opened 2024-04-08
Developer experience: allow configuring HikariCP with `java.time.Duration` instead of millisecond / second parameters


hiddewie:
Currently, configuring the `idleTimeout` property of the connection pool will use the `setIdleTimeout` method on the `HikariConfig` configuration class. 

Similarly, many of the other configuration parameters are duration values. The values range from millisecond, to second, to minute. This difference in order of magnitude can make it error-prone to configure the connection pool correctly, and review the configured values.

It would be great to be able to configure the connection pool using `java.time.Duration` values, for example
```java
configuration.setIdleTimeout(Duration.ofMinutes(5))
configuration.setIdleTimeout(Duration.ofSeconds(10))
```
This makes the intent clear of the code more clear than the current
 ```java
configuration.setIdleTimeout(300000)
configuration.setIdleTimeout(5 * 60 * 1000)
configuration.setIdleTimeout(10000) // 10 seconds, or did I make a typo got a zero too much or too little?
```

I am willing to build a pull request if this feature is found to be useful.

Implementation seems to be extending the `HikariConfigMXBean` with extra setter methods, and extending the implementation `HikariConfig`. 

Furthermore, `PropertyElf` could be extended to take into account property values of the same syntax, for example:
```properties
# 5 minutes idle timeout, all properties are equivalent
idleTimeout = 300000 
idleTimeout = 300s
idleTimeout = 5m
```

brettwooldridge:
I will consider a pull request change to the ``PropertyElf`` that allows HikariCP to accept properties of the form ``300s`` or ``5m``, for example.

I will not change the API of ``HikariConfig`` to accept ``Duration`` as this would be disruptive, and adding overridden methods with alternative signatures unnecessarily increases the surface area of the API. It is just as clear to write:
```java
import static java.util.concurrent.TimeUnit.*;
...
configuration.setIdleTimeout(MINUTES.toMillis(5))
configuration.setIdleTimeout(SECONDS.toMillis(10))
...
```

hiddewie:
Thanks for the comment. I will try to prepare a pull request for parsing the properties from duration strings in `PropertyElf`.

 > It is just as clear to write:
```java
configuration.setIdleTimeout(MINUTES.toMillis(5))
```

Yes, for writing the code yes, but my point of this issue is that it is unclear which units the parameter of `setIdleTimeout` should accept (milliseconds).

It is just as easy as the snippet above to write
```java
[...]

PR #2266, opened 2024-11-24
Support parsing `Duration` values in `PropertyElf`

hiddewie:
From https://github.com/brettwooldridge/HikariCP/issues/2191

Add support for parsing duration values in the `PropertyElf` class. This utility is used to populate the Hikari config and the data source properties from a `Properties` object (possibly parsed from a file).

The pull request adds support for millisecond, second, minute, hour and day durations, in a fixed form. Negative values are not supported.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/PropertyElf.java
+++ b/src/main/java/com/zaxxer/hikari/util/PropertyElf.java
@@ -20,7 +20,10 @@ import com.zaxxer.hikari.HikariConfig;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Method;
+import java.time.Duration;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * A class that reflectively sets bean properties on a target object.
@@ -29,6 +32,8 @@ import java.util.*;
  */
 public final class PropertyElf
 {
+   private static final Pattern DURATION_PATTERN = Pattern.compile("^(?<number>\\d+)(?<unit>ms|s|m|h|d)$");
+
    private PropertyElf() {
       // cannot be constructed
    }
@@ -140,23 +145,24 @@ public final class PropertyElf
 
       try {
          var paramClass = writeMethod.getParameterTypes()[0];
+         String value = propValue.toString();
          if (paramClass == int.class) {
-            writeMethod.invoke(target, Integer.parseInt(propValue.toString()));
+            writeMethod.invoke(target, parseDuration(value).map(duration -> (int) duration.toMillis()).orElseGet(() -> Integer.parseInt(value)));
          }
          else if (paramClass == long.class) {
-            writeMethod.invoke(target, Long.parseLong(propValue.toString()));
+            writeMethod.invoke(target, parseDuration(value).map(Duration::toMillis).orElseGet(() -> Long.parseLong(value)));
          }
          else if (paramClass == short.class) {
-            writeMethod.invoke(target, Short.parseShort(propValue.toString()));
+            writeMethod.invoke(target, Short.parseShort(value));
          }
          else if (paramClass == boolean.class || paramClass == Boolean.class) {
-            writeMethod.invoke(target, Boolean.parseBoolean(propValue.toString()));
+            writeMethod.invoke(target, Boolean.parseBoolean(value));
          }
          else if (paramClass.isArray() && char.class.isAssignableFrom(paramClass.getComponentType())) {
-            writeMethod.invoke(target, propValue.toString().toCharArray());
+            writeMethod.invoke(target, value.toCharArray());
          }
          else if (paramClass == String.class) {
-            writeMethod.invoke(target, propValue.toString());
+            writeMethod.invoke(target, value);
          }
          else {
             try {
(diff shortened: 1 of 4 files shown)


src/main/java/com/zaxxer/hikari/util/PropertyElf.java
```diff
@@ -140,23 +145,24 @@ private static void setProperty(final Object target, final String propName, fina
 
       try {
          var paramClass = writeMethod.getParameterTypes()[0];
+         String value = propValue.toString();
          if (paramClass == int.class) {
-            writeMethod.invoke(target, Integer.parseInt(propValue.toString()));
+            writeMethod.invoke(target, parseDuration(value).map(duration -> (int) duration.toMillis()).orElseGet(() -> Integer.parseInt(value)));
```
brettwooldridge, line 150: There are no methods that take ``int`` parameters that are durations/timeouts.

src/test/resources/duration-config.properties
```diff
@@ -0,0 +1,8 @@
+connectionTimeout = 11ms
+validationTimeout = 22s
+idleTimeout = 33m
+leakDetectionThreshold = 44h
+maxLifetime = 55d
+
+dataSourceClassName=com.zaxxer.hikari.mocks.StubDataSource
+dataSource.loginTimeout = 47m
```
brettwooldridge, line 8: At least one of these values should be a "naked" value without ms|s|m|h|d.

src/main/java/com/zaxxer/hikari/util/PropertyElf.java
```diff
@@ -180,4 +186,29 @@ private static String capitalizedPropertyName(String propertyName)
...
+            case "s":
+               return Optional.of(Duration.ofSeconds(number));
+            case "m":
+               return Optional.of(Duration.ofMinutes(number));
+            case "h":
+               return Optional.of(Duration.ofHours(number));
+            case "d":
+               return Optional.of(Duration.ofDays(number));
+            default:
+               throw new IllegalStateException(String.format("Could not match unit, got %s (from given value %s)", unit, value));
```
svendiedrichsen, line 208: Shouldn't this rather be an IllegalArgumentException?

brettwooldridge: Yes, ``IllegalArgumentException`` would be more appropriate.

brettwooldridge:
Other than the above, the code looks clean and I like it.

src/main/java/com/zaxxer/hikari/util/PropertyElf.java
```diff
@@ -180,4 +186,29 @@ private static String capitalizedPropertyName(String propertyName)
       // use the english locale to avoid the infamous turkish locale bug
       return propertyName.substring(0, 1).toUpperCase(Locale.ENGLISH) + propertyName.substring(1);
    }
+
+   private static Optional<Duration> parseDuration(String value)
+   {
+      Matcher matcher = DURATION_PATTERN.matcher(value);
```
brettwooldridge, line 192: Use ``var`` type inference (everywhere possible).

brettwooldridge:
It's 2am in Tokyo, I'll review the changes tomorrow.

src/main/java/com/zaxxer/hikari/util/PropertyElf.java
hiddewie: I made it an `IllegalStateException` because the regular expression should already catch any non-matching value (the accepted units are part of the regular expression). So if it ever hits the `default` clause, there is something wrong with the code, not the user's input value.

However I am entirely fine with making it `IllegalArgumentException` if that is preferred.

brettwooldridge: I'll agree to this. It does not look reachable based on the regex. Leave it as is, it should invoke surprise should it ever occurr.

brettwooldridge:
@hiddewie Please update your pull request.

hiddewie:
> @hiddewie Please update your pull request.

Sorry, with what changes do you mean?

brettwooldridge:
The ones pointed out in the pending review comments above. For example, for the code:
```
Matcher matcher = DURATION_PATTERN.matcher(value);
```

The requested change is:

> Use ``var`` type inference (everywhere possible).

Meaning:
```
var matcher = DURATION_PATTERN.matcher(value);
```
....

brettwooldridge:
I lost patience. I'll make the modifications myself.

brettwooldridge (commented):
merged and fixed

hiddewie:
> The ones pointed out in the pending review comments above. For example ...

Sorry, but these comments were not submitted yet. I only see them after the last comment yesterday, after you merged the pull request.

The only comment that was visible in Github was https://github.com/brettwooldridge/HikariCP/pull/2266#discussion_r1855489974, and that one was resolved after the discussion.

image

hiddewie:
But thanks anyway for the review and feedback here, and merging the PR 🎉 

I will close https://github.com/brettwooldridge/HikariCP/issues/2191 because the further changes will not be implemented as decided in https://github.com/brettwooldridge/HikariCP/issues/2191#issuecomment-2496067977

brettwooldridge:
Yeah, I hate that about github. Second time that's happened. UX issue IMHO. Adding a review comment should *start* the review. Instead you have to find the button at the top of the page (away from where you are commenting) to start the review process. Anyway, it's all good. Thanks for the PR.

PR #2296, opened 2025-02-24
update log level to debug

GertDE:
This prevents unnecessary log spam in production environments.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
+++ b/src/main/java/com/zaxxer/hikari/pool/ProxyConnection.java
@@ -175,7 +175,7 @@ public abstract class ProxyConnection implements Connection
 
       if (evict) {
          var exception = (nse != null) ? nse : sqle;
-         LOGGER.warn("{} - Connection {} marked as broken because of SQLSTATE({}), ErrorCode({})",
+         LOGGER.debug("{} - Connection {} marked as broken because of SQLSTATE({}), ErrorCode({})",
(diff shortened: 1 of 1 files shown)


brettwooldridge:
@GertDE Hmmmm, why is this happening in production? This is definitely not "expected" behaviour from the perspective of the pool.

GertDE:
@brettwooldridge I'm using Spring and catching QueryTimeoutException (query is interrupted after 1 second) in my application, but HikariCP still logs a warning, so i can`t grep for real exceptions.

brettwooldridge:
@GertDE What about [this solution](https://www.perplexity.ai/search/is-there-a-way-to-use-log4j2-c-1KBFMlfGRWefQ0SLTbFOuw)?

GertDE:
@brettwooldridge I've addressed this locally by raising the logging level to ERROR:
`logging.level.com.zaxxer.hikari.pool.ProxyConnection=ERROR`
In general, I consider it bad practice to log exceptions that aren't actually thrown, as it creates noise in the logs and makes real issues harder to identify. Could you share the reasoning behind logging an exception in this case? Wouldn't the DEBUG level be more appropriate for analytical purposes?

brettwooldridge:
@GertDE It is best practice that exceptions that are eaten must be logged.

https://javanexus.com/blog/handling-swallowed-exceptions-java-guide

GertDE:
Yes, that's correct, but you're handling the exception and rethrowing it, so it's not a swallowed exception.

PR #2300, opened 2025-03-18
When creating physical connections, catch throwable instead of exception

cgravatt:
Although rare, JDBC drivers can and do throw java.lang.Error instead of java.lang.Exception under certain conditions. If this occurs, the Error is swallowed since it is not currently caught and it occurs on a separate thread where the result is never checked.

 If the driver continues to throw Error, the backoff logic is bypassed, the connection creation counter spikes and eventually the pool drains of all connections.

See example

Graph of connection creation counter
image

Graph of connections for the pool
image

The trigger was a networking event. At this time, we don't know with absolute certainty that the driver was generating an Error. However, after reviewing the code, I could not determine another explanation. The backoff logic clearly was not being hit as the counter would not increase and stay at the rate. We put out a patched version of Hikari similar to the changes in this PR with hopes of confirming that is what is happening and collecting the actual error being thrown, but we are waiting for the situation to occur again.

In this case we are using oracle's jdbc driver, and we have seen it throw a java.lang.Error before in other situations.

If this change is of interest and you want an explicit test for this, I'm happy to do that, but I may need a little direction on preferences here.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -505,7 +505,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
             logger.debug("{} - Error thrown while acquiring connection from data source", poolName, e.getCause());
          }
       }
-      catch (Exception e) {
+      catch (Throwable e) {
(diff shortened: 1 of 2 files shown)


tbadalov:
> Although rare, JDBC drivers can and do throw java.lang.Error instead of java.lang.Exception under certain conditions

Probably, they don't want you to recover from them, do they?

> However, after reviewing the code, I could not determine another explanation

What was the exception? JVM could throw asynchronous exceptions, not necessarily the jdbc driver

PR #2305, opened 2025-03-26
Keep properties keys and values as it is

### LINKED ISSUE #2286: All properties being converted to strings.
In DriverDataSource line 49 all properties are getting converted to strings:

      for (var entry : properties.entrySet()) {
         driverProperties.setProperty(entry.getKey().toString(), entry.getValue().toString());
      }

This is an issue for the Snowflake driver because the only successful way to connect with an encrypted private key is to add the property "privateKey" with a value of an object of type java.security.PrivateKey. When that object gets converted to a string the connection fails with the following error:
Caused by: net.snowflake.client.jdbc.SnowflakeSQLLoggedException: Invalid parameter value type: java.lang.String, expected type: java.security.PrivateKey.

Changing the line to the following results in a successful connection:
      for (var entry : properties.entrySet()) {
         driverProperties.put(entry.getKey(), entry.getValue());
      }

Can you advise? I hate to have to override the DriverDataSource class locally.

### LINKED ISSUE #2304: HikariCP 3.4.5: Snowflake JDBC driver: privateKey expected java.security.PrivateKey, received String
In a Grails 2.x application, HikariCP is used to connect to Snowflake. A java.security.PrivateKey object is configured as the privateKey datasource property. However, during the getConnection call, HikariCP serializes this PrivateKey into a string and includes it in the driverProperties, leading to a connection error. The Snowflake JDBC driver expects a PrivateKey object, not a string representation.

**Below is the connection pool configuration**

```
HikariConfig config = new HikariConfig(properties)
            config.addDataSourceProperty("user", "username")
            config.addDataSourceProperty("password", "")
            config.addDataSourceProperty("privateKey", privateKey)
            bean.constructorArgs = [config]
```

**Runtime view:**

```
driverProperties = {Properties@17721}  size = 3
 "privateKey" -> "sun.security.rsa.RSAPrivateCrtKeyImpl@ffef541c"
 "user" -> "username"
 "password" -> ""
```
**Error message :** 

`Failed to instantiate [com.zaxxer.hikari.HikariDataSource]: Constructor threw exception; nested exception is com.zaxxer.hikari.pool.HikariPool$PoolInitializationException: Failed to initialize pool: Invalid parameter value type: java.lang.String, expected type: java.security.PrivateKey.`

quaff:
Fix GH-2304
Fix GH-2286

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
@@ -45,9 +45,7 @@ public final class DriverDataSource implements DataSource
       this.jdbcUrl = jdbcUrl;
       this.driverProperties = new Properties();
 
-      for (var entry : properties.entrySet()) {
-         driverProperties.setProperty(entry.getKey().toString(), entry.getValue().toString());
-      }
+      driverProperties.putAll(properties);
 
       if (username != null) {
          driverProperties.put(USER, driverProperties.getProperty(USER, username));
--- a/src/test/java/com/zaxxer/hikari/util/DriverDataSourceTest.java
+++ b/src/test/java/com/zaxxer/hikari/util/DriverDataSourceTest.java
@@ -18,6 +18,8 @@ package com.zaxxer.hikari.util;
 
 import org.junit.Test;
 
+import java.lang.reflect.Field;
+import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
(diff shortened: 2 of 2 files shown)


quaff:
@brettwooldridge Could you take a look?

Bonu:
@quaff @brettwooldridge thanks for the fix

honnix:
Just you know this was a backward incompatible change unfortunately. Maybe it is worth mentioning in the release notes.

brettwooldridge:
@honnix In what use-case was it incompatible? If we knew it was incompatible we would have called it out, but none of our hundreds of unit test caught it as an incompatibility.

honnix:
> @honnix In what use-case was it incompatible? If we knew it was incompatible we would have called it out, but none of our hundreds of unit test caught it as an incompatibility.

We have some code relying on conversion from boolean to string for a property value (of course it is questionable code), with this change it stopped working. It is an easy fix on our side though.

ansf:
> We have some code relying on conversion from boolean to string for a property value (of course it is questionable code), with this change it stopped working. It is an easy fix on our side though.

We were affected by the very same issue and it took some time to find the root cause after updating. 
Would have been nice, if this was pointed out more clearly in the changelog (or done in a major version bump).

Issue #2265, opened 2024-11-20
No backoff if PoolEntryCreator is interrupted


PaoloVoronini:
https://github.com/brettwooldridge/HikariCP/blob/ffdb8106bccb285d1775af1e8b040bfce4588dc1/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L758C19-L758C31

Method UtilityElf.quietlySleep() retains the interrupted flag
As a result PoolEntryCreator does not sleep at all - i.e. no backoff.
Possibly the interrupted flag also prevents creating new pool entry.

```
while (shouldContinueCreating()) {
   final var poolEntry = createPoolEntry();
   if (poolEntry != null) {
      added = true;
      connectionBag.add(poolEntry);
      logger.debug("{} - Added connection {}", poolName, poolEntry.connection);
      quietlySleep(30L);
      break;
   } else {  // failed to get connection from db, sleep and retry
      if (loggingPrefix != null && backoffMs % 50 == 0)
         logger.debug("{} - Connection add failed, sleeping with backoff: {}ms", poolName, backoffMs);

      //THE PROBLEM IS HERE
      //the thread does not sleep because it is interrupted
      quietlySleep(backoffMs); 

      backoffMs = Math.min(SECONDS.toMillis(5), backoffMs * 2);
   }
}
```

@brettwooldridge
my suggestion would be to clear the interrupted flag in the snippet above.
i.e.
```
quietlySleep(backoffMs);
//clear interrupted state
Thread.interrupted();
```

PS: this might be related to #2208 and other "database outage" issues mentioned in it

PaoloVoronini:
@brettwooldridge
my suggestion would be to clear the interrupted flag in the snippet above.
i.e.
```
quietlySleep(backoffMs);
//clear interrupted state
Thread.interrupted();
```

brettwooldridge:
What would cause the adder thread to be interrupted besides shutdown?

PaoloVoronini:
thanks Brett for the quick response!

unfortunately, i was not able to figure out the root cause reliably.
similar to other issues i mentioned, we observe this upon database outages
this is the last piece of the stacktrace (please find below)
so far i can only speculate that possibly the ojdbc sets the interrupted status of the thread for some reason

PS: just in case, i should mention that i wrote a little method to support the statement "thread does not sleep"
```
public static void main(String[] args) {
    Thread t = new Thread(() -> {
        while (true) {
            try {
[...]

PR #2306, opened 2025-04-02
#2265 bail out of PoolEntryCreator#call if the thread gets interrupted

cgravatt:
#2265 bail out of PoolEntryCreator#call if the thread's interrupt flag gets set. 

Issue: Connection adder thread does not properly check and handle interrupts. Currently, if the driver fails to create a connection and decides to mark the thread's interrupt flag (and throws an Exception), you end up in the backoff code. In the backoff code, the thread will sleep. Since the thread's interrupt flag has been set, it will immediately throw an InterruptedException which will also clear the thread's interrupt flag. 

image

However, this exception is caught and HikariCP resets the interrupt flag. Since we didn't successfully create a connection, shouldContinueCreating will return true. As a result, we try to create a connection again, but if the driver checks to see if the thread has been interrupted and bails out, we end up in an endless loop and can no longer create connections.

This updates it to respond and handle the thread being marked interrupt. Since thread operates in a loop, on each iteration of the loop, it should check to see if it has been interrupted. If it has, it should not continue and clear the interrupt so that if the thread is re-used for the next task in the ThreadPool's queue, it's good to go.

I added a test that simulates the scenario. Without the change, the test fails as the adder thread is stuck in a loop and never able to obtain and return a connection. The test simulates the idea that a database outage or network outage occurred causing a failure that caused the thread to get marked interrupt. Then the database is back online at which point the pool should recover but it currently won't without this change.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -772,12 +772,13 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
 
       /**
        * We only create connections if we need another idle connection or have threads still waiting
-       * for a new connection.  Otherwise we bail out of the request to create.
+       * for a new connection.  Otherwise we bail out of the request to create. If the thread's interrupt flag
+       * gets set, we bail out and clear the interrupt flag.
        *
        * @return true if we should create a connection, false if the need has disappeared
        */
       private synchronized boolean shouldContinueCreating() {
-         return poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
+         return !Thread.interrupted() && poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
             (getIdleConnections() < config.getMinimumIdle() || connectionBag.getWaitingThreadCount() > getIdleConnections());
       }
    }
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Fast, simple, reliable.  HikariCP is a "zero-overhead" production ready JDBC con
 
 ### Artifacts
 
-_**Java 11+** maven artifact:_
+_**Java 11 or greater** maven artifact:_
 ```xml
 <dependency>
    <groupId>com.zaxxer</groupId>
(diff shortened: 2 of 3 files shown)


cgravatt:
There are a fair amount of issues reporting the pool being unable to recover or create connections. Some of those may be resolved by this. This is the issue we experience from time to time. Also confirmed by @PaoloVoronini in #2265

src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -772,12 +772,13 @@ else if (!added)
...
        * We only create connections if we need another idle connection or have threads still waiting
-       * for a new connection.  Otherwise we bail out of the request to create.
+       * for a new connection.  Otherwise we bail out of the request to create. If the thread's interrupt flag
+       * gets set, we bail out and clear the interrupt flag.
        *
        * @return true if we should create a connection, false if the need has disappeared
        */
       private synchronized boolean shouldContinueCreating() {
-         return poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
+         return !Thread.interrupted() && poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
```
PaoloVoronini, line 786: Apparently there are 2 scenarios for the "interrupt" case:
1. pool shutdown 
2. mysterious bug 

Logging may help clients become aware of the latter:

```
private synchronized boolean shouldContinueCreating() {
    if (Thread.interrupted()) {
        logger.warn("Connection adder has been interrupted");
        return false;
    }
    return poolState == POOL_NORMAL && getTotalConnections() < config.getMaximumPoolSize() &&
        (getIdleConnections() < config.getMinimumIdle() || connectionBag.getWaitingThreadCount() > getIdleConnections());
}
```

cgravatt: I don't think I have an opinion on whether it is logged or not. We'll see if @brettwooldridge does.

cgravatt:
@brettwooldridge Just following up to see if you had thoughts on this.

cgravatt:
Looks like @brettwooldridge introduced this change in 4da46f5911f2f1d5cebb229a69baa499bfd0e7da

so this is no longer necessary.

Issue #2323, opened 2025-06-18
java.sql.Connection getSchema() contract not adhered to


adarsh-cu:
Hi, 

In the versions of HikariCP starting from 6.0.0 (all 6.x.x versions), the `java.sql.Connection` interface contract is violated for the `getSchema()` in my opinion. The method contract says

`Returns: the current schema name or null if there is none`

Below is a very simple code that can replicate the difference - 

```
HikariConfig config = new HikariConfig();
        config.setJdbcUrl("jdbc:postgresql://localhost:5432/test"); // Replace with your DB URL
        config.setUsername("test"); // Replace with your DB username
        config.setPassword("test"); // Replace with your DB password
        config.setMaximumPoolSize(10);

        try (
                HikariDataSource dataSource = new HikariDataSource(config)) {
            // Get a connection from the pool
            try (Connection connection = dataSource.getConnection()) {
                // Set correct schema before proceeding
                connection.setSchema("invalid");
                if (Objects.isNull(connection.getSchema())) {
                    log.info("Connection contract adhered to.");
                } else {
                    log.error("Connection contract violated.");
                }
            }
        } catch (Exception exception) {
            log.error("An error occurred while connecting to the database", exception);
        }
```

Previously, the call to `getSchema()` was delegated, but is now handled in the `ProxyConnection` and delegated only in the case it was dirtied? Version 5.1.0 delegates the call correctly. In our case, we are using postgres underneath (so, `PgConnection` is checking if the schema actually exists and returns a null when the schema does not exist). 

Is this a correct conclusion? If so, would it just mean that we remove the `getSchema()` method from the `ProxyConnection` introduced in https://github.com/brettwooldridge/HikariCP/pull/2199/files

Best regards,
Adarsh

quaff:
> `PgConnection` is checking if the schema actually exists and returns a null when the schema does not exist

I think it's a bug of Postgresql JDBC Driver, it should throw SQLException instead.

adarsh-cu:
> > `PgConnection` is checking if the schema actually exists and returns a null when the schema does not exist
> 
> I think it's a bug of Postgresql JDBC Driver, it should throw SQLException instead.

Thanks for your answer, but why do you think so? Because the contract of `java.sql.Connection` states that null is returned in case of a non-existing schema. As an implementation of that interface, they are returning the correct result.

quaff:
I mean `connection.setSchema("invalid")` should throw exception, not `connection.getSchema()`.

[...]

PR #2346, opened 2025-08-07
Fix regression with setSchema behavior

cmuchinsky:
The change associated with https://github.com/brettwooldridge/HikariCP/issues/2323 caused a regression with the way the schema is set on connections returned from the pool. Previously the schema always reset to the configured value before a connection was returned from the pool, but after this change any explicit alterations to the schema on the connection were kept and passed along to the next consumer of that connection after its returned to the pool.

This PR restores the prior behavior and adds some tests to verify the expected behavior.


brettwooldridge:
@cmuchinsky Thank you for the contribution. The unit test is especially helpful.

PR #2372, opened 2025-11-24
Clarify keepaliveTime defaults in HikariConfig docs

jbl428:
Update keepalive getter/setter Javadocs to state the true 120 000 ms (2 min) default instead of “0 (disabled)”.


stefan-g:
Hi, it would also be helpful to have a hint in the javadoc what value should be used when a additiontonal connection pool like pgpool is used. Should client keepaliveTime be 0 or smaller than keepalivetime of the pgpool?

PR #2387, opened 2026-02-14
Add maximumPendingConnections property for load-shedding

### LINKED ISSUE #580: Limit pending connection queue size
When connections are unavailable, HIkariCP queues connections (in pending state) until they timeout. Is there a way to limit this queue size? Or is this considered unnecessary if `connectionTimeout` is set low enough?

vlsi:
Under traffic spikes or database outages, threads can accumulate unboundedly waiting for pool connections, then fail in a thundering herd after `connectionTimeout`. This adds an opt-in admission control that rejects `getConnection()` calls immediately when too many threads are already waiting, preventing waiter buildup and reducing the blast radius of pool exhaustion.

The property defaults to 0 (unlimited) for full backward compatibility.

Fixes https://github.com/brettwooldridge/HikariCP/issues/580

### DIFF
--- a/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
+++ b/src/main/java/com/zaxxer/hikari/pool/HikariPool.java
@@ -89,7 +89,7 @@ public final class HikariPool extends PoolBase implements HikariPoolMXBean, IBag
    {
       super(config);
 
-      this.connectionBag = new ConcurrentBag<>(this);
+      this.connectionBag = new ConcurrentBag<>(this, config.getMaximumPendingConnections());
       this.suspendResumeLock = config.isAllowPoolSuspension() ? new SuspendResumeLock() : SuspendResumeLock.FAUX_LOCK;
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/zaxxer/hikari/pool/HikariPool.java
```diff
@@ -89,7 +89,7 @@ public HikariPool(final HikariConfig config)
    {
       super(config);
 
-      this.connectionBag = new ConcurrentBag<>(this);
+      this.connectionBag = new ConcurrentBag<>(this, config.getMaximumPendingConnections());
```
vlsi, line 92: The drawback of passing `getMaximumPendingConnections` is that `ConcurrentBag` won't react to the changes of `maximumPendingConnections`.

An alternative options could be:
a) Pass `HikariConfig` to `ConcurrentBag`
b) Add `interface ConcurrentBagConfig { int getMaximumWaiters(); int getMaximumObjects();}` so the bag could query the required limits

Any thoughts?

PR #2409, opened 2026-06-22
Fix setCredentialsProviderClassName writing to the wrong field

vasiliy-mikhailov:
## Problem

`HikariConfig.setCredentialsProviderClassName(String)` assigns its argument to the wrong field:

```java
public void setCredentialsProviderClassName(String credentialsProviderClassName) {
   checkIfSealed();
   try {
      this.credentialsProvider = createInstance(credentialsProviderClassName, HikariCredentialsProvider.class);
      this.exceptionOverrideClassName = credentialsProviderClassName; // wrong field
   }
   ...
}
```

Two consequences:

1. `credentialsProviderClassName` is never set, so `getCredentialsProviderClassName()` keeps returning `null` after the setter runs.
2. The unrelated `exceptionOverrideClassName` field is silently overwritten with the credentials-provider class name.

## Fix

Assign the value to `this.credentialsProviderClassName`.

## Test

Adds `testCredentialsProviderClassNameRoundTrip` to `HikariConfigTest`: it sets a provider class name and asserts the getter returns it. The test fails on `dev` (`expected:<...TestCredentialsProvider> but was:<null>`) and passes with this change.

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -890,7 +890,7 @@ public class HikariConfig implements HikariConfigMXBean
 
       try {
          this.credentialsProvider = createInstance(credentialsProviderClassName, HikariCredentialsProvider.class);
-         this.exceptionOverrideClassName = credentialsProviderClassName;
+         this.credentialsProviderClassName = credentialsProviderClassName;
       }
       catch (Exception e) {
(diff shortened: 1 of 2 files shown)


svendiedrichsen:
Is this the same as #2405 ?

vasiliy-mikhailov:
Yes — apologies, this duplicates #2405 (same fix). #2405 is the earlier PR and has the more complete test (it also asserts that `exceptionOverrideClassName` is not clobbered), so I am closing this one in favour of #2405.

PR #2411, opened 2026-07-13
Do not synchronize on System.getProperties() when generating pool name (#2104)

### LINKED ISSUE #2104: Synchronizing System.getProperties in generatePoolName / Number creates a deadlock-able window.

seonwooj0810:
Fixes #2104

### Root cause
`HikariConfig.generatePoolName()` held the `System.getProperties()` monitor across the whole read-increment-write of the global `com.zaxxer.hikari.pool_number` counter:

```java
synchronized (System.getProperties()) {
   final var next = String.valueOf(Integer.getInteger("com.zaxxer.hikari.pool_number", 0) + 1);
   System.setProperty("com.zaxxer.hikari.pool_number", next);
   return prefix + next;
}
```

The JVM and third-party code lock that *same* `Properties` object (`System.getProperty()`, class initialization touching system properties, etc.). Holding its monitor across a critical section therefore opens a lock-ordering window: another thread initializing classes can hold a class-init lock while waiting for the `Properties` monitor, while our thread holds the `Properties` monitor and (incidentally, during the block) needs that class-init lock. The reporter observed this deadlock roughly 1 in 4 startups in a constrained, multi-core-limited container. A library should not synchronize on a monitor outside its own scope.

### Change
Synchronize on the interned property-key `String` constant instead of the `Properties` object. A string literal is interned into the JVM-global pool, so `POOL_NUMBER_PROPERTY` resolves to the **same instance across all class loaders** — this preserves the cross-class-loader global uniqueness of the counter (the requirement that ruled out a per-class-loader `AtomicInteger`, e.g. two wars in one JVM must not collide). Inside the block the `Properties` object is only ever touched through its own already-synchronized accessors (`Integer.getInteger` / `System.setProperty`), each of which acquires and releases the `Properties` monitor atomically, so it is never *held* across other work — closing the lock-ordering window.

The `AccessControlException` / random-name fallback is unchanged.

### Tests
Added `PoolNameGenerationTest`:
- `testGeneratedPoolNamesAreUniqueUnderConcurrency` — 16 threads × 250 generations; asserts all 4000 names are unique, verifying the read-increment-write stays atomic (a lost update would produce a duplicate). This guards the global-uniqueness contract that previous fix attempts broke.
- `testGenerationDoesNotDeadlockWhilePropertiesMonitorIsContended` — a background thread continuously holds the `System.getProperties()` monitor while pool names are generated; generation must make progress and stay unique.

A fully deterministic reproduction of the original class-init/`Properties` lock inversion is impractical in a unit test, so the fix is argued from the lock-ordering analysis above; the added tests lock in the atomicity/uniqueness contract that the change must not regress.

Verification done: built and ran `mvn test -Dtest=HikariConfigTest,MiscTest,PoolNameGenerationTest` (JDK 21) — all 7 pass. Confirmed the previously-passing `MiscTest` still passes (an earlier iteration that added an `Object`-typed field regressed `copyStateTo`, which reflectively treats any final `Object` field as an `AtomicReference`; typing the constant as `String` avoids that).

### DIFF
--- a/src/main/java/com/zaxxer/hikari/HikariConfig.java
+++ b/src/main/java/com/zaxxer/hikari/HikariConfig.java
@@ -47,6 +47,16 @@ public class HikariConfig implements HikariConfigMXBean
    private static final Logger LOGGER = LoggerFactory.getLogger(HikariConfig.class);
 
    private static final char[] ID_CHARACTERS = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ".toCharArray();
+   // The pool-number counter is kept in a JVM-global system property so that pool numbers do not
+   // overlap across class loaders (e.g. two wars in one JVM). Making the read-increment-write of
+   // that property atomic requires a monitor shared by every class loader, but it must NOT be
+   // System.getProperties(): the JVM and third-party code lock that same Properties object (via
+   // System.getProperty(), class initialization, etc.), so holding its monitor across our critical
+   // section opens a lock-ordering window that can deadlock (see #2104). Instead we synchronize on
+   // this interned String constant, which is unique to HikariCP yet resolves to the same instance
+   // in every class loader, and we only touch the Properties object through its own
+   // already-synchronized accessors.
+   private static final String POOL_NUMBER_PROPERTY = "com.zaxxer.hikari.pool_number";
    private static final long CONNECTION_TIMEOUT = SECONDS.toMillis(30);
    private static final long VALIDATION_TIMEOUT = SECONDS.toMillis(5);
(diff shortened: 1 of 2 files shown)


seonwooj0810:
Wanted to check in on this small fix around avoiding synchronization on System.getProperties(). Open to a different approach if you'd prefer.

Issue #2042, opened 2023-02-24
DriverDataSource set global LoginTImeout effect hive-jdbc


shalk:
I have a case , hive-jdbc  have used[ DriverManager.getLoginTimeout as socketTimeout](https://github.com/apache/hive/blob/branch-3.1/jdbc/src/java/org/apache/hive/jdbc/HiveConnection.java#L812) for a long time.

In default,  global variable DriverManager.loginTimeout is `0`, hive-jdbc will use zero, so timeout is zero, it work fine.

**When i use hikarcp with mysql in my project,   hikari will set   DriverManager.loginTimeout to 30. so the hive-jdbc  sql will timeout。**

Though hive-jdbc read the  global variable maybe not make sense. i will try to fix it.

i found https://github.com/brettwooldridge/HikariCP/commit/e1563e46d6b8c423c0bb90b65ea36905b89053c2  make loginTIme not a  local variable ，but  setLoginTImeout, i want to know why hikari need to set global logintimeout.

How about  keep to  set local logintimeout ? the local logintimeout is in scopt of the hikaridatasource

PR #2412, opened 2026-07-16
fixes #2404 keep loginTimeout per DriverDataSource instead of the JVM-global DriverManager

### LINKED ISSUE #2404: Data sources without explicitly specified dataSource or dataSourceClass overwrite each others connectionTimeouts
**Steps to reproduce:**

1. Create a HikariDataSource A with `HikariConfiguration` containing a `jdbcUrl` and a `driverClassName` (but neither `dataSourceClassName`, `dataSourceJndiName` nor `dataSource`) and a `connectionTimeout` of 10s.
2. Create a HikariDataSource B with a different `jdbcUrl` and a timeout of 20s.

**Actual result:**
Both data sources use a connection timeout of 20s (the last one wins) because Hikari uses the static method `java.sql.DriverManager#setLoginTimeout`

**Expected result:**
Both data sources use their own individual connection timeouts.

This probably means having to use some vendor specific properties, but I think it is preferable to the current behavior. Even ignoring the connectionTimeout when no data source is set explicitly would be better than data sources overwriting each others timeouts.

Develop-KIM:
`DriverDataSource#setLoginTimeout` delegates to `DriverManager.setLoginTimeout`, which is a JVM-wide static. So with two pools configured by `jdbcUrl`, the second one to initialize overwrites the first one's login timeout, and `getLoginTimeout` then reports that same value back to both. That's #2404. It also leaks HikariCP's timeout into unrelated JDBC code in the same JVM, which is what #2042 hit with hive-jdbc.

I changed it to keep the value in an instance field.

One thing I'd like your call on: drivers that read `DriverManager.getLoginTimeout()` will no longer see a value set by HikariCP. That hint was already unreliable as soon as there was more than one pool (last one to start wins), but it did do something in the single-pool case, so removing it isn't free. If you'd rather keep it somehow, I'm happy to rework this.

The test in `DriverDataSourceTest` fails on dev with `expected:<10> but was:<20>` and passes with the change. Rest of the suite is unchanged locally — the 7 `PostgresTest` errors I see are just Docker not being available on my machine, same before and after.

Co-authored with Claude (the code and reasoning are mine to defend).

### DIFF
--- a/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
+++ b/src/main/java/com/zaxxer/hikari/util/DriverDataSource.java
@@ -46,6 +46,7 @@ public final class DriverDataSource implements DataSource
    private final String jdbcUrl;
    private final Properties driverProperties;
    private Driver driver;
+   private int loginTimeout;
 
    public DriverDataSource(String jdbcUrl, String driverClassName, Properties properties, String username, String password)
    {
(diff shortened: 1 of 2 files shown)


Develop-KIM:
Reopened as #2414 with the same commit — I'd deleted my fork when I closed this, so GitHub wouldn't let me reopen it here.

